PR #61, opened 2015-12-02
Lift and IsWellDefined in MatrixCategory

HereAround:
To apply "CAPPresentationCategory" to "MatrixCategory" it is necessary to have Lift installed for MatrixCategory. I also added Colift.


HereAround:
actually also "IsWellDefinedForMorphisms" is needed. So I also added "IsWellDefinedForMorphisms" and "IsWellDefinedForObjects" to MatrixCategory.

PR #75, opened 2016-09-25
Added IsZero to PROPAGATION_LIST_FOR_EQUAL_MORPHISMS

mohamed-barakat:


sebastianpos:
This functions propagates between equal objects, which by definition are in the same category. Please explain your added comment.

mohamed-barakat:
We (ab)use this function in IntrinsicCategories to propagate properties and attributes between objects (in the underlying category, 1st argument) and their hulls (in the intrinsic category, 2nd argument). So although it is entirely symmetric in your context, in ours its not.

sebastianpos:
Since there can be properties in the future which are stable under 
equality of objects, but not under isomorphisms (e.g. something like
IsEqualToZeroObject), there are actually more problems with your abuse of
`INSTALL_TODO_LIST_FOR_EQUAL_OBJECTS` than just the order of the arguments.
I think you need a new functionality which should be formulated as a feature request and
added separately to the CAP kernel.

mohamed-barakat:
I see. Thanks.

PR #84, opened 2017-04-04
IsWellDefined in MatrixCategory

HereAround:


sebasguts:
The checks if source and range are in the same category as the morphism are not necessary, since this is already checked by the constructor. Same is true for the field.

sebastianpos:
Since `IsWellDefinedForMorphisms` should be used for debugging,
I think it's okay not to rely on the constructor to work properly and
thus to have possible redundancies.

For the field, I would rather check if all the underlying fields equal
`category!.field_for_matrix_category` (which we should make an attribute in Cap).

PR #87, opened 2017-05-19
Added convenience methods for MorphismBetweenDirectSums and DirectSum(Op)

mohamed-barakat:


CAP/gap/UniversalObjects.gd
```diff
@@ -1098,6 +1098,9 @@ DeclareOperation( "AddInitialObjectFunctorial",
 DeclareOperationWithCache( "DirectSumOp",
                            [ IsList, IsCapCategoryObject ] );
 
+DeclareOperation( "DirectSumOp",
```
sebasguts: How is this function supposed to be called? Do you call it explicitly with a category? Then one might want to name it DirectSum.

sebastianpos: DirectSum is a global function in Gap that automatically redirects to DirectSumOp.

sebasguts: Okay. So one has to call DirectSumOp on empty lists, giving the category.

sebastianpos: Actually you can call `DirectSum( category, list_of_objects )`,
which redirects to `DirectSumOp( list_of_objects, category )`.

sebasguts: I see :)

CAP/gap/UniversalObjects.gi
```diff
@@ -421,6 +438,20 @@ InstallMethod( MorphismBetweenDirectSums,
 end );
 
 ##
+InstallMethod( MorphismBetweenDirectSums,
+               [ IsCapCategoryObject, IsList, IsCapCategoryObject ],
+               
+  function( S, morphism_matrix, T )
+    
+    if IsZero( S ) or IsZero( T ) then
```
sebastianpos: Depending on the category this test might be way too expensive
(e.g. in the case of sheaves on toric varieties).
Doesn't it suffice to check if the given list is either empty
or a list containing only empty lists?

sebasguts: The added implementation of `DirectSumOp` suggests that this is the use-case, and I also think this should be the way to do it. E.g., for modules, @sebastianpos way would produce the morphism one might have wanted to produce, whereas this implementation would construct a matrix made of `0`s.

mohamed-barakat: Thanks, addressed in a650b47.

PR #91, opened 2017-06-05
Renamed kernel object functorial

mohamed-barakat:


sebasguts:
Can we introduce `KernelObjectFunctorial` as an alias to `KernelFunctorial`?

mohamed-barakat:
It sounds for me like a good idea.

PR #93, opened 2017-06-05
New component functorial

mohamed-barakat:


sebasguts:
We can add this component without any problem, there will not be any compatibility problem.

We could also generate this entry automatically, what could keep us from forgetting to set it for other methods.

At the moment, this entry is not documented. The right place to document it is the `InstallAdds.autodoc`, where all method record entries are documented (also automatically generated ones).

sebastianpos:
I think for an automatic generation of these entries,
the naming convention has to be changed as follows:

-  `CokernelFunctorial` -> `CokernelObjectFunctorial`
- `KernelObjectFunctorial` stays as it was

With this convention, all entries of the form
`XFunctorial`
would have 
X = name of the function that gives the object of the universal construction,
e.g.,
`TerminalObjectFunctorial,` `DirectSumFunctorial`, etc.

mohamed-barakat:
I agree.

sebasguts:
We can associate `KernelFunctorial` to `KernelObject`, as long as the rule is to drop the `Object` in all cases. For consistent and very descriptive naming in `CAP` we could also go to `CokernelObjectFunctorial`, for me both is fine.

PR #97, opened 2017-07-17
Made argument_list contain all arguments

sebasguts:
unless explicitly stated otherwise. Up to now,
if all but the last argument were lists or int,
and no argument_list was given, it was implicitly
assumed that the last argument is a method selection argument.
This was the wrong behaviour. Please note, as stated in the
manual, if there is a method selection argument, the argument_list
has to be stated explicitly in the method record.


sebastianpos:
Very good.

PR #102, opened 2017-07-24
Subtraction for morphisms

mohamed-barakat:
One instead of two calls to CAS and one less matrix to create and temporarily save.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gi
+++ b/CAP/gap/CategoryMorphisms.gi
@@ -125,6 +125,12 @@ InstallMethod( AdditionForMorphisms,
 \+ );
 
 ##
+InstallMethod( SubtractionForMorphisms,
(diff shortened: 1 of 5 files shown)


mohamed-barakat:
I agree and will add the derivation.

CAP/gap/CategoryMorphisms.gi
```diff
@@ -125,6 +125,12 @@ InstallMethod( AdditionForMorphisms,
 \+ );
 
 ##
+InstallMethod( SubtractionForMorphisms,
+               [ IsCapCategoryMorphism, IsCapCategoryMorphism ],
+               
+\+ );
```
sebasguts: This should be `\-`

mohamed-barakat:
Made requested fix and added requested derivation. Thanks for the review.

mohamed-barakat:
Danke.

PR #103, opened 2017-07-24
Multiply with element of commutative ring

mohamed-barakat:
I need this in `FunctorCategories`.

### DIFF
--- a/CAP/gap/ConstructiveCategoriesRecord.gi
+++ b/CAP/gap/ConstructiveCategoriesRecord.gi
@@ -18,6 +18,7 @@ IsEnrichedOverCommutativeRegularSemigroup := Concatenation(
 IsAbCategory := Concatenation( [ 
   "ZeroMorphism", 
   "IsZeroForMorphisms", 
+  "SubtractionForMorphisms",
   "AdditiveInverseForMorphisms" ], ~.IsEnrichedOverCommutativeRegularSemigroup ),
 
 IsAdditiveCategory := Concatenation( [
--- a/CAP/gap/ConstructiveCategoriesRecord.gi
+++ b/CAP/gap/ConstructiveCategoriesRecord.gi
@@ -19,7 +19,8 @@ IsAbCategory := Concatenation( [
   "ZeroMorphism", 
   "IsZeroForMorphisms", 
   "SubtractionForMorphisms",
-  "AdditiveInverseForMorphisms" ], ~.IsEnrichedOverCommutativeRegularSemigroup ),
+  "AdditiveInverseForMorphisms",
+  "MultiplyWithElementOfCommutativeRingForMorphisms" ], ~.IsEnrichedOverCommutativeRegularSemigroup ),
 
 IsAdditiveCategory := Concatenation( [
   "ZeroObject",
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2017.06.02", ## Mohamed's version
+  "2017.06.03", ## Mohamed's version
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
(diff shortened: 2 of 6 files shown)


CAP/gap/ConstructiveCategoriesRecord.gi
```diff
@@ -18,7 +18,9 @@ IsEnrichedOverCommutativeRegularSemigroup := Concatenation(
 IsAbCategory := Concatenation( [ 
   "ZeroMorphism", 
   "IsZeroForMorphisms", 
-  "AdditiveInverseForMorphisms" ], ~.IsEnrichedOverCommutativeRegularSemigroup ),
+  "SubtractionForMorphisms",
+  "AdditiveInverseForMorphisms",
+  "MultiplyWithElementOfCommutativeRingForMorphisms" ], ~.IsEnrichedOverCommutativeRegularSemigroup ),
```
sebasguts: @sebastianpos Is this correct?

sebasguts:
I do not understand how this command is
* general enough to be in the CAP-kernel
* well-defined from it's documentation/definition

How I see it, if it should be part of the CAP-kernel, there should be more explanation:
* Which commutative ring?
* What is it supposed to do with the morphism?
* Should I implement partial functions for several commutative rings?

I think this should be part of an additional package. Furthermore, I do not think `IsAbCategory` should depend on it.

mohamed-barakat:
What I need in the package `Bialgebroids` is the support for R-linear categories (where R is any constructive commutative ring R): The Hom-set is an R-module and the composition is R-bilinear. I added it to `~.IsEnrichedOverCommutativeRegularSemigroup` as the action of R can be generalized to  commutative regular semigroups.

Another example (still R commutative): A = LeftPresentations( R ) and its generalized morphisms category can then be endowed with this extra R-linear structure. This allows multiplying morphisms in A or G(A) with ring elements in R.

CAP/gap/ConstructiveCategoriesRecord.gi
sebastianpos: This list should contain prerequisites for being an Ab-category.
So "MultiplyWithElementOfCommutativeRingForMorphisms"
should not be in this list.

sebastianpos:
Multiplying generalized morphisms with scalars sounds interesting.
I assume you have in mind that you want to multiply the associated morphism
with the given scalar and leave the domain and codomain as they are?

This would turn the Hom-sets of G(A) into something like "R-modules having an underlying commutative regular semigroup instead of an abelian group". Do you know if there is a theory describing such a construction? If there is, then this would give the appropriate context
for your request.

CAP/gap/ConstructiveCategoriesRecord.gi
mohamed-barakat: I added `"MultiplyWithElementOfCommutativeRingForMorphisms"` to this list, as I found `"AdditiveInverseForMorphisms"` at the end of it, and commutative regular semigroups do not have additive inverses in the strict sense. Maybe I should create a new list with a name like `IsEnrichedOverRlinearCommutativeRegularSemigroup`.

mohamed-barakat:
Yes, this is what I have in mind.

I interpret the entry `"AdditiveInverseForMorphisms"` in the list `IsEnrichedOverCommutativeRegularSemigroup` as a first step in this direction, as it enables an action of the integers.

And regarding you question on a theory describing such a construction:

In our paper on Gabriel morphisms (https://arxiv.org/pdf/1409.2028.pdf) we write below Lemma A.2:

Hence, a commutative inverse semigroup H defines a direct system of Abelian groups ... over the meet-lattice E(H). In particular ...

On can either define what an R-linear commutative regular/inverse semigroup means (which is straightforward), or replace "Abelian groups" by "R-modules" in the above description.

CAP/gap/ConstructiveCategoriesRecord.gi
sebastianpos: Again, this is the list for Ab-categories.
Thus, `"AdditiveInverseForMorphisms"` is correctly inserted here.

sebastianpos:
It would be nice to have a name which is already used
somewhere in the literature.

CAP/gap/ConstructiveCategoriesRecord.gi
mohamed-barakat: Yes, sorry, I misinterpreted the piece of code I see above. This list `IsEnrichedOverCommutativeRegularSemigroup` is appended at the end.

mohamed-barakat:
I searched but was unable to find one.

PR #111, opened 2017-10-02
Declaring properties HasEnoughProjectives/Injectives for categories.

kamalsaleh:


sebastianpos:
I recommend calling these properties
`IsCategoryWithEnoughProjectives`
and
`IsCategoryWithEnoughInjectives`
in order to avoid the automatic declaration of
`HasHasEnoughProjectives`
and
`HasHasEnoughInjectives.`

sebasguts:
Also, those should be added to ConstructiveCategoryRecord.gi and the list in CAP.gd (although I am not sure if this list is still used somewhere)

sebasguts:
You can update a PR by (force) pushing new commits to the branch the PR was made from. Please consider this next time, instead of closing the PR, so we do not loose the discussion.

PR #113, opened 2017-10-17
Declaring two properties:

kamalsaleh:
* IsAbelianCategoryWithEnoughInjectives

for abelian categories.

### DIFF
--- a/CAP/gap/ConstructiveCategoriesRecord.gi
+++ b/CAP/gap/ConstructiveCategoriesRecord.gi
@@ -73,7 +73,17 @@ IsSymmetricClosedMonoidalCategory  := Concatenation( [
 IsRigidSymmetricClosedMonoidalCategory  := Concatenation( [ 
 "TensorProductInternalHomCompatibilityMorphismInverseWithGivenObjects",
 "MorphismFromBidualWithGivenBidual"
-], ~.IsSymmetricClosedMonoidalCategory )
+], ~.IsSymmetricClosedMonoidalCategory ),
+
+IsAbelianCategoryWithEnoughProjectives := Concatenation( [
+"EpimorphismFromSomeProjectiveObject",
+"ProjectiveLift"
+], ~.IsAbelianCategory ),
+
+IsAbelianCategoryWithEnoughInjectives := Concatenation( [
+"MonomorphismIntoSomeInjectiveObject",
+"InjectiveColift"
+], ~.IsAbelianCategory )
 
 ) );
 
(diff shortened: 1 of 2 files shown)


sebasguts:
This is the follow-up of #111

CAP/gap/ConstructiveCategoriesRecord.gi
```diff
@@ -73,7 +73,15 @@ IsSymmetricClosedMonoidalCategory  := Concatenation( [
 IsRigidSymmetricClosedMonoidalCategory  := Concatenation( [ 
 "TensorProductInternalHomCompatibilityMorphismInverseWithGivenObjects",
 "MorphismFromBidualWithGivenBidual"
-], ~.IsSymmetricClosedMonoidalCategory )
+], ~.IsSymmetricClosedMonoidalCategory ),
+
+IsAbelianCategoryWithEnoughProjectives := Concatenation( [
+"EpimorphismFromSomeProjectiveObject"
+], ~.IsAbelianCategory ),
```
sebasguts: What about the lift functions?

kamalsaleh: You are right, I will update the pull request.

CAP/gap/ConstructiveCategoriesRecord.gi
```diff
@@ -73,7 +73,17 @@ IsSymmetricClosedMonoidalCategory  := Concatenation( [
...
+], ~.IsSymmetricClosedMonoidalCategory ),
+
+IsAbelianCategoryWithEnoughProjectives := Concatenation( [
+"EpimorphismFromSomeProjectiveObject",
+"ProjectiveLift"
+], ~.IsAbelianCategory ),
+
+IsAbelianCategoryWithEnoughInjectives := Concatenation( [
+"MonomorphismIntoSomeInjectiveObject",
+"InjectiveColift"
```
sebasguts: Now the object functions are gone? It this on purpose?

sebastianpos: They are not necessary since they can be derived from
`MonomorphismIntoSomeInjectiveObject`
and
`EpimorphismFromSomeProjectiveObject`.
If I remember it correctly, then the record
`CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD`
was intended to point out only one particular way to implement
a given type of category, and not to enlist all possible functions
for such a type.

sebasguts: I think you are right, yes.

PR #116, opened 2017-12-01
Kernelcokernelfunctorial

sebastianpos:
I added two structural convenience methods, 
one for `KernelObjectFunctorialWithGivenKernelObjects,`
and one for `CokernelObjectFunctorialWithGivenCokernelObjects`.
They each accept a full square of morphisms as an input
and delegate to the method that only needs a non-complete square as input.
These functions will allow Mohamed to write more generical code.

### DIFF
--- a/CAP/gap/MonoidalCategoriesDerivedMethods.gi
+++ b/CAP/gap/MonoidalCategoriesDerivedMethods.gi
@@ -330,7 +330,7 @@ AddDerivationToCAP( InternalHomOnMorphismsWithGivenInternalHoms,
                      );
     
 end : CategoryFilter := IsRigidSymmetricClosedMonoidalCategory,
-      Description := "InternalHomOnMorphismsWithGivenInternalHoms using functorality of Dual and TensorProduct" );
+      Description := "InternalHomOnMorphismsWithGivenInternalHoms using functoriality of Dual and TensorProduct" );
 
 ##
 AddDerivationToCAP( MorphismFromBidualWithGivenBidual,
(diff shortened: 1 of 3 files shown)


mohamed-barakat:
Thanks :)

sebasguts:
The PR itself looks good.

sebastianpos:
Yes, that's what I meant.

mohamed-barakat:
Could you please make increase the version number, so that I can reference it as a dependancy in `FunctorCategories`?

mohamed-barakat:
Now that I am trying to use `FunctorCategories` with the new code I figured out that I need slightly different behavior, namely

`KernelObjectFunctorialWithGivenKernelObjects( IsCapCategoryObject, IsList, IsCapCategoryObject )`

and

`CokernelObjectFunctorialWithGivenCokernelObjects( IsCapCategoryObject, IsList, IsCapCategoryObject )`

where IsList is in both cases `[ alpha, mu, nu, alpha_prime ]`.

Sorry for the inconvenience. If you like I can make a PR.

sebastianpos:
Couldn't you simply unwrap the lists in your code?
It would still be generic in this way.

mohamed-barakat:
When I had the extra record entry `input_type` I could distinguish in my code for example between `DirectSum` and `KernelObject`. After deleting that entry I have no other possibility than to require an equal behavior of `DirectSumFunctorialWithGivenDirectSums` and `KernelObjectFunctorialWithGivenKernelObjects`. And since the former takes a diagram as a second argument I must rely on the latter behaving the same way.

sebastianpos:
Such a distinction still seems possible by looking at the filter_list entry.

mohamed-barakat:
You are right, I now check if the second entry in `filter_list` is equal to `IsList` and branch from there. Then my only remaining request would be to increase the version. Thanks again.

sebastianpos:
Done.

1ad16e2

PR #121, opened 2018-01-15
Propagate categorical properties for opposite cats

sebastianpos:
We haven't propagated the categorical properties between
a category and its opposite yet.
For doing this, I created a new global list
`CAP_INTERNAL_CATEGORICAL_PROPERTIES_LIST`
which encodes categorical properties together with their dual counterparts.


sebasguts:
Looks good.

PR #124, opened 2018-01-26
Added description tags to opposite property propagation lists.

### LINKED ISSUE #123: Wrong propagation of properties in opposite category
Even though we installed dual pairs of properties
(like `[ IsMonomorphism, IsEpimorphism ]`)
they are not propagated as duals.
Here is an example:

```
gap> LoadPackage( "LinearAlgebraForCAP" );
gap> QQ := HomalgFieldOfRationals();;
gap> alpha := VectorSpaceMorphism( VectorSpaceObject( 1, QQ ), [ [ 1, 2 ] ], VectorSpaceObject( 2, QQ ) );
<A morphism in Category of matrices over Q>
gap> IsMonomorphism( alpha );
true
gap> Opposite( alpha );
<A split monomorphism in Opposite of Category of matrices over Q>
```

sebasguts:
Closes #123

### DIFF
--- a/CAP/gap/OppositeCategory.gi
+++ b/CAP/gap/OppositeCategory.gi
@@ -65,6 +65,7 @@ BindGlobal( "CAP_INTERNAL_FIND_OPPOSITE_PROPERTY_PAIRS",
         else
             current_entry := [ current_rec.installation_name,
                                CAP_INTERNAL_METHOD_NAME_RECORD.( current_rec.dual_operation ).installation_name ];
+            current_entry := [ Concatenation( current_entry[ 1 ], " vs ", current_entry[ 2 ] ), current_entry ];
         fi;
         
         if CAP_INTERNAL_METHOD_NAME_RECORD.( current_recname ).property_of = "object" then
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -9,7 +9,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
-  "2018.01.09", ## Sebas' version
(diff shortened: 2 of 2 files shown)


CAP/gap/OppositeCategory.gi
```diff
@@ -65,6 +65,7 @@ BindGlobal( "CAP_INTERNAL_FIND_OPPOSITE_PROPERTY_PAIRS",
         else
             current_entry := [ current_rec.installation_name,
                                CAP_INTERNAL_METHOD_NAME_RECORD.( current_rec.dual_operation ).installation_name ];
+            current_entry := [ Concatenation( current_entry[ 1 ], " vs ", current_entry[ 1 ] ), current_entry ];
```
sebastianpos: Do you mean `current_entry[ 1 ], " vs ", current_entry[ 2 ] `?

sebastianpos: Could you explain the fix?
Does every ToDoList entry need a description?

sebasguts: >  Do you mean current_entry[ 1 ], " vs ", current_entry[ 2 ]?

Yes

> Could you explain the fix?
> Does every ToDoList entry need a description?

No, not in general, but in this case it requires one. The reason is the specification of the entries of the list passed as third argument of `ToDoListEntryToMaintainFollowingAttributes`:

If it is only one string, it is interpreted as attribute name for both source and target of the entry. If it is a list of two strings, the first string is interpreted as description of the entry, the second as attribute name for both source and target. The only way to add two different attributes right now is to add a description.

I think with a little refactoring of the logic of `ToDoListEntryToMaintainFollowingAttributes` we could make the entry a list of a list containing two strings. But for now, this works. Maybe I should post an issue to ToolsForHomalg?

PR #127, opened 2018-01-30
Freyd

sebastianpos:
Please check the changes in the opposite category.


CAP/gap/OppositeCategory.gi
```diff
@@ -243,22 +254,30 @@ BindGlobal( "CAP_INTERNAL_INSTALL_OPPOSITE_ADDS_FROM_CATEGORY",
...
-        func := create_func( dual_name );
+        if IsBound( current_entry.dual_arguments_reversed ) then
+            
+            func := create_func( dual_name, true );
+            
+        else
+            
+            func := create_func( dual_name );
+            
+        fi;
```
sebasguts: Should we add `false` as a default for `dual_arguments_reversed`? This would make the record entry more complete, and this if would disappear.

PR #131, opened 2018-01-31
Adjust constructors in ModulePresentationsForCAP

sebastianpos:


sebastianpos:
Good point!
It also applies to LinearAlgebraForCAP.

PR #133, opened 2018-02-01
Adjust constructors in GeneralizedMorphismsForCAP

sebastianpos:


sebastianpos:
I also did the same for `LinearAlgebraForCAP` and `ModulePresentationsForCAP.`

PR #135, opened 2018-02-07
Linearalgebra

sebastianpos:
I introduced two primitives for accessing the component of
a morphism from/into a direct sum.

### DIFF
--- a/CAP/gap/UniversalObjects.gd
+++ b/CAP/gap/UniversalObjects.gd
@@ -1423,6 +1423,66 @@ DeclareOperationWithCache( "MorphismBetweenDirectSumsOp",
                            [ IsList, IsInt, IsInt, IsCapCategoryMorphism ] );
 
 #! @Description
+#! The arguments are a morphism $\alpha: A \rightarrow S$,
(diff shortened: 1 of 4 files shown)


CAP/gap/UniversalObjects.gd
```diff
@@ -1422,6 +1422,66 @@ DeclareOperationWithCache( "MorphismBetweenDirectSums",
 DeclareOperationWithCache( "MorphismBetweenDirectSumsOp",
                            [ IsList, IsInt, IsInt, IsCapCategoryMorphism ] );
 
+#! @Description
```
sebasguts: Do those primitives violate the fact, that the diagram should be in the first place, i.e., should the object list be the first argument?

sebastianpos: I think we only have this convention for those basic operations concerning the data defining a limit (or colimit):
the limit object, the diagram data, the universal property data.
In `ComponentOfMorphismIntoDirectSum` and `ComponentOfMorphismFromDirectSum`  I like to think of the first argument as the most relevant one.

sebasguts: Ok.

PR #140, opened 2018-02-22
Make some installation names CAP specific

### LINKED ISSUE #139: Installation names should be CAP specific
The following CAP methods have
installation names that are too generic (in the sense
that they are widely used within other packages):

- `IsZeroForMorphisms` has installation name `IsZero`
- `AdditionForMorphisms` has installation name `+`
- `SubtractionForMorphisms` has installation name `-`
- `IsZeroForObjects` has installation name `IsZero`

This causes trouble when turning GAP objects
into CAP objects/morphisms that already have
methods for `IsZero` (or `+`, `-`) which are 
then probably ranked higher than the CAP-methods.

Is it a good idea to change the installation names
to the more specific CAP names in these cases?

sebastianpos:
closes #139 

Some comments:

- I also found `AdditiveInverse` as a function with the same problem
- `IsZero` was used heavily within our logic files, so I replaced it with the appropriate morphisms/objects version
- I don't know why but basic operations like `IsZeroForMorphisms` were not declared as properties. In particular, they did not have Tester, which was a problem for the ToDo lists.


sebastianpos:
Ok, let's discuss this issue on tuesday.
I also noticed that we used `IsZero`
within derivations, and that
`HasIsZero` is used by Mohamed
in [`IntrinsicCategories`](https://github.com/mohamed-barakat/IntrinsicCategories/blob/42daeb1c4e524d6bb79d6597e2df43d2f19f1e5e/gap/IntrinsicCategories.gi#L174)

PR #142, opened 2018-05-02
Topos stuff

mohamed-barakat:
1. Equalizer & Coequalizer with logic and documentation
2. AssociatorLeftToRightOfDirectProductsWithGivenDirectProducts (to view product as monoidal structure)


CAP/gap/UniversalObjects.gi
```diff
@@ -276,6 +276,73 @@ InstallMethod( DirectProductFunctorial,
     
 end );
 
+##
```
sebastianpos: The UniversalMorphism.gi file should only contain
convenient methods, and not methods that look like complete derivations.


I suggest to define `AssociatorRightToLeftOfDirectProductsWithGivenDirectProducts`
as a new CAP basic operation, and to install this particular method
as a derivation. The advantage is that `AssociatorRightToLeftOfDirectProductsWithGivenDirectProducts`
would then only be installed for categories that already can compute
 `ProjectionInFactorOfDirectProductWithGivenDirectProduct`,
`UniversalMorphismIntoDirectProductWithGivenDirectProduct`, etc.

CAP/gap/DerivedMethods.gi
```diff
@@ -154,6 +154,41 @@ AddWithGivenDerivationPairToCAP( UniversalMorphismIntoZeroObject,
...
+    
+    D := List( diagram, Source );
+    
+    diagram_of_equalizer := List( [ 1 .. Length( D ) ], i -> ProjectionInFactorOfDirectProduct( D, i ) );
+    
+    diagram_of_equalizer := List( [ 1 .. Length( D ) ], i -> PreCompose( diagram_of_equalizer[i], diagram[i] ) );
+    
+    chi := UniversalMorphismIntoDirectProduct( D, tau );
+    
+    return UniversalMorphismIntoEqualizer( diagram_of_equalizer, chi );
```
sebastianpos: Unfortunately, this derivation of `UniversalMorphismIntoFiberProduct` 
will only work properly in combination with your above given
derivation of `ProjectionInFactorOfFiberProduct`. Imagine that the user
gives his own `ProjectionInFactorOfFiberProduct`, then there is
no guarantuee that the user's fiber product object (which then equals the source of the user's fiber product projection)
coincides with the equalizer object of the diagram that you create here.

To address such interdependencies between derivations,
we introduced the final derivations and used them
for the parallel derivation in the abelian case
(fiber products derived from kernels and direct sums).


However, Sebastian Gutsche and I are currently thinking about
a more flexible way to deal with such interdependencies, so
you can either wait for our new solution to be implemented
or to create your derivation similar to the one in the abelian case.

CAP/gap/UniversalObjects.gd
```diff
@@ -2758,6 +3160,268 @@ DeclareOperation( "AddFiberProductFunctorialWithGivenFiberProducts",
...
+#! @Returns a morphism in $\mathrm{Hom}( A, \mathrm{Coequalizer}( D ) )$.
+#! @Arguments D,method_selection_morphism
+DeclareOperation( "ProjectionOntoCoequalizerOp",
+                  [ IsList, IsCapCategoryMorphism ] );
+
+#! @Description
+#! The arguments are a list of morphisms $D = ( \beta_i: B \rightarrow A )_{i = 1 \dots n}$,
+#! and an object $C = \mathrm{Coequalizer}(D)$.
+#! The output is the projection
+#! $\pi: A \rightarrow \mathrm{Coequalizer}( D )$.
```
sebastianpos: \mathrm{Coequalizer}( D ) -> C

CAP/gap/UniversalObjects.gd
```diff
@@ -2758,6 +3160,268 @@ DeclareOperation( "AddFiberProductFunctorialWithGivenFiberProducts",
...
+
+#! @Description
+#! This is a convenience method.
+#! The arguments are a list of morphisms $D = ( \beta_i: B \rightarrow A )_{i = 1 \dots n}$
+#! and a morphism $\tau: A \rightarrow T $ such that
+#! $\tau \circ \beta_i \sim_{B,T} \tau \circ \beta_j$ for all pairs $i,j$.
+#! The output is the morphism
+#! $u( \tau ): \mathrm{Coequalizer}(D) \rightarrow T$
+#! given by the universal property of the coequalizer.
+
```
sebastianpos: @juliamick 
Again, there is a missing declaration.

CAP/gap/UniversalObjects.gd
```diff
@@ -2758,6 +3160,268 @@ DeclareOperation( "AddFiberProductFunctorialWithGivenFiberProducts",
...
+#! @Returns a morphism in $\mathrm{Hom}( C, T )$
+#! @Arguments D, tau, C
+DeclareOperation( "UniversalMorphismFromCoequalizerWithGivenCoequalizer",
+                  [ IsList, IsCapCategoryMorphism, IsCapCategoryObject ] );
+
+## Add Operations
+
+#! @Description
+#! The arguments are a category $C$ and a function $F$.
+#! This operations adds the given function $F$
```
sebastianpos: This operations -> This operation

(actually, this typo occurs several times)

CAP/gap/UniversalObjects.gd
```diff
@@ -2758,6 +3160,268 @@ DeclareOperation( "AddFiberProductFunctorialWithGivenFiberProducts",
...
+DeclareOperation( "AddUniversalMorphismFromCoequalizerWithGivenCoequalizer",
+                  [ IsCapCategory, IsList, IsInt ] );
+
+DeclareOperation( "AddUniversalMorphismFromCoequalizerWithGivenCoequalizer",
+                  [ IsCapCategory, IsList ] );
+
+
+#! @Description
+#! The argument is a list
+#! $L = ( ( \beta_i: B \rightarrow A, \mu: A \rightarrow A', \beta_i': B' \rightarrow A' )_{i = 1 \dots n} )$
```
sebastianpos: is a list consisting of three entries:
`beta_i,` `mu,` `beta_i'`

CAP/gap/UniversalObjects.gd
```diff
@@ -2758,6 +3160,268 @@ DeclareOperation( "AddFiberProductFunctorialWithGivenFiberProducts",
...
+#! given by the functorality of the coequalizer.
+#! @Returns a morphism in $\mathrm{Hom}(\mathrm{Coequalizer}( ( \beta_i )_{i=1 \dots n} ), \mathrm{Coequalizer}( ( \beta_i' )_{i=1 \dots n} ))$
+#! @Arguments L
+DeclareOperation( "CoequalizerFunctorial",
+                  [ IsList ] );
+
+#! @Description
+#! The arguments are an object $s = \mathrm{Coequalizer}( ( \beta_i )_{i=1}^n )$,
+#! a list
+#! $L = ( ( \beta_i: B \rightarrow A, \mu: A \rightarrow A', \beta_i': B' \rightarrow A' )_{i = 1 \dots n} )$
```
sebastianpos: see above

CAP/gap/UniversalObjects.gd
```diff
@@ -2758,6 +3160,268 @@ DeclareOperation( "AddFiberProductFunctorialWithGivenFiberProducts",
...
+#! @Returns a morphism in $\mathrm{Hom}(s, r)$
+#! @Arguments s, L, r
+DeclareOperation( "CoequalizerFunctorialWithGivenCoequalizers",
+                  [ IsCapCategoryObject, IsList, IsCapCategoryObject ] );
+
+#! @Description
+#! The arguments are a category $C$ and a function $F$.
+#! This operations adds the given function $F$
+#! to the category for the basic operation <C>CoequalizerFunctorialWithGivenCoequalizers</C>.
+#! $F: ( \mathrm{Coequalizer}( ( \beta_i )_{i=1}^n ), ( \beta_i: B \rightarrow A, \mu: A \rightarrow A', \beta_i': B' \rightarrow A' )_{i = 1 \dots n}, \mathrm{Coequalizer}( ( \beta_i' )_{i=1}^n ) ) \mapsto (\mathrm{Coequalizer}( ( \beta_i )_{i=1}^n ) \rightarrow \mathrm{Coequalizer}( ( \beta_i' )_{i=1}^n ) )$
```
sebastianpos: see above

CAP/gap/UniversalObjects.gd
zickgraf: Since this typo is already present in existing CAP-code, I will create a separate pull request fixing this globally after this pull request is merged.

CAP/gap/UniversalObjects.gd
```diff
@@ -2317,6 +2317,319 @@ DeclareOperation( "AddDirectProductFunctorialWithGivenDirectProducts",
...
+## Add Operations
+
+#! @Description
+#! The arguments are a category $C$ and a function $F$.
+#! This operations adds the given function $F$
+#! to the category for the basic operation <C>Equalizer</C>.
+#! $F: ( (\beta_i: A \rightarrow B)_{i = 1 \dots n} ) \mapsto E$
+#! @Returns nothing
+#! @Arguments C, F
+DeclareOperation( "AddEqualizer",
```
sebasguts: Should we group those for the documentation?

zickgraf: I would propose to do this in a follow-up since similar pieces of code already exist in CAP.

sebasguts: All right.

PR #153, opened 2018-06-18
IsWellDefinedForObjects/Morphisms for Serre quotients

mohamed-barakat:


GeneralizedMorphismsForCAP/gap/SerreQuotients.gi
```diff
@@ -63,3 +63,33 @@ InstallMethod( LiftNaturalIsoFromIdToSomeToSerreQuotientCategory,
     return new_nat_trans;
     
 end );
+
+BindGlobal( "CAP_INTERNAL_INSTALL_OPERATIONS_FOR_SERRE_QUOTIENT",
```
sebasguts: Could you please use `DeclareGlobalFunction` and `InstallGlobalFunction` for this function, as this function is not intended to be used in the same `.gi` file only?

GeneralizedMorphismsForCAP/gap/SerreQuotients.gi
```diff
@@ -63,3 +63,33 @@ InstallMethod( LiftNaturalIsoFromIdToSomeToSerreQuotientCategory,
...
+    AddIsWellDefinedForMorphisms( category,
+      
+      function( morphism )
+        local generalized_morphism;
+        
+        generalized_morphism := UnderlyingGeneralizedMorphism( morphism );
+        
+        return
+          membership_function( Range( CodefectProjection( generalized_morphism ) ) ) and
+          membership_function( Source( DefectEmbedding( generalized_morphism ) ) );
```
sebasguts: @sebastianpos Even if this is probably irrelevant, would it make sense to have two versions of this function, since it could benefit from an already computed `CodefectProjection` or `DefectEmbedding`?

The problem is that this would actually force us (with the current setup) to install three functions.

sebastianpos: What do you mean?
`CodefectProjection` and `DefectEmbedding` are attributes and thus they are stored anyway.

sebasguts: If we install a version of this method that first checks `membership_function( Source( DefectEmbedding( generalized_morphism ) ) )` with the filter `HasDefectEmbedding`, it might be faster. Just a random thought.

sebastianpos: Ah, now I see.
So you mean glueing `IsWellDefinedForMorphisms`
by 3 functions:
One version with filter `HasDefectEmbedding`
that first does the membership test for the defect,
a second version with filter `HasCodefectProjection`
that first does the membership test for the codefect,
and a third version with no filter at all for which there
is no prefered ordering in the testing.

Is this what you meant?

sebasguts: Exactly.

PR #170, opened 2018-08-14
Supported linear categories by adding MultiplyWithElementOfCommutativeRing

mohamed-barakat:
Note: I do not have the ambition to act with the ring on a commutative regular semigroup.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gi
+++ b/CAP/gap/CategoryMorphisms.gi
@@ -158,6 +158,22 @@ InstallMethod( AdditiveInverse,
 AdditiveInverseForMorphisms );
 
 ##
+InstallMethod( \*,
+               [ IsRingElement, IsCapCategoryMorphism ],
+               
+MultiplyWithElementOfCommutativeRingForMorphisms );
+
+##
+InstallMethod( \*,
+               [ IsCapCategoryMorphism, IsRingElement ],
+               
+  function( mor, r )
+    
+    return MultiplyWithElementOfCommutativeRingForMorphisms( r, mor );
+    
+end );
+
+##
 InstallMethod( IsEqualForCacheForMorphisms,
                [ IsCapCategoryMorphism, IsCapCategoryMorphism ],
                
--- a/CAP/gap/MethodRecord.gi
+++ b/CAP/gap/MethodRecord.gi
@@ -1054,6 +1054,25 @@ SubtractionForMorphisms := rec(
   end,
   return_type := "morphism" ),
 
+MultiplyWithElementOfCommutativeRingForMorphisms := rec(
+  installation_name := "MultiplyWithElementOfCommutativeRingForMorphisms",
+  filter_list := [ IsRingElement, "morphism" ],
+  io_type := [ [ "r", "a" ], [ "a_source", "a_range" ] ],
+  cache_name := "MultiplyWithElementOfCommutativeRingForMorphisms",
+  
+  pre_function := function( r, morphism )
+    local value_1, value_2;
+    
+    if not r in CommutativeRingOfLinearCategory( CapCategory( morphism ) ) then
+      
+      return [ false, "the first argument is not an element of the ring of the category of the morphism" ];
+      
+    fi;
+    
+    return [ true ];
+  end,
+  return_type := "morphism" ),
+
 AdditiveInverseForMorphisms := rec(
   installation_name := "AdditiveInverseForMorphisms",
   filter_list := [ "morphism" ],
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2017.07.19", ## Mohamed's version
+  "2018.08.03", ## Mohamed's version
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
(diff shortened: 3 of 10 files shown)


sebasguts:
Is it necessary to put this into CAP directly, or could this be in a different package for LinearCategories?

mohamed-barakat:
Such a package would be at this stage minuscule. I would even like to enhance `IsAbCategory` with an operation `AddMultiplyWithElementOfZZForMorphisms` to be able to multiply morphisms with integers.

sebasguts:
Yeah, but I think at some stage we agreed to not put any more types of categories in the CAP kernel, and even thinking about removing *ALL* types of categories from the kernel and moving them into a separate package.

CAP/gap/CategoryMorphisms.gi
```diff
@@ -157,6 +157,12 @@ InstallMethod( AdditiveInverse,
                   
 AdditiveInverseForMorphisms );
 
+##
+InstallOtherMethod( \*,
```
sebasguts: Please do not overload the operators. If you use a predefined data structure in CAP, it is unclear which function is called by the operator and may cause inconsistencies or errors.

mohamed-barakat:
I agree, I made this small PR as I need it in `Bialgebroids` and can help later with splitting the CAP kernel.

CAP/gap/MethodRecord.gi
```diff
@@ -1054,6 +1054,13 @@ SubtractionForMorphisms := rec(
   end,
   return_type := "morphism" ),
 
+MultiplyWithElementOfCommutativeRingForMorphisms := rec(
+  installation_name := "MultiplyWithElementOfCommutativeRingForMorphisms",
+  filter_list := [ IsRingElement, "morphism" ],
```
sebastianpos: @sebasguts 
I think this is the first time where we have a filter that is not in:
IsList, object, morphism, IsInt
Can this cause problems?

sebastianpos: Moreover, a linear category over a commutative ring should
know over which commutative ring it is defined.
I suggest to declare an attribute like
`CommutativeRingOfLinearCategory`
for categories with the property
`IsLinearCategoryOverCommutativeRing`.

Having this ring available would enable us
to type-check the ring element in the pre-function
of the multiplication.

sebastianpos:
Can you elaborate on what you mean by:

> InstallOtherMethod( MultiplyWithElementOfCommutativeRingForMorphisms,
>                [ IsRingElement, IsCapCategoryMorphism ],
> 
> \* );
> 
> causes and error if \* was not declared for [ IsRingElement, IsCapCategoryMorphism ]
> and an annoying warning if it was declared.

I tried
`DeclareOperation( "\*", [ IsRingElement, IsCapCategoryMorphism ] );`
and
```
InstallMethod( \*,
               [ IsRingElement, IsCapCategoryMorphism ],
MultiplyWithElementOfCommutativeRingForMorphisms );
```
which worked without warnings for me.

mohamed-barakat:
This was the commit message of the old commit which might be outdated by now. I know see that it had a typo anyway, it should've been `InstallMethod` and not `InstallOtherMethod` back then.

CAP/gap/MethodRecord.gi
mohamed-barakat: I agree, in which file shall I do the declaration? `CAP.gd`?

mohamed-barakat:
Shall I keep the commit with this deprecated comment or shall I force-push a new commit with the explicit declaration?

mohamed-barakat:
Ah, you see the warning if you load `LinearAlgebraForCAP`.

CAP/gap/MethodRecord.gi
sebasguts: The filter does not provide any problem. You can set whatever filter you like there. However, only the keywords `"object"`, `"morphism"`, and `"twocell"` are replaced with the specific category filters.

CAP/gap/CategoryMorphisms.gi
sebasguts: Please note that this was the issue where we discussed this back then: https://github.com/homalg-project/CAP_project/issues/139 .

We can leave the overloaded operator, but please be aware that it might (and almost certainly will) cause problems.

sebastianpos:
@mohamed-barakat 
Could you clarify what you mean by "old commit"?

sebastianpos:
Yes, all CAP packages that declared `\*`
for `IsRingElement` and a (special type) of morphism
give that warning. But once `\*` is declared
in the CAP kernel, the correct action would
be to delete the declarations in the
other packages, and install the multiplication
via the CAP interface.

mohamed-barakat:
@sebastianpos: This commit is a slightly updated version of an old one which was part of PR #103.

mohamed-barakat:
@sebastianpos: OK, I'll take care of deleting the specific declarations in the remaining CAP project packages. In which file shall I place the declaration? I would've taken CAP.gd, OK?

mohamed-barakat:
Thanks for the your suggestions. I just pushed an amended commit.

mohamed-barakat:
Sorry, I oversaw your request to declare `CommutativeRingOfLinearCategory`. The next commit-update will fix this.

mohamed-barakat:
I added the attribute `CommutativeRingOfLinearCategory` and the requested pre_function.

sebastianpos:
@sebasguts 
Do you still agree with this commit?

sebasguts:
Yes :)

mohamed-barakat:
Thanks :)

PR #188, opened 2018-09-06
Sanity check for computed_value in ApplyFunctor

heiderich:
This adds a sanity check for `computed_value` in `ApplyFunctor`. Without these changes faulty object/morphism functions might lead to errors when

`Add( AsCapCategory( Range( functor ) ), computed_value );`

is executed later on, which would throw an error that might not be very helpful for users.

### DIFF
--- a/CAP/gap/CategoriesCategory.gi
+++ b/CAP/gap/CategoriesCategory.gi
@@ -433,12 +433,17 @@ InstallGlobalFunction( ApplyFunctor,
         computed_value := CallFuncList( FunctorObjectOperation( functor ), arguments );
 
         if range_category!.output_sanity_check_level > 0 then
+            if not IsCapCategoryObject( computed_value ) then
+                Error( Concatenation("the result of the object function of the functor \"", Name(functor), "\" does not lie in the IsCapCategoryObject filter." ) );
+            fi;
             if HasCapCategory( computed_value ) then
                 if not IsIdenticalObj( CapCategory( computed_value ), range_category ) then
                     Error( Concatenation( "the category of the result of the object function of the functor \"", Name(functor), "\" does not coincide with the range of this functor" ) );
                 fi;
             else
-                Error( Concatenation("the result of the object function of the functor \"", Name(functor), "\" does not have a CAP category" ) );
+                if not range_category!.add_primitive_output then
+                    Error( Concatenation("the result of the object function of the functor \"", Name(functor), "\" does not have a CAP category" ) );
+                fi;
             fi;
         fi;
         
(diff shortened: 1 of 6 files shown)


sebastianpos:
Up till now, the philosophy of methods added to a functor is the
same as the philosophy of methods added as a basic operation to a category:
it is the user who is responsible that his/her methods match the specifications
dictated by the types of functors/basic operations.

However, I am open for the option that (as a debugging tool),
one could be able to apply a sanity check to both the
output of functors and the output of basic operations. I will open an issue for that.

sebastianpos:
I created issue #191 addressing the general question of sanity checks for outputs.

sebasguts:
This would check the output type for a call to `ApplyFunctor`, but not for applying an installed functor on objects or morphisms. In this sense, one needs to write a wrapper around the given functions to a functor before they get installed for the functor operation.

@heiderich Do you want to do this, or should we continue to discuss this in #191 and close this PR for now?

sebasguts:
Can you please rebase this on the current master branch?

CAP/gap/CategoriesCategory.gi
```diff
@@ -426,6 +426,8 @@ InstallGlobalFunction( ApplyFunctor,
                 if not IsIdenticalObj( CapCategory( computed_value ), range_category ) then
                     Error( Concatenation( "the category of the result of the object function of the functor \"", Name(functor), "\" does not coincide with the range of this functor" ) );
                 fi;
+            else
+                Error( Concatenation("The result of the object function of the functor \"", Name(functor), "\" does not have a CAP category" ) );
```
zickgraf: If I have understood the concept correctly, the case that `computed_value` has no CAP category is supported explicitly, and that is why `Add` is called later. I think we should test for `IsCapCategoryObject( computed_value ) = true` here.

Please correct me if I'm wrong.

sebastianpos: Yes, Fabian is right.
Please test for the filter `IsCapCategoryObject`
and analogously for the filter `IsCapCategoryMorphism` below.

zickgraf:
@heiderich could you rebase this on master to fix the CI failure?

sebasguts:
@heiderich This looks good. Would you be against in squashing and merging? If yes, I will merge it as is.

heiderich:
@sebasguts No, I am not at all against squashing. Do you want me to do it and force-push the squashed commit? Otherwise, feel free to do it yourself.

sebasguts:
@heiderich No need, github does it for me.

Issue #162, opened 2018-07-06
AddDerivationAsLimitToCAP


zickgraf:
This is a sub-issue of #158.

I propose a function `AddDerivationAsLimitToCAP` to derive a universal object as a limit with the following specification:

## Arguments
1. the method `ObjectMethod` for the object which shall be derived, e.g. `FiberProduct` or `KernelObject`
2. the method `ProjectionMethod` for the "projections" which shall be derived, e.g. `ProjectionInFactorOfFiberProduct` or `KernelEmbedding`
3. the method `UniversalMorphismMethod` for the universal morphism which shall be derived, e.g. `UniversalMorphismIntoFiberProduct` or `KernelLift`
4. a function `DiagramGenerator` transforming the input of 1. into a diagram, e.g. for the fiber product the usual diagram and for the kernel the diagram consisting of two parallel arrows: the morphism \alpha passed to `KernelObject` and the zero morphism

## Pseudo-code
add three derivations to CAP (in a consistent way!):

**derive `ObjectMethod(args)` as follows:**

let `D = DiagramGenerator(args)`
return `LimitObject(D)`

**derive `ProjectionMethod(args)` as follows (this is hacky, see note below):**

if the number of arguments of `ObjectMethod` and `ProjectionMethod` is the same (check this via the method record)
let `projection_number = 1`
else
one element of `args` is an integer; let `projection_number` be this integer and remove this element from `args`

let `D = DiagramGenerator(args)`
return `ProjectionOfLimitIntoNode(D, projection_number)`

**derive `UniversalMorphismMethod(args)` as follows:**

let `tau` be the last element of `args` and remove it from `args`
if `tau` is a morphism, replace it by a 'suitable' list also named `tau` which can be used below (→to discuss)
let `D = DiagramGenerator(args)`
return `UniversalMorphismIntoLimit(D, tau)`

## Note
The hackiness in the second derivation could possibly be reduced by storing more information about the arguments of `ProjectionMethod` in the method record.

zickgraf:
This is made obsolete by PR #189 and the PRs which will follow it.

Issue #158, opened 2018-07-06
[META] Derivations of universal objects as (co-)limits


zickgraf:
Note: In the following I will only talk about limits. Of course, colimits are implied analogously. Also, this issue is work-in-progress and will be expanded and made more precise over time.

I would like to develop a system to simplify the derivation of universal objects as limits, e.g. the derivation of `FiberProduct` in https://github.com/homalg-project/CAP_project/pull/142. Concretely, this system should allow the user to tell CAP

"The FiberProduct can be derived as the limit over 'this' diagram."

with as few lines of code as possible. For this, the following things are needed:

1. a data structure for finite diagrams
2. an interface for computing limits of these diagrams (e.g. methods named `LimitObject`, `ProjectionOfLimitIntoNode` and `UniversalMorphismIntoLimit` (or similar))
3. the (consistent!) derivation of the methods in 2. by using products and equalizers (the "consistent" part should be covered by https://github.com/homalg-project/CAP_project/issues/144)
4. a function called `AddDerivationAsLimitToCAP` (or similar), which takes the methods to be derived (e.g. `FiberProduct`, `ProjectionInFactorOfFiberProduct`, `UniversalMorphismIntoFiberProduct`) and a function (or multiple functions) for converting the input of these methods to diagrams which in turn can be passed to the methods in 2.

I will open separate issues for the steps outlined above to discuss the ideas in detail, and update this meta-issue accordingly.

zickgraf:
I have opened separate issues for the steps above and would like to discuss my ideas in particular with regard to feasibility and possible problems with internals like caching which I am probably not aware of.

Issue #187, opened 2018-09-03
UniversalMorphismFromInitialObjectWithGivenInitialObject has wrong io_type


zickgraf:
Currently, `UniversalMorphismFromInitialObjectWithGivenInitialObject` has `io_type = [ [ "I", "A" ], [ "I", "A" ] ]`:

https://github.com/homalg-project/CAP_project/blob/ae6103ffcbfaa34e91e5b2068a2f95eb1edc47e6/CAP/gap/MethodRecord.gi#L615-L622

I think this should be `[ [ "A", "I" ], [ "I", "A" ] ]`. If you agree, I will provide a PR.

zickgraf:
This is fixed by #189.

PR #189, opened 2018-09-06
Validate limits in the method record

zickgraf:
I have written a function `CAP_INTERNAL_ADD_LIMIT_COLIMIT_PAIR` which takes a specification of a diagram shape as an input and automatically generates the entries of the method record for limits of diagrams of this shape (and for colimits of diagrams of the opposite shape). E.g.,
```
CAP_INTERNAL_ADD_LIMIT_COLIMIT_PAIR( [ "fixedobject", "varobject" ], [ [ 2, "varmorphism", 1 ] ], "FiberProduct", "ProjectionInFactorOfFiberProduct", "UniversalMorphismIntoFiberProduct", "Pushout", "InjectionOfCofactorOfPushout", "UniversalMorphismFromPushout" );
```
generates the entries `FiberProduct`, `ProjectionInFactorOfFiberProduct` and `UniversalMorphismIntoFiberProduct` (including pre functions), the `WithGiven` versions, the functorial, and the corresponding duals. Currently, this works for `DirectProduct`, `DirectSum`, `FiberProduct`, `Equalizer`, `KernelObject`, `TerminalObject`, `InitialObject` and `ZeroObject` (and the corresponding projections, universal morphisms etc.).

I started this as a tool for making the method record and, in particular, the pre functions (more) consistent, but now I think it could be a useful addition to CAP. If you agree, I have a few questions:

1. Should this work like a preprocessor, i.e. an external tool which generates code which in turn is copy-pasted into the method record, or should it be part of CAP and generate the entries of the method record when CAP is loaded? I would prefer the second option.
2. I have not yet compared the performance of the generated pre functions to the performance of the hand-written ones. Is a "small" performance hit acceptable or is it a strict requirement that the generated pre functions are not slower than hand-written ones?
3. What would be the correct place for documenting internal stuff like this (for example the diagram specification and parsing rules)?

Of course, I also want to use this for #158. In particular, the function `DiagramGenerator` proposed in #162 can now be automatically generated and the "hackiness" in #162 is solved.


sebastianpos:
This looks very impressive and I need more understanding :-)

Concerning your third question:
We have a chapter in the documentation of the internals of
Add functions:
https://github.com/homalg-project/CAP_project/blob/master/CAP/doc/AddFunctions.autodoc
I think this might be a good place for your documentation.
In particular, I am interested in the documentation of
the command `CAP_INTERNAL_ADD_LIMIT_COLIMIT_PAIR`
since I find it difficult to interpret its first two arguments.

Maybe you could write a small section on this command. I think this
will make it easier for me to see your overall plan and it will facilitate
the review.

zickgraf:
I have decided to split this PR into multiple PRs to simplify discussion and review. Now, this PR only includes the validation of the method record. The addition of prefunctions etc. will go into separate PRs. Note that this PR depends on https://github.com/homalg-project/CAP_project/pull/206 and https://github.com/homalg-project/CAP_project/pull/212, i.e. only the last commit is relevant.

I have changed the structure in comparison to the first version for the following reason:

I want to use the information about limits for three things:

1. Validate the entries of the method record.
2. Generate prefunctions automatically.
3. Generate derivations automatically.

These three things cannot be done at the same time (prefunctions have to be added before enhancing the method record, derivations have to be added after the Adds are installed, if I understand correctly). Thus, the information about limits is now contained in the global list `CAP_INTERNAL_METHOD_NAME_RECORD_LIMITS`. The entries can be enhanced (similar to the method record) by calling the global function `CAP_INTERNAL_ENHANCE_NAME_RECORD_LIMITS`. The validation is done by the global function `CAP_INTERNAL_VALIDATE_LIMITS_IN_NAME_RECORD`.


Note 1: I have not implemented @sebasguts's suggestion of attaching the information about limits directly to the method record since I am not sure what the right way to do this is. I think putting it under `CAP_INTERNAL_METHOD_NAME_RECORD.limits` is not possible since then a method `limits` would be expected.

Note 2: In the documentation I overload some notions with multiple meanings. For example, an object can be a concrete object in a diagram but also a string "varobject" which actually stands for mutiple objects. This is done intentionally to better convey the basic idea, but if it is confusing I can of course make this more precise.

All in all, this is now ready for review.

zickgraf:
Note that this PR includes the proposed fix in #187.

zickgraf:
I accidentally did a force push here. Please ignore this, I have force-pushed again to the correct commit 4e5d07f.

PR #194, opened 2018-09-13
Preliminary support for cartesian closed and cocartesian coclosed categories

mohamed-barakat:
This will probably be outsourced to a package `Toposes` in the future, but I had to do this now since I am starting to use this code in two different independent packages `FinSetsForCAP` and `Locales`.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2018.08.03", ## Mohamed's version
+  "2018.09.02", ## Mohamed's version
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
--- a/CAP/gap/CAP.gd
+++ b/CAP/gap/CAP.gd
@@ -59,6 +59,8 @@ InstallValue( CAP_INTERNAL_CATEGORICAL_PROPERTIES_LIST,
     [ "IsSymmetricMonoidalCategory" ],
     [ "IsSymmetricClosedMonoidalCategory" ],
     [ "IsRigidSymmetricClosedMonoidalCategory" ],
+    [ "IsCartesianClosedCategory" ],
+    [ "IsCoCartesianCoclosedCategory" ],
     [ "IsStrictMonoidalCategory" ],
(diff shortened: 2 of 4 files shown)


sebastianpos:
There are several non-trivial issues with this pull request.

So far, CAP only supports one tensor product on a category.
Treating direct products/coproducts as a tensor structure
can be tacitly done by simply using the commands for products/coproducts
and some additional basic operations (like `AssociatorLeftToRightOfDirectProductsWithGivenDirectProducts`).

The issues:
1) The overall idea of this pull request seems to be that lying in the filter `IsCartesianClosedCategory`
shall mean that the CAP-tensor structure commands shall be overwritten with the direct product commands. For me, this would be quite unexpected when I am dealing
with a category that has both a tensor structure different from direct products,
but that nevertheless has direct products as an additional structure.
2) Even if we accepted that `IsCartesianClosedCategory` shall mean that we interpret the CAP-tensor structure commands by direct products, then the derivations would need a primitive like
`IsomorphismFromDirectProductToTensorProduct` in order to work coherently.

What exactly do you need from the CAP-tensor structure that isn't already available
for direct products? Once we know that, we can probably introduce primitives in the spirit of
`AssociatorLeftToRightOfDirectProductsWithGivenDirectProducts`.

mohamed-barakat:
Thanks for the thorough reply, and I see your point. What I need is really the **closedness** w.r.t. the product (and co-closedness w.r.t. the coproduct, but I suggest to leave the coproduct aside for now). So what I really need is a right adjoint internal Hom (Dual, ...). So one can say that I was lazy and wanted to avoid copying `*InternalHom*` & `*Dual*` to something like `*InternalHomRightAdjointToDirectProduct*` & `*DualWRTDirectProduct*` (I can of course do this if you prefer). So naively speaking, if the direct product has a right adjoint then I would expect it to be _the_ `InternalHom` of the category and hence the `DirectProduct` to be _the_ `TensorProduct` of this category. My naiveness is due to my unawareness of "relevant" categories which are cartesian closed and simultaneously monoidal closed w.r.t. another tensor product.

mohamed-barakat:
After discussions with @sebastianpos:

1. start supporting multiple (closed) monoidal structures, in particular, I will not overload `TensorProduct` and `InternalHom` depending on filters like `IsCartesianClosedCategory`.
1. start a package `Toposes` to incorporate cartesian closed (and co-cartesian co-closed categories).

PR #213, opened 2018-09-24
minor fix in docu

mohamed-barakat:
Hom( I_k, Pushout( D ) ) -> Hom( I_k, I )

compare with docu of ProjectionInFactorOfFiberProductWithGivenFiberProduct

### DIFF
--- a/CAP/gap/UniversalObjects.gd
+++ b/CAP/gap/UniversalObjects.gd
@@ -3664,8 +3664,8 @@ DeclareOperation( "InjectionOfCofactorOfPushoutOp",
 #! an integer $k$,
 #! and an object $I = \mathrm{Pushout}(D)$.
 #! The output is the $k$-th injection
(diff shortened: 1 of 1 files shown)


CAP/gap/UniversalObjects.gd
```diff
@@ -3665,7 +3665,7 @@ DeclareOperation( "InjectionOfCofactorOfPushoutOp",
 #! and an object $I = \mathrm{Pushout}(D)$.
 #! The output is the $k$-th injection
 #! $\iota_k: I_k \rightarrow \mathrm{Pushout}( D )$.
```
zickgraf: I think `\mathrm{Pushout}( D )` should also be replace by `I` here.

mohamed-barakat:
@zickgraf: Thanks, fixed, amended and force-pushed.

mohamed-barakat:
Yes, I agree. However, I would as a user expect some difference in the documentation honouring the suffix `WithGiven`.

PR #217, opened 2018-10-08
generic IsEqualForMorphisms and IsCongruentForMorphisms

heiderich:
These methods are meant to provide helpful error messages if called with morphisms belonging to different CAP categories or, which should normally not happen, no specific methods `IsEqualForMorphisms` and `IsCongruentForMorphisms` are installed.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gi
+++ b/CAP/gap/CategoryMorphisms.gi
@@ -223,6 +223,34 @@ end );
 ##
 ######################################
 
+# This method should usually not be selected when the two morphisms belong to the same category
+InstallMethod( IsEqualForMorphisms,
+                [ IsCapCategoryMorphism, IsCapCategoryMorphism ],
+
+  function( morphism_1, morphism_2 )
+    
+    if not IsIdenticalObj( CapCategory( morphism_1 ), CapCategory( morphism_2 ) ) then
+        Error( Concatenation( "the morphism \"", String( morphism_1 ), "\" and the morphism \"", String( morphism_2 ), "\" do not belong to the same CAP category" ) );
+    else
+        Error( Concatenation( "the morphism \"", String( morphism_1 ), "\" and the morphism \"", String( morphism_2 ), "\" belong to the same CAP category, but no specific method IsEqualForMorphisms is installed" ) );
+    fi;
+    
+end );
+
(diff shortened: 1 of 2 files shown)


sebastianpos:
Yes, but only after the call of `Finalize`.

However, I think the main point of this pull request
is to throw a meaningful error when you call 
`IsEqualForMorphisms` or `IsCongruentForMorphisms`
on morphisms of different categories.

sebasguts:
I am not sure about this. `No method found` is kind of what you would expect, because this is what happens with the other CAP commands as well.

We can merge it, no problem, but then those command will behave a little different from, say, `IsEqualForObjects`. Can it be more sensible to just return `fail` instead of throwing the second error?

`fail` indicates in our context that it cannot decide equality, which should be fine in this case.

sebastianpos:
The specifications of `IsEqualForMorphisms` or `IsCongruentForMorphisms`
say that they are applicable only if the 2 input morphisms lie in the same category.
Thus, throwing an error would be the consistent way to handle this second case,
not returning `fail`. The whole point in my opinion is simply to alter the current
error message to a more meaningful one.

One could actually do the same thing for `IsEqualForObjects` as well.

sebasguts:
Yes, I agree for the first error message.

The question is if it is sensible to return the second (No specific installed) into just `return fail;`, because it matches the behaviour after finalization.

sebastianpos:
Ah, now I see.
I think you mean you want to compare by 
`IsIdenticalObj` and if this comparision is `false`, then `return fail`
(which is the behaviour of `IsEqualForMorphisms` `IsCongruentForMorphisms`
after finalization).

I don't know if this would be a good idea.
Since the second case can only be reached before finalization,
I think it is better to remind the user of the fact that he has to
finalize the category before actually calling CAP-operations.

sebasguts:
Okay, should we then put this in the error message, something like: "Maybe you forgot to finalize the category?"

sebastianpos:
Yes, I agree.

sebasguts:
Okay.

@heiderich , are you going to do that change or should we do it?

sebasguts:
If you are on it, you could also add a similar function for `IsEqualForObjects`

heiderich:
@sebasguts I will do both.

sebasguts:
All right, thanks a lot!

PR #225, opened 2018-10-16
Derivation adjustments

sebasguts:
@sebastianpos Please have a look if those commits solve your problems

### DIFF
--- a/CAP/gap/Derivations.gi
+++ b/CAP/gap/Derivations.gi
@@ -85,7 +85,13 @@ InstallMethod( IsApplicableToCategory,
 function( d, C )
   local filter;
   filter := CategoryFilter( d );
-  return Tester( filter )( C ) and filter( C );
+  if IsFilter( filter ) then
+      return Tester( filter )( C ) and filter( C );
+  elif IsFunction( filter ) then
+      return filter( C );
+  else
+      Error( "Category filter is not a filter or function" );
+  fi;
 end );
 
 InstallMethod( InstallDerivationForCategory,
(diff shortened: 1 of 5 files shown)


CAP/gap/Derivations.gi
```diff
@@ -348,28 +358,32 @@ InstallMethod( AddDerivationPair,
...
-    for current_implementation in Concatenation( implementations_with_extra_filters1, implementations_with_extra_filters2 ) do
-        
-        current_list := CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION( current_implementation[ 1 ], operations_in_graph, loop_multiplier );
-        current_list := List( current_list, i -> [ ValueGlobal( i[ 1 ] ), i[ 2 ] ] );
-        ## TODO: Maybe move the next line to AddDerivationPairToCAP
-        current_list := Concatenation( current_list, CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION_FOR_MONOIDAL_CATEGORIES( current_implementation[ 1 ], loop_multiplier ) );
-        collected_list := CAP_INTERNAL_MERGE_PRECONDITIONS_LIST( collected_list, current_list );
-        
-    od;
+    if preconditions_complete = false then
```
sebastianpos: Your documentation above says that every other value
than `true` will be interpreted as non-complete, not only
`false.`

sebasguts: Thank you, yes. I changed the logic again afterwards. I will fix up the documentation.

CAP/gap/Finalize.gi
```diff
@@ -162,8 +165,14 @@ InstallMethod( IsFinalized,
             if ForAll( current_final_derivation.can_compute, j -> CurrentOperationWeight( weight_list, NameFunction( j[ 1 ] ) ) < infinity ) and
               ForAll( current_final_derivation.cannot_compute, j -> CurrentOperationWeight( weight_list, NameFunction( j ) ) = infinity ) then
                 
-                Add( current_installs, i );
-                
+                current_tester_func := current_final_derivation.category_filter;
+
+                if IsFilter( current_tester_func ) and Tester( current_tester_func )( category ) and current_tester_func( category ) then
```
sebastianpos: Is this also a bugfix of the previous version?
It appears to be the case that final derivations where tried to be executed
regardless of the filters, is that correct?

sebasguts: Indeed, that is the case. Do you think this can cause problems? It is indeed a silent bugfix ;)

sebastianpos:
There is still the call of the function 
`CAP_INTERNAL_DERIVATION_SANITY_CHECK( graph, derivation );`
in `Derivations.gi`
that prevents me from using CAP-operations in derivations that
are not listed as precoditions.

sebasguts:
Ah, thanks for pointing that out. I will disable that one.

sebasguts:
Fixed

PR #228, opened 2018-11-20
Only install Browse function after Browse is loaded completely

sebasguts:


sebastianpos:
It seems that you forgot to add the file
`ToolsForCategories_AfterLoading.gi`

sebasguts:
Indeed, added it.

PR #231, opened 2018-12-18
IsRigidSymmetricClosedMonoidalCategory -> IsSymmetricClosedMonoidalCaIsSymmetricClosedMonoidalCategory

mohamed-barakat:
relaxed unnecessarily strict condition in derivation

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2018.09.17", ## Mohamed's version
+  "2018.09.20", ## Mohamed's version
   ## this line prevents merge conflicts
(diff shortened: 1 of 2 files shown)


sebastianpos:
I'm not sure about this. Did you come across an example of a symmetric closed monoidal category
that is not rigid, but for which this isomorphism exists?

mohamed-barakat:
Sorry, I probably read `IsomorphismFromInternalHomToTensorProduct` as `IsomorphismFromInternalHomToDual`.

Your argument is right and also applies to the other derivation of `InternalHomOnObjects` using `IsomorphismFromTensorProductToInternalHom`. So  I replaced `IsSymmetricClosedMonoidalCategory` by `IsRigidSymmetricClosedMonoidalCategory`, amended and force-pushed.

mohamed-barakat:
In any case both derivations for `InteralHomOnObjects` should have the same prerequisites, which is without this commit not the case.

PR #238, opened 2019-01-07
Added hook for primitive replacements

sebasguts:
in check for preconditions for derived methods.
This can be used to have similar with given replacements
as for monoidal categories in other categories without having to add
a special check function


sebastianpos:
I'm not sure if I get the idea.
Will I have to call
`CAP_INTERNAL_ADD_REPLACEMENTS_FOR_METHOD_RECORD`
on all methods records that are provided by additional packages
in order to make the "symbol-appearance-counting" work properly?

sebasguts:
No. Just for such additional records where currently the only one is provided in `MonodialCategories.gi`

Before now, we call an extra function for precisely this record, and if a second record would appear, we would have to do another function, and so on. This approach on the other hand is scalable.

sebastianpos:
I don't understand.
In [FreydCategoriesForCAP](https://github.com/homalg-project/CAP_project/blob/master/FreydCategoriesForCAP/gap/FreydCategoriesForCAP.gi), I also created an addition to the method record. Why is the one for monoidal categories different?

sebasguts:
In MonoidalCategories, there are [these](https://github.com/homalg-project/CAP_project/blob/master/CAP/gap/MonoidalCategories.gi#L363) "fake not-with-given-methods", which can be used in derivations. The additional record is just to detect dependencies on those non-MethodRecord methods in derivations.

CAP/gap/Derivations.gi
```diff
@@ -285,10 +285,7 @@ InstallMethod( AddDerivation,
     if preconditions_complete = false then
         for current_implementation in implementations_with_extra_filters do
 
-            current_list := CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION( current_implementation[ 1 ], operations_in_graph, loop_multiplier );
-            current_list := List( current_list, i -> [ ValueGlobal( i[ 1 ] ), i[ 2 ] ] );
-            ## TODO: Maybe move the next line to AddDerivationToCAP
-            current_list := Concatenation( current_list, CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION_FOR_MONOIDAL_CATEGORIES( current_implementation[ 1 ], loop_multiplier ) );
+            current_list := CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION( current_implementation[ 1 ], operations_in_graph, loop_multiplier, CAP_INTERNAL_METHOD_RECORD_REPLACEMENTS );
```
sebastianpos: Since the function
`CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION`
now takes an additional argument as its input, we need to adjust its documentation.

sebasguts: Will do. I did not realize it has a documentation at all.

mohamed-barakat:
This merge breaks some functionality in CAP which was not detected by the test suite:

```
LoadPackage( "LinearAlgebraForCAP" );;
A := MatrixCategory( HomalgFieldOfRationals( ) );;
CanCompute( A, "InternalHomOnObjects" );
false
```

sebasguts:
Will investigate this.

mohamed-barakat:
Thanks 👍

zickgraf:
Even with the fix above, this breaks prepare functions:

Before the merge:
```
gap> CAP_PREPARE_FUNCTION_RECORD.ProjectionInFactorOfBinaryDirectProductToProjectionInFactorOfDirectProduct[3];
[ "IdentityMorphism", "PreCompose", "DirectProduct" ]
```

After the merge:
```
gap> CAP_PREPARE_FUNCTION_RECORD.ProjectionInFactorOfBinaryDirectProductToProjectionInFactorOfDirectProduct[3];
[ <Attribute "IdentityMorphism">, <Operation "PreCompose">, "DirectProduct" ]
```

This leads to an error because in https://github.com/homalg-project/CAP_project/blob/368af10fbd85dee17d1e1fb1ec5f2691e2243f82/CAP/gap/PrepareFunctions.gi#L43 CAP calls `CanCompute` where `current_operation` is an element of the lists above. This fails if `current_operation` is not a string.

As an example, you can try to load the package `FinSetsForCAP`.

sebasguts:
I see.
The decision to let the function return objects instead of strings was because we have written down objects instead of strings in this list here: https://github.com/homalg-project/CAP_project/blob/master/CAP/gap/MonoidalCategories.gi#L364

@sebastianpos Would you be against making those object strings? Then I could make the prepare function return strings again. Converting from objects to strings is not really safe. We could do it, but it can lead to errors.

PR #239, opened 2019-01-07
Adding random methods + some other tiny fixes

kamalsaleh:


sebasguts:
Would it be a lot of work to split this PR into smaller chunks that belong together? It would be easier to review that way. If not, I can also review it as a whole, but it might take some more days.

kamalsaleh:
I don't really think the PR needs splittings since it does not contain a lot of ideas. It contains basically 3 things.
1- Random methods which allow user to create objects and morphism with an integer parameter, which should reflect the complexity of the object or morphism created. These methods are implemented for category of (graded) module presentations and vector spaces.
2- Fixing a/the bug in the opposite category.
3- Adding some methods to the category of graded presentations which can not be derived from other methods.

CAP/gap/MethodRecord.gi
```diff
@@ -2952,6 +2952,52 @@ SolveLinearSystemInAbCategory := rec(
...
+  return_type := "morphism",
+  post_function := function( object, n, return_value )
+    if not IsEqualForObjects( object, Range( return_value ) ) then
+      Error( "The range of the returned morphism is not equal to the given object" );
+    fi;
+  end 
+),
+
+RandomMorphismWithFixedSourceAndRange := rec(
+  installation_name := "RandomMorphismWithFixedRange",
```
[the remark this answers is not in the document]
kamalsaleh: In the current commits in graded and non-graded case, a function TRY_TO_ENHANCE_HOMALG_(GRADED)_RING_WITH_RANDOM_FUNCTIONS
will be called in the category constructor method.

The output is true if it succeed to add the random functions to the ring and false otherwise and only in the affirmative case the random methods will be added to category.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
```diff
@@ -2146,3 +2172,296 @@ end );
...
+        function( category, n )
+        local homalg_ring, left_presentations, M;
+
+        homalg_ring := category!.ring_for_representation_category;
+
+        left_presentations := LeftPresentations( homalg_ring );
+
+        M := RandomObject( left_presentations, n );
+
+        return AsRightPresentation( Involution( UnderlyingMatrix( M ) ) );
```
sebastianpos: Suggestion: instead of deriving the random right object from the random left one (which is a bit strange since all you need is an arbitrary random matrix), write a single `ADD_RANDOM_OBJECT` function. This function should have a boolean `left_or_right` as an extra argument (indicating if it is meant to be used for left or right modules), and can call the function you wrote for RandomObject in the left modules case, except that you call the constructor `AsLeftOrRightPresentation( mat, left_or_right)` in the end.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
```diff
@@ -2146,3 +2172,296 @@ end );
...
+    function( category )
+      local homalg_ring;
+
+      homalg_ring := category!.ring_for_representation_category;
+
+      AddRandomMorphismWithFixedSource( category,
+        function( M, n )
+          local m, y, syz, x, u, U, e;
+          
+          if n = 0 then
```
sebastianpos: Why do you return the morphism in the zero object in this case?
The function should work fine also for `n=0` and return a well defined morphism into an object with 0-relations.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
```diff
@@ -2146,3 +2172,296 @@ end );
...
+    function( category )
+      local homalg_ring;
+
+      homalg_ring := category!.ring_for_representation_category;
+
+      AddRandomMorphismWithFixedRange( category,
+        function( U, n )
+          local u, x, syz, m, M, e;
+          
+          if n = 0 then
```
sebastianpos: In this case, the special case is justified.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
```diff
@@ -2146,3 +2172,296 @@ end );
...
+
+    end );
+
+end );
+
+##
+InstallGlobalFunction( ADD_RANDOM_MORPHISM_WITH_FIXED_SOURCE_RIGHT,
+    function( category )
+
+      AddRandomMorphismWithFixedSource( category,
```
sebastianpos: I fear this method is invalid in the case where your ring is non-commutative, since then, transposing matrices does not behave well with taking products of matrices. 

You need to write this function in the same way as you treated the case for left modules.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
```diff
@@ -2146,3 +2172,296 @@ end );
...
+
+     end );
+
+end );
+
+##
+InstallGlobalFunction( ADD_RANDOM_MORPHISM_WITH_FIXED_RANGE_RIGHT,
+    function( category )
+
+      AddRandomMorphismWithFixedRange( category,
```
sebastianpos: Same problem here.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: I don't have a simple code that can search for a random non-zero morphism from presentation with relations to a free presentation. I am open to any suggestions or ideas.

sebastianpos: Your current code should already provide non-trivial examples such morphisms.

The general idea is: compose `SyzygiesOfColumns(m)` with some other random matrix.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: Involution have the following property Inv(AB)=Inv(B)Inv(A) and it is equal to transpose only in commutative rings. This property is exactly what needed to get a well-defined morphism: Any well-defined left presentations morphism gives us a commutative diagram between and applying the involution will give us a new commutative diagram from which we can derive a morphism of right presentations.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: Since the derivation of the random morphisms in the right presentations uses the random morphisms in the left presentations, I found it Ok to also derive random objects using the left presentations showing how these two categories relate to each other. Any way, I can change it as you propose. Milliseconds quicker is better than nothing.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: Yes, Thanks!. I think I excluded this situation without a careful thought. It can be handled without any problems using the general approach.

kamalsaleh: I've updated this block.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: These changes have been done and included in the PR.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
sebastianpos: Okay, so I correct my argument and state that not all rings have an involution.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
sebastianpos: Thanks a lot, this change increased the comprehensibility of the code.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
sebastianpos: Perfect!

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: One can ask a ring S if it has involution function by `IsBound( homalgTable(S)!.Involution );`. In this PR I am adding random methods only for homalg fields, ring of integers, free polynomial rings, exterior rings and their quotients. They all have Involution functions.

There is only one very unlikely to happen scenario:

If S is some other homalg ring without involution function and one adds by hand the components S!.random_element_func and S!.random_matrix_func then in the right presentations category he will get an error whenever he ask for random object or morphism.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: Yes, since n is the number of generators of the module.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
sebastianpos: Ok, it is very good that there is a way to test for the existence of an involution.

The least one should do in the code is to test whether an involution exists, and only in those cases install the random methods for the right module case. However, for me, that would be unsatisfying since that would mean treating the right module case more restrictive than the left module case.  I'll leave the choice to you.

In the graded case, you also depend on the existence of an involution for right modules, so again, either test for the existence of an involution (again with the downside of unequal behaviour between left and right modules) or don't use involutions.

kamalsaleh: Ok, then I will treat them equally and re-implement the two methods to the right presentations without the use if Involution.

kamalsaleh: In the current status of the PR the Involution is not required anymore.

CAP/gap/CategoryMorphisms.gd
```diff
@@ -65,6 +65,137 @@ DeclareGlobalFunction( "INSTALL_TODO_LIST_FOR_EQUAL_MORPHISMS" );
...
+###################################
+
+#! @Description
+#! The arguments are two objects $a$ and $b$ in a category $C$ and an integer $n$.
+#! The output is a random morphism $\alpha: a \rightarrow b$ in $C$.
+#! The integer $n$ indicates the degree of complexity of the created morphism
+#! $\alpha$ and its interpretation differs from one category to another.
+#! For example in the category of rational vector spaces, $n$ can indicate that 
+#! all entries of the matrix of $f$ are less or equal to $n$.
+#! @Returns a morphism in $\mathrm{Hom}(a,b)$
```
sebastianpos: This description needs to take into account that the output might be `fail`.
The same holds for the other descriptions.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
```diff
@@ -2146,3 +2172,336 @@ end );
...
+          U := AsRightPresentation( u );
+
+          return PresentationMorphism( U, y, M );
+     
+      end );
+
+end );
+
+##
+InstallGlobalFunction( TRY_TO_ENHANCE_HOMALG_RING_WITH_RANDOM_FUNCTIONS,
```
sebastianpos: @mohamed-barakat and @kamalsaleh :
it would be great if these random functions for HomalgRings could be transfered to `RingsForHomalg`. In that way, other packages, like `FreydCategoriesForCAP,` could also profit. 

@kamalsaleh:
Another example how other packages can profit: the method `RandomMorphismWithFixedSource` in `LinearAlgebraForCAP` could simply use underlying random functions of the given field. At the moment, it simply produces integers interpreted in the given field in a hard coded range between `n-10` and `n+10`.

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
```diff
@@ -2146,3 +2172,336 @@ end );
...
+            return fail;
+
+          fi;
+
+          m := UnderlyingMatrix( M );
+
+          y := homalg_ring!.random_matrix_func( NrRows( m ), n );
+
+          syz := SyzygiesOfColumns( UnionOfColumns( m, y ) );
+
```
sebastianpos: If you want to, you could improve the algorithm here by computing an arbitrary random matrix that you precompose with `SyzygiesOfColumns`, instead chopping of columns and multiplying them with `e`.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
```diff
@@ -1108,6 +1632,273 @@ InstallGlobalFunction( ADD_GRADED_INTERNAL_HOM_ON_MORPHISMS_RIGHT,
...
+        
+        if  ( Rank( DegreeGroup( S ) ) = 1 and
+                ( IsSubset( Set( [ 0, 1 ] ), Set( List( degrees_of_indeterminates, HomalgElementToInteger ) ) ) or
+                    IsSubset( Set( [ 0, -1 ] ), Set( List( degrees_of_indeterminates, HomalgElementToInteger ) ) ) ) ) then
+          
+          random_homogeneous_element_func :=
+            function( degree )
+              local mat, p, coeffs, r, l1, l2;
+              
+              if degree = Degree( Zero( S ) ) then
```
sebastianpos: This line unnecessarily prevents the generation of random elements that, simply by accident, have the same degree that was assigned to the zero element. See homalg-project/homalg_project#234

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: Thank you for the suggestion. It have done that in this commit:

https://github.com/homalg-project/CAP_project/pull/239/commits/dd5cbf7910255ca87872ebd457f8bbf0f14bb5ff

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
kamalsaleh: This has been added as a quick fix to some issues appeared in a previous implementation of the random morphisms. In this commit
https://github.com/homalg-project/CAP_project/pull/239/commits/b0a4508125e494db673d67afe760137fea9eb211
I have updated the corresponding methods and hence removed this quick fix.
Thanks for  catching this :) I am much more happy with the current code.

CAP/gap/CategoryMorphisms.gd
kamalsaleh: https://github.com/homalg-project/CAP_project/pull/239/commits/e08e47f6cbb702462cc95e6452f93c3db1283499

ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi
kamalsaleh: `GradedRingForHomalg` package is the natural place for these methods. I talked with @mohamed-barakat about adding them or similar methods to homalg and he told me we can work on that when he comes back.

kamalsaleh: The range in random methods of linear algebra package is `[-n-10, n+10]` for `n>0`.

kamalsaleh:
I think with this commit 
https://github.com/homalg-project/CAP_project/pull/239/commits/150f1707d0e49586a86766347e0e4e6fab078596
I have covered the most of what in my mind regarding this pull-request.
In the future we can also derive the random methods in other categories, for example generalized morphisms or Serre quotients, etc.

kamalsaleh:
Since this PR contains now 22 commits, some of them are revert of other commits or fixes or improvements of things already in PR, I decided to split it in to 4 PR's in each there is only 
one commit. This will make reviewing/editing them more easier and accessible.

The PR's are:

1- adding random methods with documentations to CAP,
2- adding random methods to the category of matrices of homalg field,
3- adding random methods to the category of left/right presentations,
4- adding random methods to the category of graded left/right presentations.

This PR will consequently be closed, but still be referenced in all of the the future 4 PR's.

PR #243, opened 2019-01-21
Outsourced package MonoidalCategories

mohamed-barakat:


CAP/PackageInfo.g
```diff
@@ -102,6 +102,7 @@ Dependencies := rec(
   NeededOtherPackages := [ [ "AutoDoc", ">= 2016.02.16" ],
                            [ "ToolsForHomalg", ">= 2018.05.22" ],
                            [ "io", ">=0" ],
+                           [ "CAP", ">= 2019.01.01" ],
```
zickgraf:
```suggestion
                           [ "MonoidalCategories", ">= 2019.01.16" ],
```

This might fix the CI fail.

sebasguts:
This PR moves functionality from the deposited package `CAP` to the non-deposited package `MonoidalCategories`, so merging this would actually reduce functionality in the released version. Should we nevertheless merge it?

CAP/PackageInfo.g
zickgraf: CAP should not depend on MonoidalCategories anyway.

mohamed-barakat: In the recent commit it does not

MonoidalCategories/PackageInfo.g
```diff
@@ -0,0 +1,130 @@
...
+  BookName  := "MonoidalCategories",
+  ArchiveURLSubset := ["doc"],
+  HTMLStart := "doc/chap0.html",
+  PDFFile   := "doc/manual.pdf",
+  SixFile   := "doc/manual.six",
+  LongTitle := "Monoidal and monoidal (co)closed categories",
+),
+
+Dependencies := rec(
+  GAP := ">= 4.9.2",
```
sebastianpos: Is there a reason for depending on GAP 4.9.2 rather than GAP 4.9.1?

I would like to have the same GAP dependency in `MonoidalCategories` as in `CAP` (which, at the moment, is 4.9.1).

mohamed-barakat: 4.9.2 fixed some issues with the history and with break loop, which are irrelevant here. Changed to 4.9.1.

MonoidalCategories/.gitignore
```diff
@@ -0,0 +1,52 @@
+/maketest.g
```
sebasguts: This file is not necessary

MonoidalCategories/doc/Doc.autodoc
```diff
@@ -0,0 +1,2 @@
+@Chapter Tensor Product and Internal Hom
```
sebasguts: This file is not necessary

sebastianpos: I would suggest to leave this file, as it will happen that there appear more chapters in the future.

MonoidalCategories/.gitignore
mohamed-barakat: I agree.

MonoidalCategories/doc/Doc.autodoc
mohamed-barakat: I agree.

mohamed-barakat:
Thanks :)

PR #245, opened 2019-01-28
Fixed bug in symbol recognition

sebasguts:
Fixes @mohamed-barakat s comment in #238


sebasguts:
@sebastianpos Would be nice if we could merge this quickly, as it is a serious bug

PR #246, opened 2019-01-29
Make Symbol checker return strings instead of objects again

sebasguts:
@mohamed-barakat Please note that #243 needs to be adjusted to this PR.
@zickgraf This should fix the bug in the pre function business.


zickgraf:
As confirmation: yes, this fixes the prepare functions.

PR #250, opened 2019-02-26
Allow pre- and postprocessing in opposite category

sebastianpos:
controled by new entries in the Method Record.
This fixes an issue pointed out in #239


CAP/gap/MethodRecord.gi
```diff
@@ -2071,7 +2075,9 @@ UniversalMorphismIntoCoimage := rec(
   cache_name := "UniversalMorphismIntoCoimage",
   universal_object_position := "Range",
   universal_type := "Colimit",
-  
+  dual_preprocessor_func := function( arg )
+      return CAP_INTERNAL_REVERSE_LISTS_IN_ARGUMENTS( CAP_INTERNAL_OPPOSITE_RECURSIVE( arg ) );
```
sebasguts: The function `REVERSE_LISTS_IN_ARGUMENTS` is never used without the `OPPOSITE_RECURSIVE` function. Maybe we could just fuse the two, or create a third that does fuse the two, and add that function directly without creating a new function for each occurance of the pair?

sebastianpos: I created a fused version.

CAP/gap/MethodRecord.gi
```diff
@@ -3104,3 +3096,20 @@ InstallGlobalFunction( CAP_INTERNAL_ENHANCE_NAME_RECORD,
 end );
 
 CAP_INTERNAL_ENHANCE_NAME_RECORD( CAP_INTERNAL_METHOD_NAME_RECORD );
+
+##
+InstallGlobalFunction( CAP_INTERNAL_REVERSE_LISTS_IN_ARGUMENTS_FOR_OPPOSITE,
```
sebasguts: Just a little nitpick: Should this function be in the `MethodRecord.gi` or maybe even in `ToolsForCategories.gi` or so? It does not really matter though, we can always move it.

sebastianpos: Unfortunately, the function cannot be in `ToolsForCategories.gi` since this file is already loaded in the `init.g` at some very specific place.

sebasguts: I see, we also discussed that. Maybe we should have two `tools` files, one that needs to be loaded very early, and one with the rest that can be loaded using the normal scheme

sebastianpos: Ok, I suggest I merge this PR as it is now, and we resolve the Tools issue in a seperate thread.

sebasguts: Yep.

PR #254, opened 2019-02-27
Adding LiftAlongMonomorphism & ColiftAlongEpimorphism to

kamalsaleh:
the category of graded presentations.

The outputs of Lift, LiftAlongMonomorphism, Colift and
ColiftAlongEpimorphism are checked whether they are well-defined or not
and an error will be raised whenever it is not well-defined.

This checking is supposed to be temporary until the pull-request
    https://github.com/homalg-project/homalg_project/pull/222
is merged to homalg_project.

### DIFF
--- a/GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
+++ b/GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
@@ -198,6 +198,10 @@ InstallGlobalFunction( ADD_GRADED_FUNCTIONS_FOR_LEFT_PRESENTATION,
     ADD_GRADED_IS_IDENTICAL_FOR_MORPHISMS( category );
     
     ADD_GRADED_EPIMORPHISM_FROM_SOME_PROJECTIVE_OBJECT( category );
+
+    ADD_GRADED_LIFT_ALONG_MONOMORPHISM( category );
+
+    ADD_GRADED_COLIFT_ALONG_EPIMORPHISM( category );
     
     if HasIsCommutative( category!.ring_for_representation_category ) and IsCommutative( category!.ring_for_representation_category ) then
       
@@ -271,6 +275,10 @@ InstallGlobalFunction( ADD_GRADED_FUNCTIONS_FOR_RIGHT_PRESENTATION,
     ADD_GRADED_IS_IDENTICAL_FOR_MORPHISMS( category );
     
     ADD_GRADED_EPIMORPHISM_FROM_SOME_PROJECTIVE_OBJECT( category );
+
+    ADD_GRADED_LIFT_ALONG_MONOMORPHISM( category );
+
+    ADD_GRADED_COLIFT_ALONG_EPIMORPHISM( category );
     
     if HasIsCommutative( category!.ring_for_representation_category ) and IsCommutative( category!.ring_for_representation_category ) then
       
@@ -574,10 +582,20 @@ InstallGlobalFunction( ADD_GRADED_LIFT,
         lift := Lift( UnderlyingPresentationMorphism( alpha ), UnderlyingPresentationMorphism( beta ) );
         
         if lift = fail then
+        
             return fail;
+        
         fi;
         
-        return GradedPresentationMorphism( Source( alpha ), lift, Source( beta ) );
+        lift := GradedPresentationMorphism( Source( alpha ), lift, Source( beta ) );
+        
+        if not IsWellDefined( lift ) then
+          
+          Error( "An output of the Lift method is not well-defined!" );
+          
+        fi;
+        
+        return lift;
         
     end );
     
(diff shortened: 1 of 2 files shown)


zickgraf:
Just a note: once #188 is merged you could make it that the checks are only executed if the value of the `output_sanity_check_level` of the category is at least `1`.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
```diff
@@ -926,6 +954,70 @@ InstallGlobalFunction( ADD_GRADED_IDENTITY,
...
+        
+        if lift = fail then
+
+            return fail;
+
+        fi;
+
+        lift := GradedPresentationMorphism( Source( tau ), lift, Source( iota ) );
+
+        if not IsWellDefined( lift ) then
```
sebastianpos: In the case of lifts along monomorphisms, a computed non-graded lift has automatically to be graded. The same holds for colifts along epimorphisms.

kamalsaleh: You are right, since the kernel iota is zero. This checks will be removed anyway when the PR 
https://github.com/homalg-project/homalg_project/pull/222
is merged to homalg.

kamalsaleh: You are right, I've updated the PR.

mohamed-barakat:
I just merge homalg-project/homalg_project#222 into `homalg_project`.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
```diff
@@ -926,6 +954,54 @@ InstallGlobalFunction( ADD_GRADED_IDENTITY,
...
+  function( category )
+    
+    AddLiftAlongMonomorphism( category,
+
+      function( iota, tau )
+        local lift;
+
+        lift := LiftAlongMonomorphism( UnderlyingPresentationMorphism( iota ), UnderlyingPresentationMorphism( tau ) );
+        
+        if lift = fail then
```
sebasguts: `LiftAlongMonomorphism` and `ColiftAlongEpimorphism` should not return `fail`

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
```diff
@@ -574,10 +582,20 @@ InstallGlobalFunction( ADD_GRADED_LIFT,
...
             return fail;
+        
         fi;
         
-        return GradedPresentationMorphism( Source( alpha ), lift, Source( beta ) );
+        lift := GradedPresentationMorphism( Source( alpha ), lift, Source( beta ) );
+        
+        if not IsWellDefined( lift ) then
+          
+          Error( "An output of the Lift method is not well-defined!" );
```
sebasguts: I think this is not the right way to do this, for two reasons:

1. Primitive implementations should not secretly depend on other primitives, like `IsWellDefined` here. But this is not too bad.

2. `Lift` can return `fail`, and I would think it is appropriate to return `fail` here.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
sebasguts: `LiftAlongMonomorphism` and `ColiftAlongEpimorphism` should not return `fail`. They can throw an error, but maybe the garbage-in-garbage-out principle applies here, and we should assume that if they are called, the lift exists.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
sebastianpos: This change in `Lift` and `Colift` was supposed to be temporary and is outdated by #262.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
sebastianpos: Yes, you are right. The `fail` case can simply be deleted.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
sebasguts: I just approved #262. If you are fine with it, we can merge it and then update this PR and merge it, too?

sebastianpos: Yes.

GradedModulePresentationsForCAP/gap/GradedModulePresentations.gi
sebasguts: Shouldn't the fail case throw a meaningful error? Or do we just rely on CAP giving some sensible error message? I would actually prefer the latter.

sebastianpos: If there is no graded lift along monomorphism then there also will be no non-graded lift along monomorphism (provided the input consists of graded morphisms).

Thus, an error would be thrown by the underlying non-graded case.

sebasguts: Perfect, Then I will provide a patch for this PR.

sebasguts:
Since I cannot push to @kamalsaleh s PR branch, I have opened a new PR #279 containing a rebased version of this PR, and the fixes we discussed.

PR #256, opened 2019-03-03
DO NOT MERGE: avoid using InstallMethodWithCache if cache is deactivated

mohamed-barakat:
This is NOT a PR but rather a question.

@markuslh and I noticed, that if we `DeactivateCachingOfCategory`, even immediately after `CreateCapCategory`, then some methods are still installed using `InstallMethodWithCache`. We now believe that this is necessary to allow caching to be reactivated a posteriori using `SetCaching`.

Now our questions:
* Is our current understanding correct?
* Is there no significant overhead if those methods (e.g. `DirectProductOp`) are still installed using `InstallMethodWithCache`?

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2019.01.29", ## Mohamed's version
+  "2019.02.27", ## Mohamed's version
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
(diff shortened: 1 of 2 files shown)


sebastianpos:
Deactivating caches in a category does not affect the call of `InstallMethodWithCache` when installing a basic operation for a category. Yes, the intention was to make it possible to reactivate the caches again within a session. The overhead is at least 1 indirection.

I am curious. The code that you provide should avoid the overhead if you deactivate the caching of the category before adding any basic operations to it. So, do you profit from this change? If yes, then we can think about it more closely.

sebasguts:
@mohamed-barakat As you pointed out, this change will make it impossible for the category to reactivate the cache, which is currently not the intended behaviour. If you would want that behaviour, it should be documented.

Furthermore, your patch will not allow to specify any separate caching values for any function if the default is none. At least that should be possible, so we would have to do it differently.

sebastianpos:
@mohamed-barakat 
Are you satisfied with the answers and is it okay for you if I close this PR/question?

mohamed-barakat:
Yes, I am. I will close the PR :)

Issue #184, opened 2018-08-31
IsWellDefined should not have a cache


sebastianpos:
Since both the functions

`IsWellDefinedForObjects`

and 

`IsWellDefinedForMorphisms`

are intended as debugging tools, a caching mechanism seems inappropriate
for this purpose, since it is unclear how

`IsEqualForCacheForObjects`

and

`IsEqualForCacheForMorphisms`

behave on non-well-defined objects/morphisms.

PR #257, opened 2019-03-04
Merge cap category of projective graded modules

HereAround:
Package CAPCategoryOfProjectiveGradedModules is to be integrated into the package FreydCategories. As a starting point for this integration, this commit places all defining files of the package CAPCategoryOfProjectiveGradedModules into a subfolder of FreydCategories. In particular, the history of the package CAPCategoryOfProjectiveGradedModules is conserved by this operation.


sebasguts:
@HereAround Shouldn't those files be rather a new package than a subpackage? It seems strange to me to collect those in a subfolder of FreydCategories. @sebastianpos thoughts?

sebastianpos:
The files provided by @HereAround are not meant to be a "subpackage", but simply the implementation of another category for the package `FreydCategoriesForCAP.` The data are collected in a subfolder not for any technical reason, but simply for keeping the overview.

The idea for the future of `FreydCategoriesForCAP` is to collect constructors for additive categories in a single package.

sebastianpos:
The first thing we need to discuss is the notion of equality in the category of rows.
The current implementation leads to inconsistencies:

```
gap> Q := HomalgFieldOfRationalsInSingular();;
gap> S := GradedRing( Q * "x" );;
gap> SetWeightsOfIndeterminates( S, [1] );;
gap> A := GradedRow( [ [[1],1], [[2],1] ], S );;
gap> B := GradedRow( [ [[2],1], [[1],1] ], S );;
gap> mat := HomalgIdentityMatrix( 2, 2, S );;
gap> alpha := GradedRowOrColumnMorphism( A, mat, A );;
gap> IsWellDefined( alpha );
true
```
Note that with the current notion of equality of objects, we have `A = B`, which for CAP means the two objects are the same on a technical level. In particular, equal inputs yield equal output. However, if I restart the session and alter the definition of `alpha` by exchanging its range with `B`, I get
```
gap> Q := HomalgFieldOfRationalsInSingular();;
gap> S := GradedRing( Q * "x" );;
gap> SetWeightsOfIndeterminates( S, [1] );;
gap> A := GradedRow( [ [[1],1], [[2],1] ], S );;
gap> B := GradedRow( [ [[2],1], [[1],1] ], S );;
gap> mat := HomalgIdentityMatrix( 2, 2, S );;
gap> alpha := GradedRowOrColumnMorphism( A, mat, B );;
gap> IsWellDefined( alpha );
false
```
Note that I had to create a new session due to issue #184.

I think the solution would be to simply consider to objects to be equal if their degree lists are equal, and not if they are permutations of each other. Of course, one has to check that with this definition, all implemented functions satisfy CAP's main specification: equal input yield equal output.

HereAround:
I indeed followed your suggestion. The IsEqualForObjects method for graded rows and graded columns now simply compares the unzipped degree lists. Two graded rows/columns are considered equal whenever these lists are identical.

Let me elaborate on the meaning of "unzipped" a bit further. To this end, first consider the following example:

gap> Q := HomalgFieldOfRationalsInSingular();;
gap> S := GradedRing( Q * "x" );;
gap> SetWeightsOfIndeterminates( S, [1] );;
gap> A := GradedRow( [ [[1],1], [[2],100], [[1],1] ], S );;

It is very convenient to enter that the degree 2 appears 100 times, rather than entering 100 times the number 2 in the command line. This format one could call the "compact degree list". The unzipped one however, will replace [ [2], 100 ] with a list of length 200 in which each entry is a 2. By repeating this step for every entry of the "compact degree list", we obtain the "unzipped degree list". In the case at hand, this is thus

[ 1, 2, 2, 2, ..., 2, 2, 1 ]

with 100 x 2 sandwiched between the 1 at the beginning and the 1 at the end. 

Note that the 1 at the beginning and the 1 at the end are not collected together. In particular, the following code produces two graded rows, which are not equal:

gap> Q := HomalgFieldOfRationalsInSingular();;
gap> S := GradedRing( Q * "x" );;
gap> SetWeightsOfIndeterminates( S, [1] );;
gap> A := GradedRow( [ [[1],1], [[2],100], [[1],1] ], S );;
gap> B := GradedRow( [ [[1],2], [[2],100] ], S );;

Last but not least, the following two modules are considered equal, although they are defined by different inputs:

gap> Q := HomalgFieldOfRationalsInSingular();;
gap> S := GradedRing( Q * "x" );;
gap> SetWeightsOfIndeterminates( S, [1] );;
gap> A := GradedRow( [ [[1],1], [[1],1] ], S );;
gap> B := GradedRow( [ [[1],2] ], S );;

I hope you are satisfied with this? I also adjusted the corresponding tests, which previously returned true and are now supposed to yield false.

FreydCategoriesForCAP/gap/CategoryOfGradedRowsAndColumns/CategoryOfGradedColumns.gi
```diff
@@ -30,8 +30,10 @@ InstallMethod( CAPCategoryOfGradedColumns,
 
       # set its properties
       SetIsAdditiveCategory( category, true );
-      SetIsStrictMonoidalCategory( category, true );
-      SetIsRigidSymmetricClosedMonoidalCategory( category, true );
+      if IsCommutative( homalg_graded_ring ) then
```
kamalsaleh: Here I would draw attention to this issue
https://github.com/homalg-project/homalg_project/issues/191

HereAround: Thanks for this pointer. I have changed the code such that it test if the underlying non-graded ring is commutative.

sebastianpos:
All in all, the implementation looks good :-)

I pushed my changes to the branch of this PR. 
@HereAround: Please take a look at the changes. If you are satisfied with them, I will merge this PR.

HereAround:
@sebastianpos Your changes look good! I agree.

Here are a few more minor changes that I have made:
- the tests now apply FreydCategory to graded rows
- IsWellDefinedForMorphism now uses homalg-project/homalg_project#240
- ToDo-file has been updated

Please have a look at these.

PR #258, opened 2019-03-04
Package CAPCategoriesOfGradedModules fully merged into FreydCategories

HereAround:


sebastianpos:
It seems to be the cases that this PR does exactly the same as #257.

HereAround:
Unless I made a mistake, this is not the case. #257 is the status where the original package CAPCategoryOfProjectiveGradedModules is integrated as a sub folder into FreydCategoriesForCAP. This PR is beyond that, in that now the full functionality of graded rows and columns is available after loading Freud Categories, the test of graded rows and columns are part of the tests in FreydCategories, the docu of FreydCategories has been extended by graded rows and columns and so on...

sebastianpos:
Then, I fear, you made a mistake. For example, both PRs provide the same number of commits and end with cc5db0536810088d, a commit that already does changes of the type you described above.

Be that as it may, I don't think that it is necessary to have a PR-split in "copying the original package" and "adjusting the package to comply with `FreydCategoriesForCAP`". Simply close this PR, and I'll review the end result provided in #257.

HereAround:
You're right, my bad. Anyways, I follow your suggestion and close this PR.

PR #263, opened 2019-03-05
Add sanity checks for the results of primitive functions

zickgraf:
similar to the checks added in https://github.com/homalg-project/CAP_project/pull/188

### DIFF
--- a/CAP/gap/InstallAdds.gi
+++ b/CAP/gap/InstallAdds.gi
@@ -43,6 +43,16 @@ BindGlobal( "CAP_INTERNAL_ADD_MORPHISM_OR_FAIL",
     
 end );
 
+BindGlobal( "CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY",
+  
+  function( function_name, category, message )
+    
+    Error( Concatenation( "in function \033[1m", function_name,
+        "\033[0m\n       of category \033[1m",
+        Name( category ), ":\033[0m\n\033[1m       ", message, "\033[0m\n" ) );
+    
+end );
+
 InstallGlobalFunction( CapInternalInstallAdd,
   
   function( record )
@@ -259,17 +269,13 @@ InstallGlobalFunction( CapInternalInstallAdd,
                     
                     pre_func_return := CallFuncList( pre_function, arg );
                     if pre_func_return[ 1 ] = false then
-                        Error( Concatenation( "in function \033[1m", record.function_name, 
-                            "\033[0m\n       of category \033[1m",
-                            Name( category ), ":\033[0m\n\033[1m       ", pre_func_return[ 2 ], "\033[0m\n" ) );
+                        CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, pre_func_return[ 2 ] );
                     fi;
                     
                     if category!.input_sanity_check_level > 1 then
                         pre_func_return := CallFuncList( pre_function_full, arg );
                         if pre_func_return[ 1 ] = false then
-                            Error( Concatenation( "in function \033[1m", record.function_name, 
-                                "\033[0m\n       of category \033[1m",
-                                Name( category ), ":\033[0m\n\033[1m       ", pre_func_return[ 2 ], "\033[0m\n" ) );
+                            CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, pre_func_return[ 2 ] );
                         fi;
                     fi;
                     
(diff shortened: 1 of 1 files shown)


CAP/gap/InstallAdds.gi
```diff
@@ -281,8 +271,74 @@ InstallGlobalFunction( CapInternalInstallAdd,
...
+                            AddTwoCell( category, result );
+                        elif record.return_type = "morphism_or_fail" then
+                            CAP_INTERNAL_ADD_MORPHISM_OR_FAIL( category, result );
+                        fi;
+                    elif category!.output_sanity_check_level > 0 then
+                        if record.return_type = "object" then
+                            if not IsCapCategoryObject( result ) then
+                                display_error_for_current_function( "the result does not lie in the filter IsCapCategoryObject." );
+                            fi;
+                            if not HasCapCategory( result ) then
```
sebasguts: You could move out the `HasCapCategory` and `IsIdenticalObj(CapCategory...` `if`s out of the inner `if`s, if you do not want the stuff to be checked in a particular order. `HasCapCategory` can be applied to any object, since it is a filter.

CAP/gap/InstallAdds.gi
```diff
@@ -233,6 +221,12 @@ InstallGlobalFunction( CapInternalInstallAdd,
             fi;
             
         od;
+
+        display_error_for_current_function := function( message )
```
sebasguts: You could also move this function outside of `CapInternalInstallAdd`, and give the other parameters as arguments. This way you could prevent it from being created a lot of times, but only once.

CAP/gap/InstallAdds.gi
```diff
@@ -104,18 +104,6 @@ InstallGlobalFunction( CapInternalInstallAdd,
         argument_list := [ 1 .. Length( filter_list ) ];
     fi;
     
-    if record.return_type = "object" then
```
sebasguts: The position of this `add_function` variable was on purpose, actually, as then the if is only carried out at install time, not at every call of the categorical operation.

sebasguts:
Now thinking about it, I would actually like to see my last comment adressed/discussed first.

CAP/gap/InstallAdds.gi
zickgraf: I think `result` could also be a boolean here, so not having a `CapCategory` would in general not be an error. I could move `IsIdenticalObj(CapCategory...` outside but then I would first have to check for `HasCapCategory` again, so I just kept this inside.

CAP/gap/InstallAdds.gi
zickgraf: Okay, should be fixed

CAP/gap/InstallAdds.gi
zickgraf: Ah, I see, should be fixed.

CAP/gap/InstallAdds.gi
sebasguts: Not really, because you already checked `HasCapCategory`, and if you did not have one, you got an error, so you would never get to the inside.

CAP/gap/InstallAdds.gi
sebasguts: Thanks :)

CAP/gap/InstallAdds.gi
sebasguts: Thanks :)

CAP/gap/InstallAdds.gi
zickgraf: As I understand it, you suggest the following code:
```
...
elif category!.output_sanity_check_level > 0 then
    if not HasCapCategory( result ) then
        CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result has no CAP category." );
    fi;
    if not IsIdenticalObj( CapCategory( result ), category ) then
        CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result does not lie in this category." );
    fi;
    if record.return_type = "object" then
        if not IsCapCategoryObject( result ) then
            CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result does not lie in the filter IsCapCategoryObject." );
        fi;
        if not ObjectFilter( category )( result ) then
            CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result does not lie in the object filter of this category." );
        fi;
elif record.return_type = "morphism" then
...
```
This does not work, since `result` might be a boolean, for example for `IsWellDefined`. So calling `IsWellDefined` would always lead to the error `the result has no CAP category.`

If you suggest something else, I don't understand what exactly.

sebasguts: I see the problem. You could tackle that problem by asking `IsAttributeStoringRep` at the beginning, at it is the minimal condition for all those tests anyway.

zickgraf: If you suggest the following:
```
...
elif category!.output_sanity_check_level > 0 then
    if IsAttributeStoringRep( result ) then
        if not HasCapCategory( result ) then
            CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result has no CAP category." );
        fi;
        if not IsIdenticalObj( CapCategory( result ), category ) then
            CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result does not lie in this category." );
        fi;
    fi;
    if record.return_type = "object" then
        if not IsCapCategoryObject( result ) then
            CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result does not lie in the filter IsCapCategoryObject." );
        fi;
        if not ObjectFilter( category )( result ) then
            CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category, "the result does not lie in the object filter of this category." );
        fi;
elif record.return_type = "morphism" then
...
```
This does not work if the `return_type` is `other_object`, because the other object does not have to fulfill `IsIdenticalObj( CapCategory( result ), category )`.

sebasguts: You are right. Lets leave it how it is then.

PR #269, opened 2019-03-07
Disable adds by default

zickgraf:
~Note: This depends on PR #265. I will open this PR for review once #265 is merged.~

**Note 2**: if this PR is merged, the CAP Manual has to be updated accordingly.

This PR sets the default value for `add_primitive_output` to `false`.
Rationale for this change: Most users of CAP do not rely on the automatic calls of `Add` but also do not know about `DisableAddForCategoricalOperations` (of course, I do not have numbers to support this statement, but I think it is a reasonable guess). On the other hand, users relying on the automatic calls of `Add` are probably advanced users who either know about `EnableAddForCategoricalOperations` or know where to look if errors occur if `add_primitive_output` is set to `false`.

Thus, most users benefit from this change in two ways:
* faster code
* stricter output sanity checks (see #188 and #263)

In particular, stricter output sanity checks are very useful for users now to CAP and GAP.

### DIFF
--- a/CAP/gap/CAP.gd
+++ b/CAP/gap/CAP.gd
@@ -378,19 +378,23 @@ DeclareGlobalFunction( "EnableBasicOperationTypeCheck" );
 
 #############################################
 ##
-#! @Section Disable add in functions
+#! @Section Enable automatic calls of <C>Add</C>
 ##
 #############################################
 
 #! @BeginGroup
 #! @Description
-#!  Enables/disables the call of add for the output
+#!  Enables/disables the automatic call of <C>Add</C> for the output
 #!  of primitively added functions for the category <A>C</A>.
-#!  This can be savely done if the all objects and morphisms
-#!  are added to the category by their constructors.
-#! @Arguments C
-DeclareGlobalFunction( "DisableAddForCategoricalOperations" );
+#!  If the automatic call of <C>Add</C> is disabled (default),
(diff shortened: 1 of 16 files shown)


zickgraf:
Now that #265 is merged, this PR is ready for review.

CAP/gap/CAP.gd
```diff
@@ -378,19 +378,22 @@ DeclareGlobalFunction( "EnableBasicOperationTypeCheck" );
...
 #! @BeginGroup
 #! @Description
-#!  Enables/disables the call of add for the output
+#!  Enables/disables the call of <C>Add</C> for the output
 #!  of primitively added functions for the category <A>C</A>.
-#!  This can be savely done if the all objects and morphisms
-#!  are added to the category by their constructors.
-#! @Arguments C
-DeclareGlobalFunction( "DisableAddForCategoricalOperations" );
+#!  If this is disabled (default), the output of primitively added functions must
```
sebasguts: Can you not use `If this is enabled`, but something like `Enables ...`/`Disables ...`?

It just seems a bit weird to read it that way, as the docstring should document what the function does, and not what the effects are. It is not completely unclear here, but nevertheless.

sebasguts: I didn't see the first line, sorry.

zickgraf:
I have polished the documentation a bit.

PR #272, opened 2019-03-12
Fix bug in LinearAlgebraForCAP

sebastianpos:


sebastianpos:
Basically, the hard coded 0 stands for the fact that the homomorphism space between two vector spaces V,W is itself 0-dimensional if V is 0-dimensional (also if W is 0-dim., but that is irrelevant for this PR).

PR #276, opened 2019-03-14
Improve documentation

zickgraf:
~Note: this PR depends on #269. I will open it for review once #269 is merged.~

To make the technical requirements more precise, I have changed  `IsCapCategoryCell` to imply `IsAttributeStoringRep`. If a GAP object does not lie in `IsAttributeStoringRep`, it cannot be added
to a category anyway.

### DIFF
--- a/CAP/doc/AddFunctions.autodoc
+++ b/CAP/doc/AddFunctions.autodoc
@@ -138,8 +138,16 @@ The record can have the following components, used as described:
                             Please note that if you have a method selection argument for your function, you need to give the argument_list
                             to explicitly state which argument is the method selection argument.
 
-* return_type (optional): The return type can be object, morphism, or twocell. If it is one of those, the correct add function (see above) is
-                          used for the result of the computation. Otherwise, no add function is used after all.
+* return_type (optional): The return type can be one of the following:
+                          * <C>object</C>,
+                          * <C>morphism</C> or <C>morphism_or_fail</C>,
+                          * <C>twocell</C>,
+                          * <C>bool</C>,
+                          * <C>other_object</C>,
+                          * <C>other_morphism</C>.
+                          <P/>
+                          If it is one of the first three options, the correct <C>Add</C> function (see above) is
+                          used for the result of the computation. Otherwise, no <C>Add</C> function is used after all.
 
 * is_with_given: Boolean, marks wether the function which is to be installed a with given function or not.
 
--- a/CAP/gap/CAP.gd
+++ b/CAP/gap/CAP.gd
@@ -258,6 +258,7 @@ DeclareAttribute( "RangeCategoryOfHomomorphismStructure",
 #############################################
 ##
 #! @Section Logic switcher
+#! @SectionLabel Logic_switcher
 ##
 #############################################
 
@@ -318,6 +319,7 @@ DeclareProperty( "IsWellDefined",
 ####################################
 ##
 #! @Section Caching
+#! @SectionLabel Caching
 ##
 ####################################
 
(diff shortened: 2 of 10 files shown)


zickgraf:
This PR is now ready for review.

CAP/doc/AddFunctions.autodoc
```diff
@@ -138,8 +138,16 @@ The record can have the following components, used as described:
                             Please note that if you have a method selection argument for your function, you need to give the argument_list
                             to explicitly state which argument is the method selection argument.
 
-* return_type (optional): The return type can be object, morphism, or twocell. If it is one of those, the correct add function (see above) is
-                          used for the result of the computation. Otherwise, no add function is used after all.
+* return_type (optional): The return type can be one of the following:
+                          * <C>object</C>,
+                          * <C>morphism</C> or <C>morphism_or_fail</C>,
```
sebasguts: It looks a bit weird to bundle this tbh . . .

CAP/gap/CategoryMorphisms.gd
```diff
@@ -413,14 +413,21 @@ DeclareOperation( "AddIsAutomorphism",
...
 
+#! @Description
+#!  Adds <A>morphism</A> as a morphism to <A>category</A>.
+#! @Arguments category, morphism
 DeclareOperation( "Add",
                   [ IsCapCategory, IsCapCategoryMorphism ] );
 
+#! @Description
+#!  Adds <A>morphism</A> as a morphism to <A>category</A>.
+#! @Arguments category, morphism
```
sebasguts: Should those two docstrings may be grouped, and then describe the difference explicitly, i.e., in one case, the morphism needs to be a `CapCategoryMorphism`?

CAP/doc/AddFunctions.autodoc
zickgraf: I bundled this because I think about this as "a function can return an object, a morphism (with the special case `fail` in some cases), a two-cell, ...". But I also don't mind splitting it up. I think we should choose the version @sebastianpos likes better, than we have a clear vote :D

CAP/gap/CategoryMorphisms.gd
zickgraf: If we group the docstrings, only the filters of the first operation are visible in the manual, right? I don't think, this makes things clearer for the user. But I agree that we should make the relation between the two more clear. Thus, I propose the following:
```suggestion
#! @Description
#!  Adds <A>morphism</A> as a morphism to <A>category</A>. If <A>morphism</A> already lies in the filter <C>IsCapCategoryMorphism<C>, the operation <C>Add</C> can be used instead.
#! @Arguments category, morphism
```

sebasguts: Very good!

CAP/doc/AddFunctions.autodoc
sebasguts: Okay, fair enough :).

sebastianpos: I flipped a coin. Fabian's version won :-)

CAP/gap/CAP.gd
```diff
@@ -333,14 +335,26 @@ DeclareOperation( "SetCachingToCrisp",
...
                   [ IsCapCategory, IsString ] );
 
+#! @Description
+#!  Sets the caching of <A>category</A> to <A>type</A>.
+#! @Arguments category, type
 DeclareGlobalFunction( "SetCachingOfCategory" );
+
+#! @BeginGroup
+#! @Description
+#!  Sets the caching of <A>category</A> to <C>weak</C>, <C>crips</C> or <C>none</C>, respectively.
```
sebastianpos:
```suggestion
#!  Sets the caching of <A>category</A> to <C>weak</C>, <C>crisp</C> or <C>none</C>, respectively.
```

CAP/doc/AddFunctions.autodoc
sebasguts: Fair :)

sebasguts:
@zickgraf Can you apply @sebastianpos suggestion? I will merge afterwards

PR #281, opened 2019-03-22
Caching in MatrixCategory

sebastianpos:
I would like to discuss the caching in `MatrixCategory`.

Since this category is supposed to be a "work horse", it think caching (as provided by `ToolsForHomalg)` should be kept to a minimum. This is why I suggest to delete it from the current object and morphism constructors.

However, there is a cheap way to cache objects via `KeyDependentOperation`, which I suggest to incorporate into the current implementation.

### DIFF
--- a/LinearAlgebraForCAP/gap/MatrixCategoryMorphism.gd
+++ b/LinearAlgebraForCAP/gap/MatrixCategoryMorphism.gd
@@ -39,8 +39,8 @@ DeclareCategory( "IsVectorSpaceMorphism",
 DeclareOperation( "VectorSpaceMorphism",
                   [ IsVectorSpaceObject, IsHomalgMatrix, IsVectorSpaceObject ] );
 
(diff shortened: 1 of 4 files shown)


sebastianpos:
If there are no concerns then I will merge this PR in one week.

mohamed-barakat:
I think it is a good idea, in principle. I just would've wished we have an extensive functioning CI so that we are able to test possible regressions. This is how Markus and I were able to handle regressions after such subtle changes in homalg.

@zickgraf Is travis/circle-ci able to log timings/report regressions?

mohamed-barakat:
I now think the use of `KeyDependentOperation` should even be faster than caching.

zickgraf:
> Is travis/circle-ci able to log timings/report regressions?

I'm not aware of such a feature. CircleCI  logs build times but they vary quite a bit: https://circleci.com/build-insights/gh/homalg-project

Issue #57, opened 2015-09-22
Draw Diagrams in Manual


sebastianpos:
It would be helpful to have diagrams for all universal constructions
in the manual.

PR #284, opened 2019-03-26
Added Tikz-Pictures for factor objects and subobjects

HereAround:
To close https://github.com/homalg-project/CAP_project/issues/57, I have added pictures for subobjects and factor objects. They can be used to discuss the desired styling.

Once the style has been agreed on, I can provide more images.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gd
+++ b/CAP/gap/CategoryMorphisms.gd
@@ -814,8 +814,21 @@ DeclareOperation( "AddIsEqualAsFactorobjects",
 
 
 #! @Description
+#! @BeginLatexOnly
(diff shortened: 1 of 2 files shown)


sebastianpos:
Thanks a lot for the pictures :-)

We agreed on the forced-pushed format. Please note that:
- there is no need for a new folder `Pictures`
- when including the pictures into the manual, we use `BeginLatexOnly` and `EndLatexOnly`.
@HereAround:
If you're okay with these changes, then I'll merge this PR.

HereAround:
Thanks for the update!

The changes look good. I fully agree. Please merge.

PR #286, opened 2019-03-26
Adding The random methods declations to CAP-method record

kamalsaleh:
The random methods are:

* RandomObject( category, n ), 
* RandomMorphism( category, n ), 
* RandomMorphismWithFixedSource( object, n ), 
* RandomMorphismWithFixedRange( object, n )
* RandomMorphismWithFixedSourceAndRange( object_1, object_2 ) 

to the method record with the documentation and two derivation methods for
RandomMorphism.

The complexity integer n should some-how encode how complicated the created random object or morphism is, and hence may have different interpretations in different categories.


CAP/gap/MethodRecord.gi
```diff
@@ -3131,7 +3131,67 @@ SolveLinearSystemInAbCategory := rec(
   return_type := "morphism_or_fail"
 ),
 
-  ) );
+RandomObjectByInteger := rec(
```
sebasguts: I am not sure if this is worth it, but do you want to specify `io_types` for those entries?

PR #288, opened 2019-03-26
Adding random methods to the category of left/right presentations over homalg ring.

kamalsaleh:
`RandomObject( cat, n )` should return an object whose matrix has at most `n` rows and `n` columns.
`RandomMorphismWithFixedSource( M, n )` should return a matrix whose range has `n` relations and its number of generators is between between `n/2` and `3*n/2`.
`RandomMorphismWithFixedRange(M,n)` should return a matrix whose source has n generators and its number of generators is between `n/2` and `3*n/2`.
`RandomMorphismWithFixedSourceAndRange(M,N,n)` is installed only when the homalg ring is commutative and should return a morphism between `M` and `N`, which is a linear combination of the generators of external `Hom(M,N)` with coefficients are |n|-powers of random ring elements.
`RandomMorphism( cat, n )` will be derived from `RandomObject` and `RandomMorphismWithFixedSource`.

This PR is part 3 of https://github.com/homalg-project/CAP_project/pull/239


kamalsaleh:
Better methods can be added using https://github.com/homalg-project/CAP_project/pull/1091 => This PR is closed.

PR #289, opened 2019-03-26
Adding random methods to the category of graded left/right presentations over graded homalg ring.

kamalsaleh:
`RandomObject( cat, n )` should return an object whose matrix has at most `n` rows and `n` columns.
`RandomMorphismWithFixedSource( M, n )` should return a matrix whose range has n relations and its number of generators is between between `n/2 `and `3n/2`.
`RandomMorphismWithFixedRange(M,n)` should return a matrix whose source has `n` generators and its number of generators is between `n/2` and `3n/2`.
`RandomMorphism( cat, n )` will be derived from `RandomObject` and `RandomMorphismWithFixedSource`.

This PR is part 4 of https://github.com/homalg-project/CAP_project/pull/239


kamalsaleh:
This is now obsolete by https://github.com/homalg-project/CAP_project/pull/1091

PR #296, opened 2019-03-29
More pictures for the docu of CAP

HereAround:
This adds pictures for all universal objects in the documentation.

### DIFF
--- a/ActionsForCAP/PackageInfo.g
+++ b/ActionsForCAP/PackageInfo.g
@@ -19,6 +19,8 @@ Version := Maximum( [
 
 Date := ~.Version{[ 1 .. 10 ]},
 Date := Concatenation( ~.Date{[ 9, 10 ]}, "/", ~.Date{[ 6, 7 ]}, "/", ~.Date{[ 1 .. 4 ]} ),
+License := "GPL-2.0-or-later",
+
 
 Persons := [
   rec(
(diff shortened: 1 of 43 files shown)


sebastianpos:
Thank you for the effort. I would like to ask you to make an adjustment to the picture for the image/coimage. In the current picture, it is hard to see that `c` and `iota` form a factorization of `alpha`.
Factorizations are usually depicted as triangles (see e.g. Definition 1.31 [here](https://dokumentix.ub.uni-siegen.de/opus/volltexte/2017/1179/pdf/Dissertation_Sebastian_Posur.pdf)).
It would be great if you could adjust image/coimage accordingly.

HereAround:
No problem - better?

sebastianpos:
You drew the arrow `tau_1` in the picture for coimages twice (one shorter than the other) in order to obtain a twoheadarrow. However, this strategy led to a strange look when this arrow is bended (it also has the effect that straight arrows with twoheadarrows appear to be thicker).
Since I don't know how to construct twoheadarrows with the style for arrow tips that you used, I had to replace them. 

If you know how to create twoheadarrows with your previous style, please revert my commit. Otherwise, please adjust the arrowtips in the other diagrams accordingly.

Moreover, I halved the size of the coimage diagram, adjusting it to the size we agreed on in your previous commit (see `IsDominating` / `IsCodominating` ). It would be great if you could do the same for the other diagrams as well.

HereAround:
All done. Please note that I defined a new arrow tip, which I called "twohead". The 'problematic' arrow in the picture for coimage has thus been replaced by \draw[-twohead] ... I also adjusted the picture for codomination accordingly.

Further suggestions?

HereAround:
Your welcome. Also the suggestions look very nice indeed and I agree to them.

Still, for some reason it says above that I "dismissed stale reviews..." - this was unintended. Therefore, please check, if this causes some problem. Otherwise, please merge.

PR #299, opened 2019-04-01
Fix bug for GradedRows/Cols

sebastianpos:
Creation of multiple instances of these categories caused an error.

@kamalsaleh 
Can you confirm that this solves the problem that you pointed out?


kamalsaleh:
Yes, that solved the problem.

PR #305, opened 2019-04-05
install generic Display method for CAP morphisms

heiderich:
that displays its source and range

### DIFF
--- a/CAP/gap/CategoryMorphisms.gi
+++ b/CAP/gap/CategoryMorphisms.gi
@@ -566,3 +566,14 @@ InstallMethod( String,
     return Concatenation( "A morphism in ", Name( CapCategory( morphism ) ) );
     
 end );
+
+InstallMethod( Display,
+               [ IsCapCategoryMorphism and HasCapCategory ],
+  function( morphism )
+
+    Print( Concatenation( "A morphism in ", Name( CapCategory( morphism ) ) ) );
+    Print( "\n\nSource:\n");
+    Display( Source( morphism ) );
+    Print( "\n\nRange:\n");
+    Display( Range( morphism ) );
+end );
--- a/CAP/gap/CategoryMorphisms.gi
+++ b/CAP/gap/CategoryMorphisms.gi
@@ -568,7 +568,7 @@ InstallMethod( String,
 end );
 
 InstallMethod( Display,
-               [ IsCapCategoryMorphism and HasCapCategory ],
+               [ IsCapCategoryMorphism ],
   function( morphism )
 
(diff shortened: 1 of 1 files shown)


CAP/gap/CategoryMorphisms.gi
```diff
@@ -566,3 +566,14 @@ InstallMethod( String,
     return Concatenation( "A morphism in ", Name( CapCategory( morphism ) ) );
     
 end );
+
+InstallMethod( Display,
+               [ IsCapCategoryMorphism and HasCapCategory ],
```
sebastianpos:
```suggestion
               [ IsCapCategoryMorphism ],
```

I suggest to delete the filter `HasCapCategory` here.
- Advantage: in instances of categories for which we provide the `Display` method, this generic `Display` will not be judged by GAP as being "more specific"
- Disadvantage: this method will throw an error if you input a morphism that doesn't have a Cap category.

In my opinion, the disadvantage is not too bad since the error message will be: no method found for `CapCategory`, which exactly points to the semantical error someone made in such a situation:
every GAP object in the filter `IsCapCategoryMorphism` is supposed to have a `CapCategory`.

sebasguts:
This is not only an improvement, but also removes functionality. 

CAP currently displays more information about the morphism, which is lost using that method. I would suggest adding `Source` and `Range` to the [morphisms generic view setup](https://github.com/homalg-project/CAP_project/blob/master/CAP/gap/CategoryMorphisms.gi#L484) and not simply overwriting the display method. I am not sure if we might need to slightly extend the generic view functionality in tools for homalg, though.

I am kind of wondering how useful it is, too. If one does not overload morphism displays, one might also not overload object displays, so the output would somehow look like
```
A morphism in C
Source:
An object in C
Range:
An Object in C
```
 which I would regard worse than the current display output.

sebastianpos:
I think the intention was to improve the generic `Display` function for morphisms in situations where the `Display` function for objects are overloaded by the user but not necessarily the morphism `Display` function, e.g., in situations where there is no actual morphism datum (like for thin categories) or in a preliminary stadium of your implementation.

But even if you don't overload anything, I would consider this a useful feature, since an output like
```
A zero morphism in C
Source:
An object in C
Range:
A zero object in C
```
is more specifc than
`A zero morphism in C`

How hard would it be to extend the generic view mechanism?
 If we don't want to extend it, would it be also okay to replace 
`Print( Concatenation( "A morphism in ", Name( CapCategory( morphism ) ) ) );`
 with
`Print( String( CapCategory( morphism ) ) );`
?

PR #307, opened 2019-04-08
Adding Derivation for SubtractionForMorphisms in Ab-categories using

kamalsaleh:
AdditiveInverse & AdditionForMorphisms.

### DIFF
--- a/CAP/gap/DerivedMethods.gi
+++ b/CAP/gap/DerivedMethods.gi
@@ -674,6 +674,18 @@ end : Description := "IsInitial using IsZeroForObjects",
       CategoryFilter := IsAdditiveCategory ); #Ab-Category?
 
 ##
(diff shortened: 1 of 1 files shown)


sebastianpos:
This derivation is already part of the CAP code:

https://github.com/homalg-project/CAP_project/blob/ee40337589b3fc193919a63b69bfe6aeed51b66f/CAP/gap/DerivedMethods.gi#L1025-L1032

Why did you think that it was missing?

kamalsaleh:
This is my mistake. I expected it in an example of mine to be derived when a category has addition and additive inverse, so I added it their manually thinking there is no derivation for subtraction. Then after many days I wrote this derivation "with the correct required property (Ab-category)" . Now, I see that my category in the example didn't have the property Ab-category.

sebastianpos:
Okay, thank you for the clarification.

PR #311, opened 2019-04-08
Monoidal structure for FreydCategoriesForCAP

HereAround:
This PR attempts to make the entire functionality of https://github.com/HereAround/CAPPresentationCategory available within FreydCategories for CAP. Changes include the following:

- monoidal structure implemented for FreydCategories
- extended display methods by "FullInformation"
- functions IsValidInputForFreydCategory and IsMonoidalStructurePresent have been added
- by use of these, the constructor of FreydCategory has been extended
- the example for graded rows/columns has been split to test their Freyd categories in a separated file
- test for FreydCategory( gradedColumns ) has been added
- bugfix in universal morphism from/into zero object of graded rows/column - arguments of this function were given in wrong order, leading to a "wrong number of columns/rows" error in extended tests

This is admittedly a fairly large PR. However, the major change is the implementation of the monoidal structure. The other points then derive from this change, except for the extended display methods. I have added them, since they mean only a minor change and we thus bring all the functionality of
https://github.com/HereAround/CAPPresentationCategory into FreydCategories with a single PR.

Among others we will need to discuss the following:
- well-definedness of associators
- is "rigid and symmetric closed monoidal structure" required for monoidal structure of Freyd category (-> method "IsMonoidalStructurePresent" need to be adjusted correspondingly)

We also need to think about "ProjectionOfBiasedWeakFiberProduct". Is this a required operation for any category that we want to build a FreydCategory upon?
- If yes, then the category cat at line 29 in the example file SnakeLemmaProof.g is not a valid input. 
- If no, then all uses of "ProjectionOfBiasedWeakFiberProduct" in FreydCategory.gi need to be grouped within an if-statement. Currently, e.g. line 681 in FreydCategory.gi is an exception to this - it installs (unless I am mistaken) "EpimorphismFromSomeProjectiveObjectForKernelObject" for all input categories by use of "ProjectionOfBiasedWeakFiberProduct".

Currently, I have therefore outcommented line 1535 in FreydCategories.gi. Once we find a solution for this, this has to be adjusted. Also, we should discuss the operations listed in lines 1517 to 1534 and 1583 to 1596 and identify a minimal set among these which are required by the underlying category to qualify as input category upon which we can define basic functionality/monoidal structure in its Freyd category.

[Examples, which test the monoidal structure implemented in this PR are not yet present in the examples. I would propose to use the tests of https://github.com/HereAround/PresentationsByProjectiveGradedModules. This however, is only possible after I have made this category available within FreydCategoriesForCAP...]

PR #315, opened 2019-04-09
Adding graded module presentation to Freyd categories

HereAround:
This PR installs the categories SfpgrLeft and SfpgrRight with special names, constructors and display methods. Also a specialized and (presumably) faster algorithm to compute InternalHom is prepared.

However, the latter depends of course on the monoidal structure. That one I added in the PR https://github.com/homalg-project/CAP_project/pull/311. These structures are, however, not yet present in the official FreydCategories-package. Hence, the corresponding lines are currently outcommented. However, eventually this should be changed.

Also note, that I have not yet added tests for this new category. The reason is, that I split the example file for graded rows/columns in https://github.com/homalg-project/CAP_project/pull/311. I thus propose to first discuss this other PR. Once it can (hopefully) be accepted, I can update/extend the tests for Freyd( graded rows/columns ) by using the test files of my package https://github.com/HereAround/PresentationsByProjectiveGradedModules. The latter can either happen within this PR (if desired) or in a separate PR.


HereAround:
I have update this PR. First of all, I am cherry-picking the following PRs:
- https://github.com/homalg-project/CAP_project/pull/329
- https://github.com/homalg-project/CAP_project/pull/332
- https://github.com/homalg-project/CAP_project/pull/334
- https://github.com/homalg-project/CAP_project/pull/335
Based on this I then install the grade module presentations into FreydCategory. Regarding the opening post, only the tests remain an open issue. I will add them soon.

PR #316, opened 2019-04-10
Added further final derivations

sebasguts:


sebasguts:
Things that need to be discussed:
* Preconditions check?
* Additional preconditions?

sebastianpos:
There are two more issues desirable to address:
- this PR makes given mulitplicities in final derivations superfluous and they should thus be deleted from the interface
- final derivations are marked as being installed primitively (apparently this has always been the case). This should be fixed.

If you agree, I will merge this PR and we can address these two issues in separate PRs.

Issue #318, opened 2019-04-15
Saturation of derivations fails if conditions are given in CategoryFilter


sebastianpos:
Suppose we have derivations L and S and a primitively installed operation P.
Let L and S both depend on P. Moreover, let S also depend on L, but this dependence is written in the `CategoryFilter.` Then, it depends on the ordering of the derivations whether S can be computed in the category or not.

I give a minimal example in which the ordering is chosen such that the category cannot compute S.
```
LoadPackage( "CAP" );

##
AddDerivationToCAP( SolveLinearSystemInAbCategory,
  function( left_coefficients, right_coefficients, right_side )
    return PreCompose( left_coefficients, right_side );
  end :
  CategoryFilter := function( cat )
    
    if CanCompute( cat, "Lift" ) then
            
            return true;
            
    fi;
    
    return false;
    
  end );

##
AddDerivationToCAP( Lift,
  function( alpha, beta )
    
    return PreCompose( alpha, beta );
    
end );

category := CreateCapCategory( );

AddPreCompose( category, ReturnTrue );

Finalize( category );
```

We get
```
gap> CanCompute( category, "SolveLinearSystemInAbCategory" );
false
```

sebasguts:
Yes, this is somehow intentional, and I am not sure I would regard it a bug.

Derivations are ordered in a graph, but only with their preconditions. Since you put `Lift` not in the preconditions, but in the `CategoryFilter`, for CAP, both derivations have the same precondition, namely `PreCompose`. Now, if you add `PreCompose`, the derivation mechanism tries to build a spanning tree for possible derivations, and while building it, `Solve...` does not get installed, because the `CategoryFilter` is not fulfilled.

Then `Lift` gets added, but since `Lift` is no precondition of `Solve...`, the system is not going to check if it can now apply the derivation for `Solve...`.

All in all, I would say this is simply a misuse.

sebastianpos:
Unfortunately, putting preconditions inside the `CategoryFilter` of a derivation for a category **A** is the only tool we have in order to check whether some preconditions hold in some other category **B**.
This principle was first used here: https://github.com/homalg-project/CAP_project/blob/d6ead0813b51dbbf5901b46347d3959e88fbe172/CAP/gap/DerivedMethods.gi#L2449

Now, it may happen that the categories **A** and **B** coincide for some instances where the cited derivation is triggered, and differ for other instances. If they coincide, we arrive at a situation depicted in the minimal example above.

[...]

PR #322, opened 2019-04-17
Reevaluate weight_list while finalizing

sebastianpos:


sebastianpos:
Looks good for me, and it solves the problem in #318.
You may merge it if you like.

PR #323, opened 2019-04-17
New display methods for (graded) rows and columns

HereAround:
- For a morphism of (graded) rows and columns, the Display method now also display source and range of this morphism.
- The tests were adjusted accordingly.


HereAround:
Thanks for the feedback. I considered it easiest, to integrate your suggestions into this PR. Therefore, I have just pushed two commits which integrate your suggested changes.

HereAround:
Two more changes that come handy in this PR:
1. I removed the method "FullInformation" in graded rows/columns. With the previous changes it is identical to the Display-method.
2. I have also adjusted the display-method of Freyd category. The latter now uses Print( StringMutable ( ... ) ) to first display general information on a morphism/object in a Freyd category. Subsequently, detailed information on the relation morphism/morphism datum are being displayed. The tests are adjusted accordingly.

PR #324, opened 2019-04-21
Bugfix in AddUniversalMorphismFromZeroObjectWithGivenZeroObject in category of graded columns

HereAround:
Arguments for this function were provided in wrong order


FreydCategoriesForCAP/gap/CategoryOfGradedRowsAndColumns/CategoryOfGradedColumns.gi
```diff
@@ -435,7 +435,7 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_CAP_CATEGORY_OF_GRADED_COLUMNS,
     # @Returns a morphism
     # @Arguments zero_object, object
```
zickgraf: Probably this line should be changed, too.

HereAround: The lines are not used to generate a documentation via autodoc (yet). So this description does not influence the functionality.

However, you are right. For reasons of consistency this should be changed. I have just done so.

Thanks for pointing this out.

PR #343, opened 2019-06-03
Do not trigger method reordering for installation of logical implications

sebasguts:
I decided not to use the official API functions `Suspend/ResumeMethodReordering` but instead set the controlling flag manually, to not retrigger any reordering before the category is finalized. Functions using the new flags should not be called before finalizing.

In case this is considered too bad of a hack, we can go back to using the official functions or provide a similar way of suspending/resuming method reordering in the GAP kernel.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -9,7 +9,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
-  "2019.06.02", ## Sebas' version
+  "2019.06.03", ## Sebas' version
   ## this line prevents merge conflicts
   "2019.02.26", ## Sepp's version
   ## this line prevents merge conflicts
--- a/CAP/gap/LogicForCAP.gi
+++ b/CAP/gap/LogicForCAP.gi
@@ -469,7 +469,9 @@ InstallGlobalFunction( INSTALL_LOGICAL_IMPLICATIONS_HELPER,
                        
   function( category, current_filter )
     local i, theorem_list, current_theorem;
-    
+
+    SuspendMethodReordering();
+
     for i in category!.logical_implication_files.Propositions.( current_filter ) do
         
         theorem_list := READ_THEOREM_FILE( i );
@@ -493,6 +495,8 @@ InstallGlobalFunction( INSTALL_LOGICAL_IMPLICATIONS_HELPER,
         od;
         
     od;
+
+    ResetMethodReordering();
     
 end );
 
--- a/CAP/gap/CAP.gi
+++ b/CAP/gap/CAP.gi
@@ -156,16 +156,14 @@ InstallGlobalFunction( CREATE_CAP_CATEGORY_FILTERS,
                        
   function( category )
     local name, cell_filter, filter_name, filter;
-    
-    SuspendMethodReordering();
 
     name := Name( category );
     
-    filter_name := NewFilter( Concatenation( name, "InternalCategoryFilter" ) );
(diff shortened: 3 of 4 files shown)


sebasguts:
Now looking at CircleCI, the feature I am using does not seem to be in the current GAP release. I suggest not merging it before it is released.

CAP/PackageInfo.g
```diff
@@ -100,7 +100,7 @@ PackageDoc := rec(
 ),
 
 Dependencies := rec(
-  GAP := ">= 4.9.1",
+  GAP := ">= 4.10.0",
```
mohamed-barakat: Is it possible to remain backwards compatible? See below

CAP/gap/Finalize.gi
```diff
@@ -194,6 +194,8 @@ InstallMethod( IsFinalized,
         derivation_list := derivation_list{ Difference( [ 1 .. Length( derivation_list ) ], current_installs ) };
         
     od;
+
+    RECALCULATE_ALL_METHOD_RANKS();
```
mohamed-barakat: if CompareVersionNumbers( GAPInfo.Version, "4.10.0" ) then
    ValueGlobal( "RECALCULATE_ALL_METHOD_RANKS" )();
fi;

CAP/gap/LogicForCAP.gi
```diff
@@ -470,6 +470,9 @@ InstallGlobalFunction( INSTALL_LOGICAL_IMPLICATIONS_HELPER,
   function( category, current_filter )
     local i, theorem_list, current_theorem;
     
+    REORDER_METHODS_SUSPENSION_LEVEL := REORDER_METHODS_SUSPENSION_LEVEL + 1;
```
mohamed-barakat: See above.

CAP/gap/LogicForCAP.gi
```diff
@@ -493,6 +496,9 @@ InstallGlobalFunction( INSTALL_LOGICAL_IMPLICATIONS_HELPER,
         od;
         
     od;
+
+    REORDER_METHODS_SUSPENSION_LEVEL := REORDER_METHODS_SUSPENSION_LEVEL - 1;
```
mohamed-barakat: See above.

CAP/gap/Finalize.gi
sebasguts: Currently, this patch is only valid for master/4.11

mohamed-barakat:
With this PR the Wall computation time of the main procedure `ConstructibleProjection` went down from 28 sec. to 10 sec. With GAP 4.9.2 it is 4 sec.

fingolfin:
With your 2x2x2 example, if I call `SuspendMethodReordering();` once before running it (and without this PR!), timing goes down from 40 to 5 seconds for me.

sebasguts:
Indeed, there are still some reorderings triggered now, basically for every category. With my current patch, it goes down to 6 seconds on my computer. I will try to use the save way and see how this turns out.

CAP/PackageInfo.g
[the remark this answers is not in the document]
fingolfin: Sorry, I wrote these comments a couple hours ago (before Mohamed's), but then forgot to submit them (again...)

CAP/gap/Finalize.gi
mohamed-barakat: Then
```
if IsBoundGlobal( RECALCULATE_ALL_METHOD_RANKS ) then
    ValueGlobal( "RECALCULATE_ALL_METHOD_RANKS" )();
fi;
```

PR #344, opened 2019-06-05
Display func for cap morphism

sebasguts:


zickgraf:
This PR introduces a discrepancy between the View method (which prints properties like mono, epi, etc.) and the Display method (which does not print such properties after this PR). Since the original authors are not working on CAP anymore, since no-one seems to have missed the functionality of this PR since three years, and since it's easy to reproduce if one needs it, I will close the PR.

PR #345, opened 2019-06-05
Use ResumeMethodReordering instead of ResetMethodReordering

sebasguts:


fingolfin:
I approve, but can't actually approve, so hopefully @mohamed-barakat will do that

PR #354, opened 2019-06-12
d&i AffineDimension for IsGradedLeftOrRightPresentation

mohamed-barakat:
This is now possible:

FullSubcategoryByMembershipFunction( Sgrmod, M -> AffineDimension( M ) <= 0 );

### DIFF
--- a/GradedModulePresentationsForCAP/PackageInfo.g
+++ b/GradedModulePresentationsForCAP/PackageInfo.g
@@ -18,7 +18,7 @@ Version := Maximum( [
            ##
            "2018.03.20", # Sepps version
            ##
-           "2018.12.24", # Mohamed's version
+           "2019.04.03", # Mohamed's version
            ##
            ] ),
 Date := ~.Version{[ 1 .. 10 ]},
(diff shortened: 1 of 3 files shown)


mohamed-barakat:
Ping?

sebastianpos:
What exactly do you mean by the affine dimension of a graded module?

mohamed-barakat:
It is the Krull dimension of the annihilator ideal of the underlying nongraded module.

sebastianpos:
But then I suppose there should be a distinction in the code between left and right modules.

mohamed-barakat:
Yes, you are right, I mistakenly used the general filter for declaration in the installation. Will fix.

mohamed-barakat:
Fixed and forced pushed the changes.

sebastianpos:
Thanks. Could you also please update the documentation (i.e., describe what the affine dimension is) and provide an error handling for the case when the ring `ring_for_representation_category` is not commutative?

mohamed-barakat:
Done.

PR #355, opened 2019-06-12
installed InfoStringOfInstalledOperationsOfCategory

mohamed-barakat:
Installed InfoStringOfInstalledOperationsOfCategory and used it for a better InfoOfInstalledOperationsOfCategory which now works in Julia and correctly displays the monoidal doctrines.

### DIFF
--- a/CAP/gap/PrintingFunctions.gd
+++ b/CAP/gap/PrintingFunctions.gd
@@ -11,4 +11,6 @@
 
 DeclareGlobalVariable( "CAP_INTERNAL_CATEGORY_PROPERTY_RANK_AND_STRING" );
 
+DeclareGlobalFunction( "InfoStringOfInstalledOperationsOfCategory" );
+
 DeclareGlobalFunction( "InfoOfInstalledOperationsOfCategory" );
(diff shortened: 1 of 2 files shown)


mohamed-barakat:
This updated PR leads to the following improved behavior of `InfoOfInstalledOperationsOfCategory`:

```gap
gap> Q := HomalgFieldOfRationalsInSingular( );
Q
gap> S := GradedRing( Q["x,y"] );
Q[x,y]
(weights: yet unset)
gap>
gap> Sgrmod := GradedLeftPresentations( S );
The category of graded left f.p. modules over Q[x,y] (with weights [ 1, 1 ])
gap>
gap> InfoOfInstalledOperationsOfCategory( Sgrmod );
40 primitive operations were used to derive 165 operations for this category which
* IsAbCategory
* IsMonoidalCategory
```

sebastianpos:
Moreover, I assume that the correct handling within Julia relies on the fact that `Display` is used rather than `Print.`

mohamed-barakat:
Yes, this is the reason.

PR #367, opened 2019-06-28
added example file CohP1 to GradedModulePresentationsForCAP

mohamed-barakat:


mohamed-barakat:
@sebasguts, @zickgraf: Why doesn't this first (test) example change the code coverage?

zickgraf:
> Probably because `GradedModulePresentationsForCAP` is not enlisted in `CAP_project/makefile`.

Correct. `GradedModulePresentationsForCAP` is not tested in the CI. If you want to test it in the CI, but don't want to change the packages tested when running `make test`, you can add it to `ci-test` in the makefile.

mohamed-barakat:
> Correct. `GradedModulePresentationsForCAP` is not tested in the CI. If you want to test it in the CI, but don't want to change the packages tested when running `make test`, you can add it to `ci-test` in the makefile.

Thanks @zickgraf.

@sebasguts, @sebastianpos: Is it reasonable to add `GradedModulePresentationsForCAP` to the central `make test` and  activate `ci-test` in `makefile`?

sebasguts:
@mohamed-barakat Yes, it would be good to also run these tests in the CI. Afaik one can use Singular in the tests, and we also already do this, so this is fine. You could add the lines in the main makefile to this PR, so we see the tests.

sebastianpos:
Could you please rebase this PR to the current master? This should make the CI work again. Thanks.

mohamed-barakat:
Thanks, done. I hope it works now.

sebastianpos:
Thank you. Now, it works fine.

PR #372, opened 2019-08-02
Add filter types list_of_objects/morphisms/twocells

zickgraf:
It would be good to check the substitutions (`IsList` -> `list_of_objects` or `list_of_morphisms`) which are not covered by the validator a second time. Note that I have excluded some occurrences of `IsList` in the method record (e.g. `HomologyObjectFunctorialWithGivenHomologyObjects`) since I'm not sure if the semantics would be appropriate there.

### DIFF
--- a/CAP/gap/MethodRecord.gi
+++ b/CAP/gap/MethodRecord.gi
@@ -517,7 +517,7 @@ ZeroMorphism := rec(
 
 DirectSum := rec(
   installation_name := "DirectSumOp",
-  filter_list := [ IsList, "object" ],
+  filter_list := [ "list_of_objects", "object" ],
   argument_list := [ 1 ],
   universal_type := "LimitColimit",
   return_type := "object",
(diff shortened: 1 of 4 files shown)


CAP/gap/MethodRecord.gi
```diff
@@ -2240,7 +2240,7 @@ AstrictionToCoimageWithGivenCoimage := rec(
 
 UniversalMorphismIntoCoimage := rec(
   installation_name := "UniversalMorphismIntoCoimage",
-  filter_list := [ "morphism", IsList ],
+  filter_list := [ "morphism", "list_of_morphisms" ],
```
sebastianpos: Depending on your application of this new specification, you may have to be careful here. This list of morphisms is not of arbitrary length, but has to be of length 2 (it stands for a possible factorization of the first argument).

CAP/gap/MethodRecord.gi
```diff
@@ -2545,7 +2545,7 @@ CoastrictionToImageWithGivenImageObject := rec(
 
 UniversalMorphismFromImage := rec(
   installation_name := "UniversalMorphismFromImage",
-  filter_list := [ "morphism", IsList ],
+  filter_list := [ "morphism", "list_of_morphisms" ],
```
sebastianpos: Same applies here (and of course in the WithGiven cases)

CAP/gap/MethodRecord.gi
zickgraf: You are right, I have reverted the four occurrences. Thanks for noticing!

CAP/gap/MethodRecord.gi
zickgraf: See above

PR #378, opened 2019-08-09
 Add general sanity checks 

zickgraf:
~The first two commits are from #372.~


zickgraf:
Thanks for noticing! I did not expect a large performance drop and thus did not look at the performance impact, which obviously was a bad idea.....

I investigated this a bit and found one main culprit: it turns out, concatenating and comparing strings in GAP is quite expensive. This amounts to about 3/4 of the additional time taken. I will provide a follow-up PR reducing this overhead.

Additionally, I could try to cache results but I'm not sure this would be worth the additional complexity.

sebastianpos:
I don't think it is necessary to add too much complexity to the testing. Whenever one is interested in speed, one should disable the type checking. So if you can provide some simple improvements that will be fine.

PR #380, opened 2019-09-15
added rule: composition of isos is an iso

mohamed-barakat:


sebastianpos:
Do you have an explanation for the drastic decrease of the code coverage?

mohamed-barakat:
No, unfortunately not.

PR #389, opened 2019-09-24
conditional installation of IsWellDefinedFor*

mohamed-barakat:
only install IsWellDefinedForObjects and IsWellDefinedForMorphisms
if CanCompute( underlying_category, "IsWellDefinedForMorphisms" )


mohamed-barakat:
I did not adapt the indentation on purpose for the difference to be obvious. If required, I can add a commit that fixes the indentation.

sebastianpos:
It's okay, thank you.

PR #401, opened 2019-10-09
Make dependency of Freyd on QPA2 optional

sebastianpos:


zickgraf:
If I remember correctly, the dependency on QPA2 was the reason why we did not include Freyd in the CI tests (because contributors would have to install QPA2 to investigate test failures). If this hard dependency is gone, maybe Freyd could be added to the CI (should be completely analogous to #376).

sebastianpos:
At this point, the current tests still depend on QPA2.
Do you know a good strategy to exclude certain test files if a package is not loaded?

zickgraf:
I know a strategy, but I would certainly not call it "good", see https://github.com/homalg-project/FinGSetsForCAP/blob/master/examples/ReconstructGroup.g and look for `IsPackageMarkedForLoading( "FinSetsForCAP", ">= 2018.09.17" )`. However, when using the new `extract_examples` option of AutoDoc there might exist better strategies. I will think about this.

Question: are there enough tests NOT depending on QPA2 to make this worth it?

sebastianpos:
Okay, the following test files do not depend on QPA2:
```
CategoryOfColumns.g
CategoryOfRows.g
CokernelImageClosureExample.gi
FreydCategoryOfGradedRowsAndColumns.g
GradedRowsAndColumns.g
HomOfFpFunctors.g
StandardTensorStructure.g
```
So yes, I would say it is worth it.

PR #402, opened 2019-10-09
Add interface for a categorical Simplify command

sebastianpos:
In this draft, I suggest a categorical version of a `Simplify` command. The idea is to create "simpler" but isomorphic representations of objects, and "simpler" but congruent versions of morphisms,  
with a complexity of the simplification controlled by a positive integer (or infinity). Moreover, convenience methods that use the infinity values are provided.

As applications, I imagine 
- an analog of homalgs `ByASmallerPresentation` (but without any side effect)
- Canceling of denominators in matrices with fractions
- Usage of `SimplifiedObject` within categorical algorithms whenever an object can be replaced with an isomorphic one

For now, I only suggest a `SimplifiedMorphism` version which fixes its source and its range,
but maybe variants like `SimplifiedMorphismWithVaryingSource`,
`SimplifiedMorphismWithVaryingRange`,
`SimplifiedMorphismWithVaryingSourceAndRange` are also interesting
(e.g., in the category of rows over the integers, `SimplifiedMorphismWithVaryingSourceAndRange`
could yield a Smith normal form, `SimplifiedMorphismWithVaryingSource` could yield a Hermite normal form).

Also, I did not yet specify if complexity `infinity` should yield an idempotent operation
or if the complexity `1` should yield the identity.

Any thoughts on this topic are welcome.

### DIFF
--- a/CAP/examples/testfiles/Simplify.gi
+++ b/CAP/examples/testfiles/Simplify.gi
@@ -1,0 +1,112 @@
+#! @Chapter Examples and Tests
+
+#! @Section Homomorphism structure
+
+#! @Example
+ReadPackage( "CAP", "examples/testfiles/StringsAsCategory.gi" );;
+C := StringsAsCategory();;
+obj1 := StringsAsCategoryObject( "qaeiou", C );;
+obj2 := StringsAsCategoryObject( "qxayeziouT", C );;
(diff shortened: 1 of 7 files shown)


sebastianpos:
Now, I provide a complete interface for the simplification of morphisms.
This includes five cases in which a morphism can be considered as an object
in an ambient category.

PR #405, opened 2019-10-28
Option in category constructor for avoiding any overhead

sebastianpos:
The idea is to provide an option in the category constructor that reduces CAP's overhead to zero, e.g., when the option is enabled, the call of `KernelLift` directly triggers GAP's method selection in order to search for the function that was given by the user (or by the derivation machinery).

### DIFF
--- a/CAP/examples/testfiles/FieldAsCategory.gi
+++ b/CAP/examples/testfiles/FieldAsCategory.gi
@@ -38,7 +38,7 @@ InstallMethod( FieldAsCategory,
   function( field )
     local category;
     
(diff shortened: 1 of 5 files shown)


zickgraf:
In PR #378 I'm also touching `CapInternalInstallAdd`. Since your change seems to be "wrap the existing code in an `if` and add an `else` case" I think it would be much simpler to rebase this PR on #378 than the other way round. So, could you maybe merge #378 first?

sebastianpos:
Okay.

sebastianpos:
I rebased this PR to master (which was unfortunately harder than I thought).
I also changed the indentation such that the diff is now easier to read.

PR #406, opened 2019-10-28
Add mandatory filters to type for obj/mor

sebastianpos:
If I understand the documentation of `ObjectifyWithAttributes` correctly, this might lead to a speed-up.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -11,7 +11,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2019.06.05", ## Sebas' version
   ## this line prevents merge conflicts
-  "2019.10.04", ## Sepp's version
+  "2019.10.29", ## Sepp's version
   ## this line prevents merge conflicts
   "2019.08.10", ## Fabian's version
   ## this line prevents merge conflicts
--- a/CAP/examples/testfiles/FieldAsCategory.gi
+++ b/CAP/examples/testfiles/FieldAsCategory.gi
@@ -50,7 +50,7 @@ InstallMethod( FieldAsCategory,
     
     AddObjectRepresentation( category, IsFieldAsCategoryObject );
     
-    AddMorphismRepresentation( category, IsFieldAsCategoryMorphism );
+    AddMorphismRepresentation( category, IsFieldAsCategoryMorphism and HasUnderlyingFieldElement );
 
     INSTALL_FUNCTIONS_FOR_FIELD_AS_CATEGORY( category );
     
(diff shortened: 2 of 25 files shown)


zickgraf:
I'm not sure this is a good idea: Using this PR, if one accidentally creates a morphism `phi` without a source, `HasSource( phi )` is true but `Source( phi )` throws an error. I think this makes debugging errors a lot harder, and makes sanity checks throw weird instead of meaningful errors. Additionally, I'm not sure if there can be a speed-up in reality: I guess in almost all applications morphisms have more attributes than just a category, source and range. Since these attributes are not part of the type, I assume that in these cases the fallback to ordinary Objectify mentioned in the documentation is triggered anyway.

sebastianpos:
Good objections.

Concerning the first:
A morphism can only get its CAP-Type (the one I manipulate in this PR) if it runs through `ObjectifyMorphismForCAPWithAttributes`. But that function tries to access `Source/Range` immediately after the creation of the Gap-Object (provided you haven't disabled the predicate logic), so any accidentally missing Source/Range leads to an error right away.

Concerning the second: 
Any additional attributes that occur in your personal constructor should be added to the type (if you care about speed) as follows:
`AddMorphismRepresentation( category, IsMyCatMorphism and HasMyPersonalAttribute )`

zickgraf:
Regarding my first concern:
I have an idea/suggestion: what about a function `ObjectifyMorphismWithSourceAndRangeForCAPWithAttributes` with argument signature `morphism, category, source, range[, additional_attribute1, value1, ...]`. This would avoid any sanity checks but make sure source and range are always given, even if the logic is turned off. `ObjectifyMorphismForCAPWithAttributes` could simply be a (possibly deprecated?) wrapper around `ObjectifyMorphismWithSourceAndRangeForCAPWithAttributes`.

Regarding my second concern:
Ah, I see :-)

sebastianpos:
I like the idea! This should possibly have been the signature of `ObjectifyMorphismForCAPWithAttributes` from the beginning.

But since `ObjectifyMorphismForCAPWithAttributes` is used everywhere, we have to use a different name like you suggested for the new signature function: `ObjectifyMorphismWithSourceAndRangeForCAPWithAttributes`.

Yes, the other should be marked as deprecated. I would not advice to use it as a wrapper for `ObjectifyMorphismWithSourceAndRangeForCAPWithAttributes` since that would mean to search for `Source/Range` within the input list, but simply to leave it as it is with a date for its deletion (a year from now).

sebastianpos:
I provided the necessary changes.

The current tests in Freyd call Algebroids which doesn't use the new function yet, thus the deprecation warning is triggered.

Once this PR is merged, I will provide a PR for Algebroids.

PR #408, opened 2019-10-29
defined CAP_INTERNAL_METHOD_NAME_LIST_FOR_ATTRIBUTE_CATEGORY

mohamed-barakat:
Based on our discussion I have added opt-in lists for the category constructors looping over the entries in the method record.

@sebastianpos: Is this draft PR compatible with what you have imagined?


sebastianpos:
Are you sure you have provided the intended commits?
I only see a big list of functions that is intersected with CAP's method record,
but no opt-in lists.

mohamed-barakat:
This is meant as the opt-in list :)

sebastianpos:
I'm sorry but I don't understand.

For me, `AttributeCategories` is a tool to create new categories.
There should not be a global list for all categories that arise via `AttributeCategories`,
but each instance that is created via `AttributeCategories` should pipe its own opt-in list to `AttributeCategories`.

mohamed-barakat:
Thanks for the clarification, I now understand your objection. I will modify the PR accordingly.

sebastianpos:
Any news on this PR? Its current status still does not address my concerns. Probably, it can simply be closed?

mohamed-barakat:
Yes, this draft PR can be closed. I started new package called `CategoryConstructor` which should be used in `IntrinsicCategories`, `FunctorCategories`, `Bicomplexes`, `GradedCategories`, and the new package `SubcategoriesForCAP`. It is for me also a way to think and address your concerns.

PR #410, opened 2019-10-29
fixed bug in lift along mono/epi in ActionsCategory

mohamed-barakat:


mohamed-barakat:
The new package `GradedCategories` was able to detect a difference. With the current implementation I got an error that two morphisms are not composable and with this commit the error disappeared. I will prepare an example using the new packages `GradedCategories` and `InternalModules` (the latter is not yet on github).

sebastianpos:
You should first check if one of the categories in question do not satisfy CAP's specifications.

mohamed-barakat:
Yes, when I prepare the example I will check again if the mistake is on my side :)

mohamed-barakat:
There was a bug in my `AddCokernelObject` which was discovered by @zickgraf :) This PR was indeed obsolete.

sebastianpos:
Can I learn about this bug?

mohamed-barakat:
I did not dualize `KernelObject` for the attribute category in the correct way which resulted in noncomposable morphisms (due to sloppiness). The correct dualization of

https://github.com/homalg-project/InternalModules/blob/master/gap/SModules.gi#L59

is now line

https://github.com/homalg-project/InternalModules/blob/master/gap/SModules.gi#L83

PR #411, opened 2019-10-29
fixed serious bug in CAP_INTERNAL_GET_CORRESPONDING_OUTPUT_OBJECTS

mohamed-barakat:
causing

ProjectionInFactorOfFiberProductWithGivenFiberProduct( [ mor1, mor2 ], 1, P );

to return

ProjectionInFactorOfFiberProductWithGivenFiberProduct( [ mor1, mor2 ], 2, P );

### DIFF
--- a/CAP/gap/ToolsForCategories.gi
+++ b/CAP/gap/ToolsForCategories.gi
@@ -686,7 +686,11 @@ InstallGlobalFunction( CAP_INTERNAL_GET_CORRESPONDING_OUTPUT_OBJECTS,
             if ForAll( current_output[ 2 ], i -> i in "0123456789" ) then
                 list_position := Int( current_output[ 2 ] );
             else
-                list_position := function_input[ Position( input_list, current_output[ 2 ] ) ];
+                list_position := Position( input_list, current_output[ 2 ] );
+                if list_position = fail then
+                    Error( "unable to find ", current_output[ 2 ], " in input_list" );
+                fi;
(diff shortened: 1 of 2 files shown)


CAP/gap/ToolsForCategories.gi
```diff
@@ -686,7 +686,7 @@ InstallGlobalFunction( CAP_INTERNAL_GET_CORRESPONDING_OUTPUT_OBJECTS,
             if ForAll( current_output[ 2 ], i -> i in "0123456789" ) then
                 list_position := Int( current_output[ 2 ] );
             else
-                list_position := Position( input_list, current_output[ 2 ] );
+                list_position := function_input[ Position( input_list, current_output[ 2 ] ) ];
```
sebastianpos: If `Position` returns `fail` then we don't have a meaningful error message anymore.
I suggest to write
```
list_position := function_input[ list_position ];
```
right after the `list_position = fail` statement, i.e., in line 695.

mohamed-barakat: There are two assignments of `list_position` in the `if-else` statement. The first one is correct and the second was buggy.

Your suggestion will also alter the first correct assignment and break it.

sebastianpos: Yes, you are right. But this is easy to overcome. Could you please fix it?

mohamed-barakat: I added a commit with an error message.

PR #412, opened 2019-10-30
relax pre_function for MultiplyWithElementOfCommutativeRingForMorphisms

mohamed-barakat:
Allow MultiplyWithElementOfCommutativeRingForMorphisms to accept multiplication with GAP's rationals.

### DIFF
--- a/CAP/gap/MethodRecord.gi
+++ b/CAP/gap/MethodRecord.gi
@@ -1160,7 +1160,7 @@ MultiplyWithElementOfCommutativeRingForMorphisms := rec(
   
   pre_function := function( r, morphism )
     
-    if not r in CommutativeRingOfLinearCategory( CapCategory( morphism ) ) then
+    if not IsRat( r ) and not r in CommutativeRingOfLinearCategory( CapCategory( morphism ) ) then
       
       return [ false, "the first argument is not an element of the ring of the category of the morphism" ];
       


mohamed-barakat:
`MatricesForHomalg` coerces GAP rationals into arbitrary field elements (also finite fields) and issues an error if the coercion leads to a division by zero, in particular, the error handling will be passed over to `MatricesForHomalg`. So this PR is meant to introduce the same convenience as in homalg: `1/2 * singular_matrix` is then a short form for `((1/2) / singular_ring) * singular_matrix`.

But if you are not happy with coercing GAP rationals I can close the PR and do it elsewhere in my packages.

mohamed-barakat:
Then I will close the PR. Thanks for the clarification.

mohamed-barakat:
Just for the sake of transparency, since I control what "in" means for homalg rings, I can circumvent this PR one layer lower ;)

sebastianpos:
I strongly advice you **not** to allow 
`1/2 in F_3`
for `F_3` a `IsHomalgRing`. If you allowed it, we would lose our only instance of rings and ring elements in `GAP` that behave exactly as one would wish for in a typed setup.

mohamed-barakat:
I agree. I am also still reluctant. What about adding a new CAP operation

`MultiplyWithCoercibleRationalsForMorphisms`

which each category instance may or may not install?

sebastianpos:
Yes, that is the proper way to think about this problem.

If you do provide such a primitive, you have to think carefully about its specifications:
Who is in charge of testing coercibility?
If it is the user, then what is the output if coercibility fails? probably `fail`
If it is CAP, how do we provide such a test? Do `HomalgRings` all provide such a test?

mohamed-barakat:
> If you do provide such a primitive, you have to think carefully about its specifications:
> Who is in charge of testing coercibility?

I would suggest the user, and in case he is using homalg rings, he can delegate the test to the ring.

> If it is the user, then what is the output if coercibility fails? probably `fail`

No, I suggest raising a division by zero error.

> If it is CAP, how do we provide such a test? Do `HomalgRings` all provide such a test?

This is currently passed to the underlying CAS which will raise an error, but I can catch this earlier and raise an error in homalg, probably at the cost of performance.

sebastianpos:
If you suggest it's the user's responsibility to handle the coercibility AND you want to have a "division by zero"-error,
then you must specify that the user's function given to CAP yields `fail`, and then CAP needs to throw the "division by zero" error. 
If you don't do it like this, you would need to specify that the user has to raise a "division by zero" error, which is very uncomfortable.

sebastianpos:
Rather than defining a new primitive, I had another idea:

In #422 I simply propose a convenience function that is only triggered by the `*` operator. It demands an additional function from homalg rings.

mohamed-barakat:
I am happy with this idea. I have enhanced the homalg ring constructor to add a generic `interpret_rationals_func`. This corresponding commit in `MatricesForHomalg` references your PR #422. Thank you very much.

PR #414, opened 2019-10-31
The action/coaction category of a noncomputable category is again noncomputable

mohamed-barakat:
to allow having categories without decidable congruence of morphisms.

### DIFF
--- a/ActionsForCAP/gap/ActionsCategory.gi
+++ b/ActionsForCAP/gap/ActionsCategory.gi
@@ -521,6 +521,7 @@ end );
 InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
   
   function( category )
+    local underlying_category;
     
     ##
     AddIsEqualForCacheForObjects( category,
@@ -529,6 +530,10 @@ InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
     ##
     AddIsEqualForCacheForMorphisms( category,
       IsIdenticalObj );
+    
+    underlying_category := UnderlyingCategory( category );
+    
+    if CanCompute( underlying_category, "IsCongruentForMorphisms" ) then
     
     ##
     AddIsEqualForObjects( category,
@@ -540,6 +545,10 @@ InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
         
     end );
     
+    fi;
+    
+    if CanCompute( underlying_category, "IsEqualForMorphisms" ) then
+    
     ##
     AddIsEqualForMorphisms( category,
       function( morphism_1, morphism_2 )
@@ -548,6 +557,10 @@ InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
         
     end );
     
+    fi;
+    
+    if CanCompute( underlying_category, "IsCongruentForMorphisms" ) then
+
     ##
     AddIsCongruentForMorphisms( category,
       function( morphism_1, morphism_2 )
@@ -555,6 +568,8 @@ InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
         return IsCongruentForMorphisms( UnderlyingCell( morphism_1 ), UnderlyingCell( morphism_2 ) );
         
     end );
+    
+    fi;
     
 end );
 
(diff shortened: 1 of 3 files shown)


sebastianpos:
Could you please try the following:

Use the option `is_computable` introduced in #416 and pass it as
`is_computable := false` to the category for which you don't wish a derivation of `IsCongruentForMorphisms`. Please try if this works in your case (instead of commenting the corresponding final derivations).

If it works for you, I will merge #416, you can uncomment in this PR the corresponding final derivations and then rebase.

I would prefer this solution to commenting the final derivations since it breaks code that we don't see in this repo (e.g., `FinSets`).

mohamed-barakat:
#416 works for me, thank you very much, Sebastian.

mohamed-barakat:
Once you merge, I will rebase this PR. I will also undo the changes in `CAP/examples/testfiles/FieldAsCategory.gi`. So only `ActionsForCAP` will be affected by this PR.

ActionsForCAP/gap/ActionsCategory.gi
```diff
@@ -522,6 +523,10 @@ InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
     AddIsEqualForCacheForMorphisms( category,
       IsIdenticalObj );
     
+    underlying_category := UnderlyingCategory( category );
+    
+    if CanCompute( underlying_category, "IsCongruentForMorphisms" ) then
```
sebastianpos: There is a serious change in the behaviour of the action category due to this line.

If `underlying_category` happens to be not computable (i.e., no function for `IsCongruentForMorphisms`),
then `IsEqualForObjects` will be set to `ReturnFail`, making the proper functioning of the action category a gamble unless you don't set its caches to strict. Maybe your error that you mentioned in #410 is due to this change (which I assume you have already implemented locally).

Two possible tests:
1) Test if your bug disappears once you set the caching to strict
2) Test if your bug disappears if you change the notion of equality of objects in `ActionCategory` in a way that it relies on `IsEqualForMorphisms` instead of `IsCongruentForMorphisms` in the underlying category. I'm not sure yet if this is a proper solution, since I'm not sure why I haven't chosen this notion from the very beginning.

mohamed-barakat: The error did not disappear with the first test: `SetCachingOfCategoryCrisp( underlying_category )`.

mohamed-barakat: The error did not disappear with the second test:
```
    if CanCompute( underlying_category, "IsEqualForMorphisms" ) then

    ##
    AddIsEqualForObjects( category,
      function( action_object_1, action_object_2 )

        return IsEqualForObjects( ActionDomain( action_object_1 ), ActionDomain( action_object_2 ) )
               and
               IsEqualForMorphisms( StructureMorphism( action_object_1 ), StructureMorphism( action_object_2 ) );

    end );
```

sebastianpos: The first test is wrong. You don't set the caching of the underlying category to crisp, but of the category that you are manipulating in this PR, which is the action category.

mohamed-barakat: The error did not disappear after adding:

`SetCachingOfCategoryCrisp( action_category )`

sebastianpos: Then I'll need to actually see that error eventually.

Nevertheless, `SetCachingOfCategoryCrisp( action_category )` will be mandatory to add in the case where
`    if CanCompute( underlying_category, "IsCongruentForMorphisms" ) then`
yields `false.`

mohamed-barakat: This last comment should be added to the action category constructors. What do you think?

sebastianpos: Yes, you can add it there.

sebastianpos: Any news concerning this PR?
The only missing point seems still to be the caching, which should be set to crisp in the non-computable case.

mohamed-barakat: First of all the original title of this PR is by now misleading and should be changed.

Still, one of the points here is that `AddIsEqualForObjects` calls
```
IsCongruentForMorphisms( StructureMorphism( action_object_1 ), StructureMorphism( action_object_2 ) );
```
which makes no sense if `CanCompute( underlying_category, "IsCongruentForMorphisms" ) = false`.

The second point is that in this case the explicit call of `AddIsCongruentForMorphisms` should also be prevented.

mohamed-barakat: I agree that `IsEqualForObjects` must be there. Could you please make a suggestion?

sebastianpos: If the underlying category is not computable, then do the following 2 steps for the action category:
- set `IsEqualForObject` and `IsEqualForMorphisms` to `IsIdenticalObj`  
- set the caching to crisp.

mohamed-barakat:
I did not adjust the indentation of your code on purpose. If you like, you can do this in an extra commit or I can do it in your name.

sebastianpos:
It's fine, thanks.

PR #421, opened 2019-11-02
Allow several categories of rows/cols

### LINKED ISSUE #420: Allow several categories of Rows/Cols for one single ring
At the moment, the category of rows/cols is an attribute of the ring.
I'm planning to change that in order to allow multiple versions of the category of rows/cols.
In particular, there may coexist a "clever" and a "workhorse" version within the same session.

I will provide a PR in the near future.

sebastianpos:
Closes #420

### DIFF
--- a/FreydCategoriesForCAP/PackageInfo.g
+++ b/FreydCategoriesForCAP/PackageInfo.g
@@ -13,7 +13,7 @@ Subtitle := "Freyd categories - Formal (co)kernels for additive categories",
 Version := Maximum( [
   "2019.03.04", ## Martin's version
   ## this line prevents merge conflicts
-  "2019.10.09", ## Sepp's version
+  "2019.11.02", ## Sepp's version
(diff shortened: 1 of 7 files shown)


sebastianpos:
@mohamed-barakat 

please test if this change doesn't affect your packages building on rows/cols too much.

mohamed-barakat:
@sebastianpos: I have to adapt `ZariskiFrames` to use the new two-argument constructor `AsCategoryOfRowsMorphism` instead of the deprecated one-argument version. I won't come to it this week.

sebastianpos:
If you are nevertheless okay with these changes, I would like to merge this soon.

mohamed-barakat:
Merging it will break `ZariskiFrames` but this is OK since I will work on `ZariskiFrames` by the end of the week.

sebastianpos:
Ok.

PR #422, opened 2019-11-03
Convenience method for IsRat multiplication

sebastianpos:
This PR can only work properly if every homalg ring provides a function 
`interpret_rationals_func`, which takes an `IsRat` as an input and outputs
an interpretation of that `IsRat` as a ring element or `fail` else.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gd
+++ b/CAP/gap/CategoryMorphisms.gd
@@ -952,11 +952,24 @@ DeclareOperation( "AddAdditiveInverseForMorphisms",
 DeclareOperation( "MultiplyWithElementOfCommutativeRingForMorphisms",
                   [ IsRingElement, IsCapCategoryMorphism ] );
 
+#! @Description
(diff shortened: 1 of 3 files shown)


CAP/gap/CategoryMorphisms.gd
```diff
@@ -952,12 +952,25 @@ DeclareOperation( "AddAdditiveInverseForMorphisms",
 DeclareOperation( "MultiplyWithElementOfCommutativeRingForMorphisms",
                   [ IsRingElement, IsCapCategoryMorphism ] );
 
+#! @Description
+#! This is a convenience method. It has two arguments.
+#! The first argument is either a rational number $q$
+#! or an element $r$ of a commutative ring $R$.
+#! The second argument is a morphism $\alpha: a \rightarrow b$ in a linear category
+#! of the commutative ring $R$.
```
mohamed-barakat: of -> over

PR #423, opened 2019-11-04
Make Freyd tests faster

sebastianpos:


zickgraf:
In my tests for #424 I just noticed that disabling the caching hurts the performance quite a lot on my notebook. Approximate numbers of the runtime of `SnakeLemmaProof.g`:

* with `DeactivateDefaultCaching` and `DisableInputSanityChecks`: 46s
* without `DeactivateDefaultCaching` and with  `DisableInputSanityChecks`: 28s
* without both (already including #424): 54s

PR #424, opened 2019-11-04
Improve performance of generic sanity checks

zickgraf:
Approximate runtimes for `make test` in `FreydCategories` on my notebook:

* before the generic sanity checks where introduced: 2:38 min
* after the generic sanity checks where introduced: 4:32 min
* with this commit: 3:11 min

So with this commit, the new sanity checks increase the runtime by about 20% which seems acceptable for me. Nevertheless, I will try to profile this a bit more once I find the time.


sebastianpos:
The timings of Freyd have of course changed with #423.

zickgraf:
When running my tests I did not include #423.

PR #428, opened 2019-11-04
Fix getters in sanity checks

zickgraf:
The short-hand notation `x -> ...` creates unary functions, so all calls of `human_readable_identifier_getter` throw an error currently. Sorry for not testing this properly :/


sebastianpos:
No problem.

Thank you again.

PR #432, opened 2019-11-06
Fixes and performance improvements of sanity checks

zickgraf:
This creates input and output sanity check functions in advance to save time when the actual function call happens. This decreases the time the Freyd tests take on my notebook from 3:33 min to 3:21 min (with #423 reverted).


sebastianpos:
Is this PR ready for review?

zickgraf:
It is. Thinking about it, I realize it's probably confusing if one pushes commits to existing PRs. Next time I will open a separate PR to make things clearer :-)

sebastianpos:
Thanks.

PR #433, opened 2019-11-07
Add more derivation descriptions

zickgraf:
Note: I deliberately use two different styles ("A using B" vs. "Use B for A") because these different styles are used in other descriptions in the corresponding sections.


sebastianpos:
Thanks.

PR #441, opened 2019-11-24
Set FreydCategory to be abelian with enough projectives

kamalsaleh:
whenever the underlying_category is category of rows or cols of some
commutative ring.

### DIFF
--- a/FreydCategoriesForCAP/gap/FreydCategory.gi
+++ b/FreydCategoriesForCAP/gap/FreydCategory.gi
@@ -45,6 +45,8 @@ InstallGlobalFunction( FREYD_CATEGORY,
         
         SetIsAbelianCategory( freyd_category, true );
         
+        SetIsAbelianCategoryWithEnoughProjectives( freyd_category, true );
+    
     fi;
     
     conditions := [ "TensorProductOnObjects",
--- a/LinearAlgebraForCAP/gap/LinearAlgebraForCAP.gi
+++ b/LinearAlgebraForCAP/gap/LinearAlgebraForCAP.gi
@@ -30,6 +30,10 @@ InstallMethod( MatrixCategory,
     
     SetIsAbelianCategory( category, true );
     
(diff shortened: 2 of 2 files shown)


kamalsaleh:
For me, IsAbelianCategoryWithEnoughProjectives means there are methods to compute "IsProjective", "EpimorphismFromSomeProjectiveObject" and "ProjectiveLift", but these are currently not computable  in any abelian Freyd category, for example in the Freyd category of **graded** rows over Q[x,y].
```
LoadPackage( "FreydCategories" );
S := GradedRing( HomalgFieldOfRationalsInSingular( )*"x,y" );
rows := CategoryOfGradedRows( S );
frey := FreydCategory( rows );
IsAbelianCategory(frey);
#true
CanCompute( frey, "IsProjective" );
#false
CanCompute( frey, "EpimorphismFromSomeProjectiveObject" );
#true
CanCompute( frey, "ProjectiveLift" );
#false
```

kamalsaleh:
It would be nice to have the ability to derive properties, I believe this can be done, by AddToDoList, where is the entry is the category and the ToDo list will be triggered when HasIsFinalized has been set to true, and the ToDo list executes a function that asks about computability of certain methods, and can set accordingly properties to the category.

sebastianpos:
Up to now, properties like `IsAbelianCategoryWithEnoughProjectives` or `IsAbelianCategory` are understood as mere propositions in CAP, that means, their value does not imply the existence of any algorithm, but only of the mathematical validity of the statement.

In your case, I would suggest to simply use the `CanCompute` command. This is the safest solution.

kamalsaleh:
I am used to think that an category with property IsAbelianCategory should be able to compute all operations related to abelian categories, exactly as IsAdditiveCategory or IsTriangulatedCategory.

But your point is also important , since mathematical property can trigger more logic and theorems.
I will adjust the the PR are requested and locally in my packages I will install an operation "IsAbelianCategoryWithComputableEnoughProjectives".

PR #449, opened 2019-12-03
Add identifiers to cells which record how the cell was created

zickgraf:
When debugging I often have the problem that at some point in the computation, e.g. in a break-loop somewhere in the code, I get a "wrong" object/morphism but do not know how this object/morphism was created (it might have been created by going through lots of derivations for example). This PR tries to add an identifier to category cells, which records how the cell was created. Example:
```
LoadPackage( "ModulePresentationsForCAP" );
QQ := HomalgFieldOfRationals();
o := ZeroObject( QQ );
o!.identifier := "0";
id := IdentityMorphism( o );
iota := KernelEmbedding( id );
```
Then `iota!.identifier` is `KernelEmbedding( IdentityMorphism( 0 ) )`.

The current state is just the most simple demo and does not even handle lists etc.

If you think this might be a useful addition to CAP, my plan would be to provide a PR which handles the most basic things (like lists etc.), and then expand on this in the future when I'm debugging and need more features.

### DIFF
--- a/CAP/gap/CAP.gd
+++ b/CAP/gap/CAP.gd
@@ -29,6 +29,8 @@ DeclareGlobalVariable( "CAP_INTERNAL" );
 
 DeclareGlobalFunction( "CAP_INTERNAL_NAME_COUNTER" );
 
+DeclareGlobalFunction( "CAP_INTERNAL_IDENTIFIER_COUNTER" );
+
 DeclareGlobalFunction( "CATEGORIES_CACHE_GETTER" );
 
 DeclareGlobalFunction( "GET_METHOD_CACHE" );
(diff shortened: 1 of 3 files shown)


sebastianpos:
The idea of having such expressions as a debuging/analyzing tool is good, but I think building up the strings is way to expensive in order to have this feature turned on by default. So, if you ensure that it is a debugging feature that has to be turned on explicitly by the user, then we are good :-)

zickgraf:
I will probably make this lazy-evaluated as I did with the sanity checks. And this time I will run performance tests in advance :D

sebastianpos:
Please make it opt-in only.

PR #451, opened 2019-12-04
Ensure that (co)action category of a non-computable category is non-computable

zickgraf:
Without this the derivation mechanism adds `IsCongruentForMorphisms`.


sebastianpos:
Thanks.

PR #454, opened 2019-12-12
Add warning if WithGiven operation is installed, but no object operation

zickgraf:
Also check for the "without given" operation which could be used to derive
the object operation.

Example:
```
gap> LoadPackage( "CAP" );
gap> cat := CreateCapCategory( );
gap> AddKernelEmbeddingWithGivenKernelObject( cat, {a,b} -> a );
gap> Finalize( cat );
WARNING: You have installed a method for KernelEmbeddingWithGivenKernelObject but neither for KernelObject nor for KernelEmbedding. If this is intended, yo\
u can suppress this warning by passing the option "SuppressWithGivenWarning := true" to Finalize/IsFinalized.
```

### DIFF
--- a/CAP/gap/Finalize.gi
+++ b/CAP/gap/Finalize.gi
@@ -128,7 +128,33 @@ InstallMethod( IsFinalized,
                [ IsCapCategory ],
                
   function( category )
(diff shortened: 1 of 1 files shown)


zickgraf:
I will have to investigate the CI errors, so this can be ignored for now. Sorry for the spam resulting from this :-(

zickgraf:
This is superseded by #492.

PR #456, opened 2019-12-17
fix a corner case in the method of extending a functor to additive closure of source

kamalsaleh:
Sorry, my mistake.


sebastianpos:
No problem, thanks.

PR #457, opened 2020-01-07
Adding two new categorical methods BasisOfExternalHom and CoefficientsOfMorphismWithGivenBasisOfExternalHom for k-linear categories.

kamalsaleh:
Hallo @sebastianpos  :)

Main use of them:
1- We can now work with external hom
2- We can use them to derive homomorphism strucutre in certain cases, for example if k is a field.

This PR contains:
* Declaration & documentation of the above methods.
* Adding these operations to the category of matrices.
* Adding final derivation to derived homomorphism-structure in certain cases.
* Adding final derivation to derive the above methods from homomorphism-structure in certain cases.

I created this as a draft in case it needs some adjustments.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gd
+++ b/CAP/gap/CategoryMorphisms.gd
@@ -2003,10 +2003,11 @@ DeclareOperation( "HomStructure",
 
 
 #! @Description
-#! The arguments are object  $a,b$ in a pre-additive or $k$-linear category $C$.
-#! The output is a list $L$ of morphisms which is a basis of $\mathrm{Hom}_{C}(a,b)$ in the sense
-#! that any given morphism $\alpha: a \to b$ is the linear combination of $L$ with the
-#! coefficients in <C>CoefficientsOfMorphismWRTBasisOfExternalHom</C>($\alpha$).
+#! The arguments are objects $a,b$ in a $k$-linear category $C$.
+#! The output is a list $L$ of morphisms which is a basis of $\mathrm{Hom}_{C}(a,b)$ in
+#! the sense that any given morphism $\alpha: a \to b$ can uniquely be written as a
+#! linear combination of $L$ with the coefficients in
+#! <C>CoefficientsOfMorphismWithGivenBasisOfExternalHom</C>($\alpha,L$).
 #! @Returns a list of morphisms in $\mathrm{Hom}_{C}(a,b)$
 #! @Arguments a, b
 DeclareOperation( "BasisOfExternalHom",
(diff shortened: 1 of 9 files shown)


sebastianpos:
Hey Kamal :-)
one quick question: do they changes in `MonoidalCategories` have anything to do with this PR?
If no, could you please remove them from this PR?
Thanks a lot in advance!

CAP/gap/CategoryMorphisms.gd
```diff
@@ -2002,6 +2002,73 @@ DeclareOperation( "HomStructure",
                   [ IsCapCategoryObject, IsCapCategoryObject ] );
 
 
+#! @Description
+#! The arguments are object $a,b$ in a $k$-linear category $C$.
```
sebastianpos:
```suggestion
#! The arguments are objects $a,b$ in a $k$-linear category $C$.
```

CAP/gap/CategoryMorphisms.gd
```diff
@@ -2002,6 +2002,73 @@ DeclareOperation( "HomStructure",
...
+                  [ IsCapCategory, IsFunction, IsInt ] );
+
+DeclareOperation( "AddBasisOfExternalHom",
+                  [ IsCapCategory, IsList, IsInt ] );
+
+DeclareOperation( "AddBasisOfExternalHom",
+                  [ IsCapCategory, IsList ] );
+
+#! @Description
+#! The arguments are a morhpism  $\alpha: a \to b$ in a $k$-linear category $C$ and
```
sebastianpos:
```suggestion
#! The arguments are a morphism  $\alpha: a \to b$ in a $k$-linear category $C$ and
```

CAP/gap/CategoryMorphisms.gd
```diff
@@ -2002,6 +2002,73 @@ DeclareOperation( "HomStructure",
...
+                  [ IsCapCategory, IsFunction, IsInt ] );
+
+DeclareOperation( "AddCoefficientsOfMorphismWithGivenBasisOfExternalHom",
+                  [ IsCapCategory, IsList, IsInt ] );
+
+DeclareOperation( "AddCoefficientsOfMorphismWithGivenBasisOfExternalHom",
+                  [ IsCapCategory, IsList ] );
+
+#! @Description
+#! The is a convenience method.
```
sebastianpos:
```suggestion
#! This is a convenience method.
```

CAP/gap/CategoryMorphisms.gd
```diff
@@ -2002,6 +2002,73 @@ DeclareOperation( "HomStructure",
...
+#! The argument is a morhpism  $\alpha: a \to b$ in a $k$-linear category $C$.
```
sebastianpos:
```suggestion
#! The argument is a morphism  $\alpha: a \to b$ in a $k$-linear category $C$.
```

CAP/gap/DerivedMethods.gi
```diff
@@ -3329,3 +3329,159 @@ AddFinalDerivation( IsEqualForCacheForMorphisms,
...
+                      CoefficientsOfMorphismWithGivenBasisOfExternalHom
+                    ],
+  function( a, b )
+    local hom_a_b, D, B, ConditionsListComplete;
+    
+    hom_a_b := HomomorphismStructureOnObjects( a, b );
+    
+    D := DistinguishedObjectOfHomomorphismStructure( CapCategory( a ) );
+    
+    B := ValueGlobal( "BasisOfExternalHom" )( D, hom_a_b );
```
sebastianpos: Why did you use this "trick" instead of calling
`BasisOfExternalHom( D, hom_a_b )`
?

CAP/gap/DerivedMethods.gi
```diff
@@ -3329,3 +3329,159 @@ AddFinalDerivation( IsEqualForCacheForMorphisms,
...
+        
+      fi;
+      
+      return true;
+      
+  end,
+  Description := "Adding BasisOfExternalHom using homomorphism structure" );
+
+##
+AddFinalDerivation( CoefficientsOfMorphismWithGivenBasisOfExternalHom,
```
sebastianpos: I think it's better to bundle this final derivation and the one for `BasisOfExternalHom` as a "final derivation package", like it was done for the weak kernel here:
https://github.com/homalg-project/CAP_project/blob/7a2391216c737a2e9cc004d0dc1ca8f7bf541682/FreydCategoriesForCAP/gap/FreydCategoriesDerivedMethods.gi#L282

If you don't bundle it, I fear that it is not possible that both final derivations can be triggered at the same time.

kamalsaleh:
I have overlooked that commit. I will remove it from this PR and cherry-pick it in another branch until it is needed :).

sebastianpos:
You may ignore my first concern, I think I now see what this file is good for: you derive hom structures for other categories, not for matrix categories.

LinearAlgebraForCAP/gap/DerivedMethodsForMatrixCategories.gi
```diff
@@ -0,0 +1,232 @@
...
+      
+      return false;
+      
+    fi;
+    
+    matrix_cat := MatrixCategory( k );
+    
+    if not HasRangeCategoryOfHomomorphismStructure( cat ) then
+      
+      SetRangeCategoryOfHomomorphismStructure( cat, matrix_cat );
```
sebastianpos: This line is extremely dangerous.
You set an attribute within a function that is only supposed to check whether a certain derivation might be applicable. You cannot guarantee that even if the check yields true that the final derivation will be actually installed, and you would have set this attribute wrongly in that case.

I can think of two possible solutions:
1) we need to make a CAP primitive for `RangeCategoryOfHomomorphismStructure` and then you can derive this one as well
2) you bundle all the final derivations in this file to a "final derivation package" (like I suggested with the other final derivations above), and I provide a mechanism that calls a user given function (in which you could then set this attribute) in the case such a package is actually installed.

Can you think of other solutions? If no, maybe it's also not a bad idea to actually employ 1) and 2).

CAP/gap/DerivedMethods.gi
kamalsaleh: Sanity checks for the final derivations thinks this "BasisOfExternalHom" is being used in the category for which we derive the method. Hence, insists that it should be in the list of required methods.

LinearAlgebraForCAP/gap/DerivedMethodsForMatrixCategories.gi
```diff
@@ -0,0 +1,232 @@
...
+    
+    k := CommutativeRingOfLinearCategory( cat );
+    
+    dim_hom_source := Dimension( hom_source );
+    
+    dim_hom_range := Dimension( hom_range );
+    
+    if dim_hom_source * dim_hom_range = 0 then
+      
+      return ValueGlobal( "ZeroMorphism" )( hom_source, hom_range ); # sanity checks for final derivations
```
kamalsaleh: Same happens here, there should be a way to let sanity checks (for final derivation) to omit some operations, if they happen in other categories.

kamalsaleh: @sebastianpos

kamalsaleh:
One easy way to test this PR is pull the package "FunctorCategories", branch "forkamal":
https://github.com/mohamed-barakat/FunctorCategories
There I add the two methods and get the homomorphism structure.

CAP/gap/DerivedMethods.gi
kamalsaleh: Very nice. I have updated the derivation.

LinearAlgebraForCAP/gap/DerivedMethodsForMatrixCategories.gi
kamalsaleh: To be done tomorrow.

kamalsaleh: The idea of adding an optional function  to a "final derivation package" which is designed to be called before actually starting to add the components of the package, would be very helpful for future instances. So it would be very nice to have this mechanism.

The idea of adding a primitive operation which sets/computes the attribute "RangeCategoryOfHomomorphismStructure" is also feasible, exactly as if we are adding "ZeroObject" method; which is also an attribute to the category.

kamalsaleh: In this commit https://github.com/homalg-project/CAP_project/pull/457/commits/c575c87c6b9efefab338dc7b3a6064e1a55976e5, I bundled the several final derivations to a single finale derivation package, and as soon as you implement the above or other mechanism, I will accordingly re-update the `CATEGORY_FILTER_CHECKER`.

kamalsaleh:
In order to reduce the number of commits, I have re-based and force pushed.

sebastianpos:
Looks good, thanks.

Could you also open an issue concerning the sanity check of final derivations? Thanks!

PR #458, opened 2020-01-10
Call func before installation of a derived method

sebastianpos:
@kamalsaleh 
This is the mechanism that should resolve the problem addressed in #457 concerning `SetRangeCategoryOfHomomorphismStructure`.
Please try if it works in your context. If yes, I can merge this PR.

To provide a function that is called before the installation of a derived method, simply put
`FunctionCalledBeforeInstallation := my_func( cat )`
in the options of `AddDerivationToCAP`.
Note that the function is assumed to be unary and expects a category as an input.

### DIFF
--- a/CAP/gap/Derivations.gd
+++ b/CAP/gap/Derivations.gd
@@ -156,6 +156,12 @@ DeclareOperation( "InstallDerivationForCategory",
 DeclareOperation( "DerivationResultWeight",
                   [ IsDerivedMethod, IsDenseList ] );
 
+#! @Description
+#! Input is a derived method. Output is a unary function that takes as an input
+#! a category and does not output anything. This function is always called before
+#! the installation of the derived method for a concrete instance of a category.
+#! @Arguments d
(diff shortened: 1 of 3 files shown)


kamalsaleh:
I thought this PR would touch final derivation mechanism rather than standard `AddDerivationToCAP`.
This changes can not be triggered from final derivation package.

sebastianpos:
You are right, the final derivations have an extra place for their installation.
I'll provide an additional commit after the weekend.

kamalsaleh:
Ok, then have a nice weekend :)

sebastianpos:
@kamalsaleh 
Okay, I provided the same mechanism for final derivations. Please try again :-)

kamalsaleh:
Thank you Sebastian, works perfectly. I will re-base, merge-commits and force-push to https://github.com/homalg-project/CAP_project/pull/457

sebastianpos:
Okay, I'll merge then.

PR #459, opened 2020-01-13
Include documentation for family properties

zickgraf:
This needs a recent AutoDoc version. The relevant AutoDoc-PR is https://github.com/gap-packages/AutoDoc/pull/234 (not yet merged).

I assume that the `TODO` comments were referring to the fact that the comments were not picked up by AutoDoc.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gd
+++ b/CAP/gap/CategoryMorphisms.gd
@@ -71,15 +71,14 @@ DeclareGlobalVariable( "PROPAGATION_LIST_FOR_EQUAL_MORPHISMS" );
 ##
 ###################################
 
-## TODO
-# @Description
-# The argument is a morphism $\alpha$.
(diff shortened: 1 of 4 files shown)


sebastianpos:
I think it would be wiser to simply get rid of the `DeclareFamilyProperty` concept in CAP. As far as I can see, it has no real implications anymore.

zickgraf:
Ah, I did not actually look at what `DeclareFamilyProperty` does. Should it suffice to replace `DeclareFamilyProperty` by `DeclareProperty` while dropping the additional arguments? If yes, I would update this PR accordingly.

sebastianpos:
I think so, yes.

Later, I can take care of deleting all the code implementing "family properties".

zickgraf:
I have replaced all occurrences of `DeclareFamilyProperty` by `DeclareProperty` and the CI is green, so this is ready for review :-)

sebastianpos:
Thank you :-)

PR #472, opened 2020-01-24
Create example for the computation of subobject lattices

sebastianpos:
and provide overhead options for QuiverRows, FreydCategories, and AdelmanCategory


zickgraf:
I have a question: as far as I know, options get passed down the whole function stack, see for example the following code, which correctly displays `test`:
```
func1 := function()
    Display( ValueOption( "myoption" ) );
end;

func2 := function()
    func1();
end;

func2( : myoption := "test" );
``` 
Why do you explicitly "unpack" the overhead option?

sebastianpos:
You are right, it is unnecessary to give `overhead` as an argument to the functions
`INSTALL_FUNCTIONS_FOR_ADELMAN_CATEGORY`
and
`INSTALL_FUNCTIONS_FOR_QUIVER_ROWS`.
I will change that. Thank you.

sebastianpos:
Okay, now I understand better. You mean I can omit any change in the category constructors at all in order to pass an option to `CreateCapCategory`. This is even better :-)

zickgraf:
Exactly, that was my assumption but since I did not test it I rather phrased it as a question :-)

PR #486, opened 2020-04-04
added missing newline in Display methods

mohamed-barakat:


mohamed-barakat:
Done.

I also decided to use `Display( string )` in the last lines instead of `Print( string, "\n" )`.

PR #487, opened 2020-04-05
added convenience operation/methods InternalHom

mohamed-barakat:
The new convenience operation `InternalHom` accepts two CAP cells
as arguments and invokes either `InternalHomOnObjects` if the arguments
are two objects and `InternalHomOnMorphisms` otherwise. If in the latter case
one of the arguments is an object, it is replaced by its identity morphism.

### DIFF
--- a/MonoidalCategories/gap/ClosedMonoidalCategories.gd
+++ b/MonoidalCategories/gap/ClosedMonoidalCategories.gd
@@ -797,6 +797,7 @@ DeclareOperation( "AddIsomorphismFromInternalHomToObjectWithGivenInternalHom",
 
 ##
 #! @Description
+#! This is a convenience method.
 #! The arguments are two cells $a,b$.
 #! The output is the internal hom cell .
 #! If $a,b$ are two CAP objects the output is the internal Hom object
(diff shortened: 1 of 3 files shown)


MonoidalCategories/gap/ClosedMonoidalCategories.gd
```diff
@@ -788,3 +788,22 @@ DeclareOperation( "AddIsomorphismFromInternalHomToObjectWithGivenInternalHom",
...
+
+####################################
+##
+#! @Section Convenience Methods
+##
+####################################
+
+##
+#! @Description
+#! The arguments are two cells $a,b$.
```
sebastianpos, line 800:
```suggestion
#! This is a convenience method.
#! The arguments are two cells $a,b$.
```

MonoidalCategories/gap/ClosedMonoidalCategories.gd
```diff
@@ -788,3 +788,22 @@ DeclareOperation( "AddIsomorphismFromInternalHomToObjectWithGivenInternalHom",
...
+#! The output is the internal hom cell .
```
sebastianpos, line 801:
```suggestion
#! The output is the internal hom cell.
```

PR #489, opened 2020-04-08
Bug fix by adjusting the morphism constructor in product categories and add convenience methods

kamalsaleh:
```
LoadPackage( "LinearAlgebraForCAP" );
LoadPackage( "RingsForHomalg" );
field := HomalgFieldOfRationals( );
mat_1 := MatrixCategory( field );
mat_2 := MatrixCategory( field );
SetCachingOfCategory( mat_1, "none" );
SetCachingOfCategory( mat_2, "none" );
IsIdenticalObj( Product( mat_1, mat_2 ), Product( mat_1, mat_2 ) );
#! false
```
This leads to the following error:

```
P := Product( mat_1, mat_2 );
#! Product of: Category of matrices over Q, Category of matrices over Q
ProductOp_OnMorphisms( [ ZeroObjectFunctorial(mat_1), ZeroObjectFunctorial(mat_2) ], P );
    Error, an object that lies in the CAP-category with the name
    Product of: Category of matrices over Q, Category of matrices over Q
   was tried to be added to a different CAP-category with the name
   Product of: Category of matrices over Q, Category of matrices over Q
```
This PR computes the source and range so that they lies in P and then sets them.

### DIFF
--- a/CAP/gap/ProductCategory.gd
+++ b/CAP/gap/ProductCategory.gd
@@ -61,6 +61,9 @@ DeclareOperation( "ProductOp_OnObjects",
 DeclareOperation( "ProductOp_OnTwoCells",
                   [ IsList, IsCapCategory ] );
 
+DeclareOperation( "\/",
+                  [ IsList, IsCapProductCategory ] );
+
 DeclareOperation( "ProductOp",
                   [ IsList, IsCapCategoryCell ] );
 
--- a/CAP/gap/ProductCategory.gi
+++ b/CAP/gap/ProductCategory.gi
@@ -286,6 +286,31 @@ InstallMethod( ProductOp,
 end );
 
 ##
+InstallMethod( \/,
+          [ IsList, IsCapProductCategory ],
+  function( list, category )
+
+    if IsCapCategoryObject( list[ 1 ] ) then
+
+      return ProductOp_OnObjects( list, category );
+
+    elif IsCapCategoryMorphism( list[ 1 ] ) then
+
+      return ProductOp_OnMorphisms( list, category );
+
+    elif IsCapCategoryTwoCell( list[ 1 ] ) then
+
+      return ProductOp_OnTwoCells( list, category );
+
+    else
+
+      Error( "Wrong input!\n" );
+
+    fi;
+
+end );
+
+##
 InstallMethod( ProductOp,
                [ IsList, IsCapCategoryMorphism ],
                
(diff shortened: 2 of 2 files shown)


sebastianpos:
I actually cannot reproduce the error.

I looked at the code of the Product category and fear that we have a logic error:
1) the constructors of objects and morphisms for Product categories (i.e., calling `Product( a, b )` ) assume that there is only a single Product category for each pair of categories, since one does not pass the product category to this constructor.
2) the Product category constructor itself is build on the cache of the category of its first argument. If this is not crisp, there might appear different instances of the same product categories.

Could you please try the following:
Replace in the [constructor](https://github.com/sebastianpos/CAP_project/blob/1df85be6520b639297cc8d415f80e0e425e0c39b/CAP/gap/ProductCategory.gi#L178) of the product category the
`InstallMethodWithCacheFromObject`
by
`InstallMethodWithCache`
and set it to crisp (by pushing an option) and try again if the error occurs for you?

I think with the fix that you proposed, the main issue remains: calling `Product( a, b )` several times on the same input might lead to non-identical categories.

kamalsaleh:
The error can be reproduced by adding the following two lines after creating the two categories.
```
SetCachingOfCategory( mat_1, "none" );
SetCachingOfCategory( mat_2, "none" );
```
I have corrected the code. I will try now your suggestion.

kamalsaleh:
Replacing `InstallMethodWithCacheFromObject` in the constructor by `InstallMethodWithCrispCache` fixes the issue.

kamalsaleh:
I can hence, modify this PR to include only two commits: the first installes the method with crisp cache and the second commit includes the convenient method. Should I change and I force push?

sebastianpos:
Yes, that would be great. Thank you for the testing and the hint to reproduce the error.

kamalsaleh:
The change causes a lot of errors, such as
`Error, no 1st choice method found for "CachingObject" on 3 arguments`
 in tests.

kamalsaleh:
I think it is still good idea, to compute the source and range and set them in the morphism constructor, rather than computing them later. Such a hot fix would  solve the problem for me for now and close this PR. Then I can create an issue to take care about the caching attributes in the installation method.

sebastianpos:
With the latest change everything should work fine.

I'll merge this PR if you agree.

kamalsaleh:
Of course :)

PR #492, opened 2020-04-15
Use derivation mechanism for with/without given pairs

### LINKED ISSUE #455: Unexpected behaviour when installing WithGiven operations
When installing a WithGiven operation, the corresponding without given operation is also installed automatically: https://github.com/homalg-project/CAP_project/blob/master/CAP/gap/InstallAdds.gi#L513

and added as a primitive operation: https://github.com/homalg-project/CAP_project/blob/master/CAP/gap/InstallAdds.gi#L507

Thus, if one only installs for example `KernelEmbeddingWithGivenKernelObject`, afterwards `CanCompute( cat, "KernelObject" )` returns true, but of course calling `KernelObject` leads to an infinite recursion via `KernelObject( alpha ) = Source( KernelEmbedding( alpha ) ) = KernelEmbeddingWithGivenKernelObject( alpha, KernelObject( alpha ) )`.

I tried to add a warning for this in #454, but this failed because even if the object function is not primitively installed it might be derived in a way not leading to an infinite recursion.

I guess the proper solution for this would be to remove the code in https://github.com/homalg-project/CAP_project/blob/master/CAP/gap/InstallAdds.gi#L513 and instead use the derivation mechanism to derive `KernelEmbedding` from `KernelEmbeddingWithGivenKernelObject` and `KernelObject`. But I am not sure about other implications the removal might have.

Do you have any ideas/suggestions for improving the behaviour or adding a warning?

Note: the without given operation is not added to the list of primitively installed functions: https://github.com/homalg-project/CAP_project/blob/master/CAP/gap/InstallAdds.gi#L509 So at least this can be tested.

zickgraf:
Fixes #455 by considering `F` and `FWithGivenX` as two different basic operations, where `F` can only be derived from `FWithGivenX` if `X` is also given or derived from something else.

Supersedes #454.

This is WIP since I want to improve the logic detecting the pairs and the redirect handling (at least I said so in my comment in #455 :D )


zickgraf:
This is now much larger than expected. I will explain the reasoning behind each commit. If you like, I can split this into several PRs.

1. 47ef2a73639223a9be4da06eb41afbc4076693b6 does what we have discussed before: get rid of the concept of pair functions and use the derivation mechanism instead. Note: the logic when (not) to use a redirect function was completely broken as far as I can tell: if one installs `F` with weight 100 and afterwards `FWithX` with weight 50, `F` does not redirect to `FWithX` although it should if I am not mistaken. Thus, I have rewritten the logic.

2. 512d152456ef36100cbc9b17dd058c6557b4137e: Since I had to detect `WithGiven` methods, I have taken a look at the detection mechanism. Parts of it happened in `InstallAdds.gi` and parts of it in `MethodRecord.gi`. For consistency I moved the detection mechanism and additionally all things which I considered as "enhancing the method record" to `MethodRecord.gi`. This is mostly copy and paste plus some minor adjustments.

3. 2ce7a6288586db2320d0a0fdcdbedfb758cfeb26 removes another instance where WithGiven methods were detected manually.

4. a100593a865724795a16dcdaf0fc079a44be742f: I noticed that the existing code assumes that every diagram is given by one argument. I wanted to make this more flexible for future use and used `number_of_diagram_arguments` for that.

5. ae9e48a668f472fe39f198daa579cd690da39fe7: I noticed two things:
	* `number_of_diagram_arguments` defaulted to 1. I think this is not a good idea: if diagrams given by two arguments are introduced in the future and one does not know about `number_of_diagram_arguments`, this leads to obscure errors.
	* `number_of_diagram_arguments` was not set to 0 for `ZeroObject` et al., which was unexpected for me.

6. e176326a451b364ae72fd8fdc7557db9c5824984 is a simple typo.

This is now ready for review. I think its easier to review each commit separately than the PR as a whole. As mentioned above, I can also split the PR if you like.

CAP/gap/DerivedMethods.gi
```diff
@@ -3471,3 +3471,53 @@ AddFinalDerivation( BasisOfExternalHom,
...
 
+BindGlobal( "CAP_INTERNAL_INSTALL_WITH_GIVEN_DERIVATION_PAIR", function( without_given_name, with_given_name, object_name, object_arguments )
+    
+    AddDerivationToCAP( ValueGlobal( with_given_name ),
+                        [ [ ValueGlobal( without_given_name ), 1 ] ],
+      function( arg )
+        
+        return CallFuncList( ValueGlobal( without_given_name ), arg{[ 1 .. Length( arg ) - 1 ]} );
+        
+    end : Description := Concatenation( with_given_name, "by calling ", without_given_name, " with the last argument dropped" ) );
```
sebastianpos, line 3482:
```suggestion
    end : Description := Concatenation( with_given_name, " by calling ", without_given_name, " with the last argument dropped" ) );
```

CAP/gap/DerivedMethods.gi
```diff
@@ -3471,3 +3471,53 @@ AddFinalDerivation( BasisOfExternalHom,
...
+    
+    AddDerivationToCAP( ValueGlobal( without_given_name ),
+                        [ [ ValueGlobal( with_given_name ), 1 ],
+                          [ ValueGlobal( object_name ), 1 ] ],
+      function( arg )
+        
+        return CallFuncList( ValueGlobal( with_given_name ),
+                                    Concatenation( arg, [ CallFuncList( ValueGlobal( object_name ), arg{object_arguments} ) ] ) );
+        
+    end : Description := Concatenation( without_given_name, "by calling ", with_given_name, " with ", object_name, " as last argument" ) );
```
sebastianpos, line 3492:
```suggestion
    end : Description := Concatenation( without_given_name, " by calling ", with_given_name, " with ", object_name, " as last argument" ) );
```

zickgraf:
I have now checked "Allow edits from maintainers.", feel free to push :-)

zickgraf:
Ah, this also needs a rebase on master because of the change in `RingsForHomalg`. Let me know if I should do the rebase, I don't want to interfere with your push!

sebastianpos:
Yes, please do the rebase.

zickgraf:
Done. The CI got triggered but somehow does not show up yet.

sebastianpos:
Perfect. Thanks a lot for your effort!

PR #493, opened 2020-04-17
Adapt to https://github.com/homalg-project/homalg_project/pull/318

zickgraf:


zickgraf:
I have retriggered the CI and it is green :-)

@sebastianpos It would be nice if you could merge this soon to make the homalg CI pass again. I think the review should be easy: in the source of `iota1` the second and third relation are swapped, so in the source of `iota2` (encoding the relations between the relations) the second and third entry must be swapped. All of these relations are computed in three different ways, which explains all six changes.

zickgraf:
Thanks, now all CIs pass again :-)

PR #502, opened 2020-04-21
added missing newline in Display method for objects in AdditiveClosure

mohamed-barakat:


sebastianpos:
Could you briefly explain in what instances the current version did not work properly?

mohamed-barakat:
When "wrapping" a category C, I sometimes like to use the ViewObj/Display methods of C in the ViewObj/Display methods for the wrapping category

```gap
InstallMethod( Display,
  [ IsWrappedCapCategoryObject ],

function( obj )
  Display( UnderlyingCell( obj ) );
  Display( "An object in ", Name( CapCategory ), " given by the above data" );
end );
```
Indeed `Display( UnderlyingCell( obj ) );` produce a trailing newline for almost all of the CAP categories, but I found some few CAP categories with Display methods for objects/morphisms with no trailing newline. I can of course use
```gap
Display( "\nAn object in ", Name( CapCategory ), " given by the above data" );
```
(which I, in fact, do because I prefer an extra empty line as a visual delimiter), but again the behavior is not unified, hence the fix. I also think that a Display method in GAP should always have a trailing newline. You can see this in the difference between
```gap
gap> Print( "a" ); Print( "b" );
ab
gap> Display( "a" ); Display( "b" );
a
b
```

sebastianpos:
Ok, thanks for the explanation.

PR #504, opened 2020-04-24
Relations and Linear closure

sebastianpos:
- Implementation of a raw version of the category of relations  that is also applicable in a non-abelian context.
- Enhancing the R-linear closure in two ways:
  1) allow input which does not come with a sorting function (which, of course, should only be used if no sorting function is available, since sorting functions give a speedup)
  2) allow for a cocycle with values in R that twists the multiplication of morphisms in the category.

### DIFF
--- a/FreydCategoriesForCAP/examples/Relations.g
+++ b/FreydCategoriesForCAP/examples/Relations.g
@@ -1,0 +1,26 @@
+#! @Chapter Examples and Tests
+
+#! @Section Category of relations
+
+LoadPackage( "RingsForHomalg" );;
+LoadPackage( "LinearAlgebraForCAP" );;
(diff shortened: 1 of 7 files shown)


kamalsaleh:
This is abstraction of which categories :) ?

sebastianpos:
The category of relations is equivalent to the generalized morphism category in the case when the input is abelian. I want to be able to use it in the case when the input is only a regular category.

The twisted linear closure is an abstraction of group rings twisted by cocycles. Another example for the usage of twisted linear closures is given by Friedrich Knop: instead of taking the linear closure of the category of relations, one may define a composition that is twisted by a scalar determined by a very natural family of cocycles.

PR #509, opened 2020-04-27
 Install WithGiven derivations while installing adds 

zickgraf:
This way, WithGiven derivations are automatically installed for other packages, too. Before, only derivations for methods in `CAP_INTERNAL_METHOD_NAME_RECORD` were installed.


sebastianpos:
Currently, the tests for `FreydCategoriesForCAP` fail, which should be resolved by
https://github.com/homalg-project/Toposes/pull/8

PR #511, opened 2020-04-28
Use UnionOfRows/Columns for lists

zickgraf:
Applying UnionOfRows/Columns to lists instead of iteratively to pairs gives a small but noticeable performance improvement in case of large lists.

Some notes:
1. I wanted to replace all cases but noticed that lots of them are in examples. To save time I only replaced the cases occurring in actual code (with the exception of VectorSpacesAllMethods.gi because I started with this file).
2. Some occurrences are really weird: UnionOfRows/Columns is applied using `Iterated` to a list with two arguments. Do I miss something or was this unnecessary even before UnionOfRows/Columns was available for lists?
3. I really hope I did not make a mistake somewhere, fingers crossed.... :D

### DIFF
--- a/CAP/examples/testfiles/VectorSpacesAllMethods.gi
+++ b/CAP/examples/testfiles/VectorSpacesAllMethods.gi
@@ -218,13 +218,10 @@ AddInjectionOfCofactorOfDirectSum( vecspaces,
     
     coproduct := QVectorSpace( dim );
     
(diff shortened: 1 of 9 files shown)


zickgraf:
The CI does not show up but finished successfully: https://github.com/zickgraf/CAP_project/commits/UnionOfRows_lists

mohamed-barakat:
@zickgraf: If you update your commit it the ci will be triggered and the result will appear here again (e.g., you can amend and force push).

zickgraf:
@mohamed-barakat You are right, thanks!

PR #512, opened 2020-04-29
change view method for CanonicalProjection functor of Serre quotients

mohamed-barakat:
changed: Embedding in -> Localization functor of


sebastianpos:
Thanks, I don't know why the name was like that.

mohamed-barakat:
I can imagine that it was copied from the corresponding method for generalized morphisms categories and not adapted.

PR #513, opened 2020-04-29
do not call HomalgFieldOfRationalsInSingular multiple times

mohamed-barakat:
but get the ring from the given data in the method for DegreeZeroPartFunctor


sebastianpos:
@mohamed-barakat  I'm sorry, but there appeared a conflict. Could you resolve it please?

PR #514, opened 2020-04-29
use option (precision := "gmp") for all 4ti2 functions

mohamed-barakat:
needs 4ti2Interface v2020.04.01

avoids annoying warnings in 4ti2 version 1.6.9

WARNING: Overflow detection is not available on this architecture/compiler.
WARNING: To guarantee correct results, run 4ti2 with arbitrary precision
WARNING: by using the option `-parb'


mohamed-barakat:
The default is now "arb" for arbitrary precision in 4ti2 interface as recommended by the 4ti2 binaries.

PR #519, opened 2020-05-11
Remove manual pdf

sebastianpos:
In order to facilitate keeping the CAP manual up to date, I did the following:

- I forked the CAP_Manual repo to the homalg_project and added the CAPManual.pdf file there. The idea is that all future changes to the manual should be performed in this fork.
-  I deleted the CAPManual.pdf from the CAP webpage and replaced it with a link to the CAPManual.pdf provided by the forked CAP_Manual repository.

This PR removes the instance of the CAPManual.pdf within the CAP_project repo. Thus, the only remaining instance will be here:
https://github.com/homalg-project/CAP_Manual/blob/master/CAPManual.pdf


zickgraf:
Out of curiosity: Is there a reason for not including the source code of the manual in this repository? That would be the most natural thing for me.

sebastianpos:
Good idea, we should do that. However, we should do it in a way such that we do not lose the history of the CAP_Manual repository.

In any case, deleting the current pdf file at this place would be the first step. If you agree, I will merge this PR.

zickgraf:
> Good idea, we should do that. However, we should do it in a way such that we do not lose the history of the CAP_Manual repository.

Of course, git subtree to the rescue :-) I think
```
git subtree add --prefix=CAP_Manual /path/to/CAP_Manual master
```
should achieve this. (Maybe drop the prefix `CAP` because its already implied by being part of the `CAP_project`?)

> In any case, deleting the current pdf file at this place would be the first step. If you agree, I will merge this PR.

Maybe it makes sense to keep the pdf at the top level to make it easier to find. However, this would mean that one has to copy the pdf one level up after compilation in the subdirectory. 
=> I do not have a strong opinion on this.

sebastianpos:
Since copying the pdf file one level up for me appears to be error-prone, I would simply suggest to provide a link to the manual in the Readme.md file instead.

zickgraf:
Sound fine :-)

PR #522, opened 2020-05-11
Update manual w.r.t. groups example

### LINKED ISSUE #518: Example from the manual does not work
In the manual:

https://github.com/homalg-project/CAP_project/blob/master/CAPManual.pdf

The introductory example with the category of groups (p. 5, 6) fails with the following error message when constructing the identity of the group S3:

gap>  id_S3 := IdentityMorphism( S3);
Error, in function IdentityMorphism
       of category groups:
       the result does not lie in the filter IsCapCategoryMorphism.
 at /proc/cygdrive/C/gap-4.11.0/pkg/CAP-2019.06.07/gap/InstallAdds.gi:64 called from
CAP_INTERNAL_DISPLAY_ERROR_FOR_FUNCTION_OF_CATEGORY( record.function_name, category,
 "the result does not lie in the filter IsCapCategoryMorphism."
 ); at /proc/cygdrive/C/gap-4.11.0/pkg/CAP-2019.06.07/gap/InstallAdds.gi:323 called from
CallFuncList( func, arg ) at /proc/cygdrive/C/gap-4.11.0/pkg/ToolsForHomalg-2019.09.02/gap/CachingObjects.gi:775 called from
<function "IdentityMorphism method">( <arguments> )
 called from read-eval loop at *stdin*:16
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk>

sebastianpos:
plus some other fixes.

Closes #518


Tutorial.tex
```diff
@@ -22,24 +22,29 @@ \section{The category of groups}
 \input{GAP_tex/group_example/AddPreCompose_main}
 
 Now the composition of two morphisms will be computed by the \texttt{\textbackslash *} operation.
-Another thing every category needs is a function to compute the identity morphism. For this
-category of groups this can be done as follows.
+Another thing every category needs is a function to compute the identity morphism.
```
kamalsaleh: Here there is a typo :)

PR #524, opened 2020-05-11
d&i CapCategorySwitchLogicPropagationOn/Off

mohamed-barakat:
This allows disabling INSTALL_TODO_LIST_FOR_EQUAL_OBJECTS/MORPHISMS
without switching off the logic entirely.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2020.05.01", ## Mohamed's version
+  "2020.05.02", ## Mohamed's version
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2020.04.15", ## Mohamed's version
+  "2020.05.01", ## Mohamed's version
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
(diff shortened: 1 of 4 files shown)


zickgraf:
I would suggest the following addition:
Replacing
```
<C>CapCategorySwitchLogicOff</C>: see <Ref Sect="Section_Logic_switcher" />.
```

by

```
<C>CapCategorySwitchLogicOff</C> or <C>CapCategorySwitchLogicPropagationOff</C>: see <Ref Sect="Section_Logic_switcher" />.
```
in https://github.com/homalg-project/CAP_project/blob/cb1ce9ab1ea9f2e85c43ae8a179f532c176bdbe1/CAP/gap/CAP.gd#L435

mohamed-barakat:
> I would suggest the following addition:
> Replacing
> 
> ```
> <C>CapCategorySwitchLogicOff</C>: see <Ref Sect="Section_Logic_switcher" />.
> ```
> 
> by
> 
> ```
> <C>CapCategorySwitchLogicOff</C> or <C>CapCategorySwitchLogicPropagationOff</C>: see <Ref Sect="Section_Logic_switcher" />.
> ```
> 
> in
> 
> https://github.com/homalg-project/CAP_project/blob/cb1ce9ab1ea9f2e85c43ae8a179f532c176bdbe1/CAP/gap/CAP.gd#L435

Thanks, done.

mohamed-barakat:
> I like this PR.
> 
> I would like to discuss if we want to set the `predicate_logic_propagation` per default to `false`.
> In my opinion, this would be a good idea:
> 
> 1. The current default actually prevents the user who doesn't know about CAP's internals from working with higher quantities of objects, since the logic propagation between equal objects/mors quickly becomes very expensive.
> 2. Up till now, the concrete benefit of having logic propagation between equals was low in all the cases that I worked with.

Yes, I agree. We should turn the propagation off by default. It should be a deliberate decision to turn it on. Shall I alter the commit now or wait for further feedback?

sebastianpos:
I think you may now alter the commit.

mohamed-barakat:
I added one more commit which adds to `CapCategorySwitchLogicPropagationOn/Off` further methods

* `CapCategorySwitchLogicPropagationForObjectsOn/Off`
* `CapCategorySwitchLogicPropagationForMorphismsOn/Off`

PR #530, opened 2020-05-14
Implement Down-feature

sebastianpos:
I often encounter the situation that I have an object/morphism in some category and want to access its underlying defining data. Unfortunately, depending on the category, these data can have different names, e.g. `UnderlyingMatrix`, `UnderlyingOriginalObject`, `MorphismDatum`, etc.

This PR is an attempt to provide a convenience method for such situations.

I have only implemented it in some special cases since the capabilities of this new feature are of course open for discussion.


sebastianpos:
If there are no complaints then I'll merge this PR tomorrow :-)

PR #510, opened 2020-04-27
Add Coclosed methods to MonoidalCategories

TKuh:
I'm quite happy that the fixes in MonoidalCategories were finally merged, so here is the code for the coclosed structures. Also, thanks for your nice words. :)

As the amount of code is rather large, I created this as a draft to discuss on how to merge it. I would propose to split this PR into the following parts:

- CoclosedMonoidal*
- SymmetricCoclosedMonoidal*
- RigidSymmetricCoclosedMonoidal*
- Create*MonoidalCategories (the rewriting mechanism used to create the Topos package)
- Coclosed methods and examples for LinearAlgebraForCAP

Also, if you wish, I can send you my thesis which has TikZ diagrams for almost every method/derivation/etc. Since a high amount of the code translates directly into diagrams, this should make it far less painful to review. :)

PR #531, opened 2020-05-18
CoclosedMonoidal*

TKuh:
As discussed in https://github.com/homalg-project/CAP_project/pull/510 here is the first part.

- In my thesis the unit and counit of the cohom-tensor adjunctions are called "coclosed evaluation" and "coclosed coevaluation". However, initially I called them `CoEvaluation` and `DualCoEvaluation`, which is certainly bad but I couldn't come up with something better. This will change later via https://github.com/TKuh/CAP_project/commit/316798cd1b0c79f2d71b7a51b0465e952e6e491d. 

- You should probably look closely at `*MethodRecord.gi`, especially the `io_type`. I couldn't find any information about it and mostly guessed its meaning. :/

### DIFF
--- a/MonoidalCategories/doc/Doc.autodoc
+++ b/MonoidalCategories/doc/Doc.autodoc
@@ -13,3 +13,5 @@
 @Section Symmetric Closed Monoidal Categories
 
 @Section Rigid Symmetric Closed Monoidal Categories
+
+@Section Coclosed Monoidal Categories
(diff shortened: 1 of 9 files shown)


sebastianpos (approved):
I'm again very sorry for the delay. I actually read your thesis when you sent it to me: It was great to see this stuff on coclosed categories written up, so thanks a lot for that.

Your PR looks nice, and if your thesis is available on the internet, then I would suggest to link it in the documentation. Otherwise, it could be hard to follow what's going on here since the notions are non-standard.

I checked the io_type entries in the MethodRecord, they look good to me.

And yes, the name `CoEvaluation` is indeed bad in this context ;-)

PR #540, opened 2020-05-26
use HomalgIdentityMatrix in BasisOfExternalHom

mohamed-barakat:
this code
* avoids data transfer when working with external rings
* activates the logic, especially when the source is the distinguished object;
  pre-composing alpha with the i-th basis morphism automatically translates into
  CertainRows( UnderlyingMatrix( alpha ), [ i ] )

This resulted in a measurable speedup in computations of BasisOfExternalHom
in FunctorCategories


mohamed-barakat:
> It would be great if you could provide the same code for the categories of rows, right here after the test if we are in the abelian case:

Done.

PR #541, opened 2020-05-30
added IsSkeletalCategory to CAP_INTERNAL_CATEGORICAL_PROPERTIES_LIST

mohamed-barakat:
and marked three categories to be skeletal:
* MatrixCategory
* CategoryOfRows
* CategoryOfColumns

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -5,7 +5,7 @@ PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
 
 Version := Maximum( [
-  "2020.05.02", ## Mohamed's version
+  "2020.05.16", ## Mohamed's version
   ## this line prevents merge conflicts
   "2015.04.01", ## Oystein's version
   ## this line prevents merge conflicts
--- a/CAP/gap/CAP.gd
+++ b/CAP/gap/CAP.gd
@@ -136,6 +136,7 @@ end );
 Perform(
 ## This is the CAP_INTERNAL_CATEGORICAL_PROPERTIES_LIST
     [ [ "IsEnrichedOverCommutativeRegularSemigroup" ],
+      [ "IsSkeletalCategory" ],
       [ "IsAbCategory" ],
       [ "IsLinearCategoryOverCommutativeRing" ],
       [ "IsAdditiveCategory" ],
--- a/FreydCategoriesForCAP/PackageInfo.g
+++ b/FreydCategoriesForCAP/PackageInfo.g
@@ -15,7 +15,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2019.11.02", ## Sepp's version
   ## this line prevents merge conflicts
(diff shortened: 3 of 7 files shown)


mohamed-barakat:
From inspecting `IsEqualForObjects` I assume that `QuiverRows` is also skeletal. If yes, I can amend the PR.

sebastianpos:
And no, `QuiverRows` is not skeletal, for the same reason why `GradedRows` is not skeletal. You have
```
S(0) + S(1)
```
and
```
S(1) + S(0)
```
as unequal but isomorphic objects, which is a good thing. If we forced these objects to be equal, we would need to sort the rows and columns of our homogeneous matrices all the time.

sebastianpos:
I just realized that the categories Rows and Columns are also not necessarily skeletal:
this depends on the ring having invariant basis number:
https://en.wikipedia.org/wiki/Invariant_basis_number

Could you please provide a fix that only sets this property in contexts where this property holds? For starters, you may simply use the `is_defined_over_field` property.

Thanks.

mohamed-barakat:
homalg rings support the properties HasLeft/RightInvariantBasisProperty. I will use them if this is OK for you.

sebastianpos:
Wikipedia states

> This form reveals that the definition is left–right symmetric, so it makes no difference whether we define IBN in terms of left or right modules; the two definitions are equivalent. 

Thus, I don't see what these homalg ring properties stand for.

mohamed-barakat:
Nice, I was unaware of this. I learned this from the book of T.Y. Lam "Lectures on Modules and Rings" where, if I remember correctly, he makes this distinction. I assume that he is aware of this and I simply missed the result in the book proved at a later point. In any case, I have the following rules in `MatricesForHomalg/gap/LIRNG.gi`:

```
          ## Is/Left/Right/Noetherian
          [ IsLeftNoetherian,
            "implies", HasLeftInvariantBasisProperty ], ## [Lam06, bottom of p. 26]

          [ IsRightNoetherian,
            "implies", HasRightInvariantBasisProperty ],## [Lam06, bottom of p. 26]          ## Is/Left/Right/Noetherian
          [ IsLeftNoetherian,
            "implies", HasLeftInvariantBasisProperty ], ## [Lam06, bottom of p. 26]

          [ IsRightNoetherian,
            "implies", HasRightInvariantBasisProperty ],## [Lam06, bottom of p. 26]
```

But I now see that I oversaw the case of the trivial ring. I will fix this in homalg first and then prepare a PR.

mohamed-barakat:
It is stated in Lam's book that the IBN is left-right symmetric. I wonder why I missed it.

mohamed-barakat:
I now understand my confusion: Lam 06 is not the book I mentioned above, it is "Serre’s Problem on Projective Modules".  The matrix characterization is stated a page later, but I missed its significance for the left-right symmetry, maybe because he didn't mention it explicitly. And back then I explicitly excluded the zero ring. When I changed my mind later I forget to update the deduction rules in LIRNG.gi.

PR #546, opened 2020-06-23
Add CompilerForCAP

zickgraf:
Finally, the pull request of `CompilerForCAP`, including an example in `LinearAlgebraForCAP` showing that the compiler could compile the primitive operation `MorphismBetweenDirectSums` on its own from the generic derivation.

I have written lots of documentation and hope its in a "useful" state, but some parts of it depend on knowledge about GAP's syntax trees, which do not seem to have documentation sadly :-(

I have not tested this with many examples, so applying this to other examples probably gives some "this is not supported" errors. See the FAQ in the manual for explanations for some expected errors.

@sebastianpos Do you want to review all of this, or is this "my" package and you only want to review the parts touching `CAP`? In the first case, we should probably meet for a review session, since I think some/many parts of the code need additional explanations. I would love to add these explanations as comments to the code, but for this I would first have to write documentation for GAP's syntax tree -> this would need a lot of time :-(

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -13,7 +13,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2020.04.16", ## Sepp's version
   ## this line prevents merge conflicts
-  "2020.04.27", ## Fabian's version
+  "2020.06.17", ## Fabian's version
   ## this line prevents merge conflicts
   "2020.01.10", ## Kamal's version
 ] ),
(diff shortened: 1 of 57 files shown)


sebastianpos:
I'm totally happy with you being in charge of the package `CompilerForCAP`.

If you want, you may merge a PR that adds your package to `CAP_project`, and only leave me with the changes that you had to do w.r.t. the other packages.

zickgraf:
Okay, perfect! For atomicity I would like to keep this as a single commit if possible. Could you review this and simply ignore everything in `CompilerForCAP`? If you view the changes in GitHub, all (mostly small) changes to `CAP` are at the beginning, and there are two small changes to `LinearAlgebraForCAP` and the makefile right at the end. Everything else belongs to `CompilerForCAP`.

zickgraf:
Thanks! Yes, please merge :-)

mohamed-barakat:
Congratulations Fabian and thank you for the wonderful work.

PR #547, opened 2020-06-25
Allow to use different matrix data structures for AdditiveClosure

zickgraf:
~Depends on #575.~

As mentioned before, I would like to make AdditiveClosure compatible with other matrix data structures, in particular homalg matrices (in the case that the UnderlyingCategory is a homalg ring). I wanted to use MatrixObj as an interface for this, but this is not advanced enough. Thus, I have the following proposal and would like to gather some feedback on this.

Concept behind this PR:
1. AdditiveClosure should really "invent" matrix calculus, that is, it should not make use of any matrix features.
2. To allow other data structures than lists of lists, AdditiveClosure should never use list accesses (i.e. `list[i]`) but only matrix accesses (i.e. `matrix[i,j]`).
4. If performance is relevant, rely on the new compiler.

This PR implements this concept:
1. I have replaced the + und - operations on matrices by explicit operations on the entries.
2. I have replaced all list accesses by matrix accesses.
3. Then it is easy to allow other matrix data structures, we only have to provide two operations: one for interpreting matrix elements as morphisms of the underlying category, and one interpreting lists of lists of such morphisms as matrices again.
4. I have implemented two new derivations: `ProjectionInFactorOfDirectSum` using `UniversalMorphismFromDirectSum` and the dual for injections.
5. I have removed `ProjectionInFactorOfDirectSum` (and injections), `MorphismBetweenDirectSums`, `UniversalMorphismFrom/IntoZeroObject` (all of these can be derived).

Now AdditiveClosure really "invents" matrix calculus in the sense that the remaining operations map exactly to matrix operations:
1. IdentityMorphism = IdentityMatrix
2. ZeroMorphism = ZeroMatrix
3. PreCompose = *
4. AdditionForMorphisms = +
5. UniversalMorphismFromDirectSum = UnionOfRows
6. UniversalMorphismIntoDirectSum = UnionOfColumns
7. HomomorphismStructureOnMorphisms = KroneckerMatrix (including the dual version I asked about)

In a sense, AdditiveClosure is now what I hoped MatrixObj would be.

Regarding performance: the runtime for the Freyd tests seems more or less unchanged on my system.

Note: Currently this PR has only one big commit. If you like the concept, I would split it into several commits so it's easier to review.

### DIFF
--- a/FreydCategoriesForCAP/doc/Doc.autodoc
+++ b/FreydCategoriesForCAP/doc/Doc.autodoc
@@ -2,6 +2,8 @@
 
 @Chapter Additive closure
 
+@Chapter Example on additive closure
+
 @Chapter Adelman category
 
 @Chapter Category of rows
--- a/FreydCategoriesForCAP/examples/AdditiveClosure.g
+++ b/FreydCategoriesForCAP/examples/AdditiveClosure.g
@@ -1,0 +1,92 @@
+#! @Chapter Example on additive closure
+
+#! @Section Using matrix data structures
+
+LoadPackage( "FreydCategoriesForCAP" );;
+
+#! @Example
+QQ := HomalgFieldOfRationalsInSingular();;
+
+R := QQ * "x,y,z";;
+
+matrix_element_as_morphism :=
+        r -> RingAsCategoryMorphism( RingAsCategory( R ), r );;
+
+list_list_as_matrix := {listlist, m, n} ->
+    HomalgMatrix(
+        List( listlist,
+            row -> List( row,
+                entry -> UnderlyingRingElement( entry )
+            )
+        ), m, n, R );;
+
+CR := RingAsCategory( R );;
+CRplus := AdditiveClosure( CR :
+    matrix_element_as_morphism := matrix_element_as_morphism,
+    list_list_as_matrix := list_list_as_matrix
+);;
+
+M := HomalgMatrix( "[[x^2,4*y]]", 1, 2, R );;
+N := HomalgMatrix( "[[1,3*x], [2*y^3,5*y]]", 2, 2, R );;
+P := M * N;;
+
+o := AsAdditiveClosureObject( RingAsCategoryUniqueObject( CR ) );;
+
+A := o;;
+B := DirectSum( o, o );;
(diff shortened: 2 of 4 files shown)


sebastianpos:
Yes, I like the concept, and would be glad if you split the PR into smaller pieces.

It would also be nice to have a list of specifications that the underlying data structure `MorphismMatrix` has to satisfy in order to be usable by the `AdditiveClosure`. For what I see so far, such a list includes:
-  a function `matrix_element_as_morphism`
- a function `list_list_as_matrix`
- an operation `[,]`

Also, we may think about optional values that might be pushed to the category constructor, like functions for composition, addition, etc. (in a sense, this could lead us to the original idea of MatrixObj "restricted" to entries that have to be morphisms in a category. I put restricted in quotes since I think for MatrixObj, the only condition for two matrices being composable is having compatible numbers of rows and columns, which is not true in the additive closure).

Is it correct that in your application (which I assume uses HomalgMatrices), the compiler already optimizes the composition to the product of homalg matrices?

zickgraf:
Perfect, I will split the PR into smaller pieces and, of course, also update the documentation (the list you mention as a specification is correct!).

> Also, we may think about optional values that might be pushed to the category constructor, like functions for composition, addition, etc.

I think such data should be incorporated via the compiler (as logic templates). This way it can be reused at all places where some code compiles to a composition, addition etc. (although I currently have no example for such a thing actually occurring).

> Is it correct that in your application (which I assume uses HomalgMatrices), the compiler already optimizes the composition to the product of homalg matrices?

Currently it does not, since `PreCompose` is not a bottleneck in my applications. However, writing a suitable logic template would only take a few minutes.

A general note: adding logic templates is rather expensive with regard to the compilation time, that's why currrently I only add those I really need. However, I have some ideas which should drastically improve the performance, I just need to find the time to implement them.

zickgraf:
With #575 merged, this is now ready for review.

sebastianpos (approved):
Looks very good :)

You could think about making homalg matrices the default choice of the additive closure whenever the input category is a `RingAsCategory` with an underlying homalg ring.

PR #549, opened 2020-06-27
 Add hom structure to RingAsCategory constructor

zickgraf:
for commutative and exterior homalg rings.

Note: I have not (yet) formally written down that what I do actually defines a homomorphism structure. If you say it's obvious, we can merge it, otherwise we should wait :D

~Depends on #547.~

### DIFF
--- a/FreydCategoriesForCAP/examples/AdditiveClosure.g
+++ b/FreydCategoriesForCAP/examples/AdditiveClosure.g
@@ -46,6 +46,11 @@ IsWellDefined( gamma );
 #! true
 MorphismMatrix( gamma ) = P;
 #! true
+delta := Lift( gamma, beta );;
+IsWellDefined( delta );
+#! true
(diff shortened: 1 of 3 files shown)


zickgraf:
This is now ready for review (modulo the note in the first comment).

sebastianpos (approved):
Looks good to me.

Have you thought about an interface for "ring and its multiplication map given as a fp module" yet, from which you could then derive a homomorphism structure?

zickgraf:
Yes, I have, and it is already included in the PR, just not exposed to the outside world. I need:

1. A generating system of the ring as a fp C-module.
2. The ring as a fp C-module (that is, the relations).
3. A function to interpret ring elements of R as row vectors in C (corresponds to mu, and is also used to build H).
4. A ring inclusion of C into R (so I can consider the row vectors in C as row vectors in R and multiply with the generating system).

If anyone needs this, I'm happy to discuss names and expose this to the outside world :-)

PR #550, opened 2020-06-27
Support resolving \[\,\] in the compiler

zickgraf:
~The first two commits are from #547 and #549.~

@sebastianpos The actual commit only touches files in `CompilerForCAP`. Shall I simply merge such commits myself, or shall I wait for your approval?


zickgraf:
> You may merge every commit that only touches your package `CompilerForCAP`, unless you want someone else looking at it first.

Perfect! I have extracted the relevant commit (which only touches `CompilerForCAP`) and will merge it once the CI passes.

PR #551, opened 2020-06-28
 Add hom structure to LeftPresentations 

zickgraf:
using FreydCategory( AdditiveClosure( RingAsCategory( homalg_ring ) ) ). Thanks to the compiler the performance is not too bad (although I have not tested it extensively yet).

Only the last commit is relevant.

Regarding RightPresentations: Am I correct that RightPresentations is equivalent to FreydCategory( OppositeCategory( AdditiveClosure( RingAsCategory( homalg_ring ) ) ) )? If yes, I would also add a hom structure to RightPresentations in the same way.


sebastianpos:
In the long run, I would like to have the package `ModulePresentationsForCAP` replaced by `FreydCategoriesForCAP` (the same holds for `LinearAlgebraForCAP`). In this sense, I do not encourage to put too much effort into this package. 

Could you elaborate a bit on what you did? In particular, is it true that the ideas of this PR could also be realized within the package `FreydCategoriesForCAP` itself (like compiler optimizations)?

zickgraf:
> Could you elaborate a bit on what you did? In particular, is it true that the ideas of this PR could also be realized within the package FreydCategoriesForCAP itself (like compiler optimizations)?

During the creation of a `LeftPresentations`-category I try to create `FreydCategory( AdditiveClosure( RingAsCategory( homalg_ring ) ) )`. If the latter has a homomorphism structure (which it has for commutative polynomial and exterior homalg rings thanks to #549), I simply transfer this homomorphism structure to `LeftPresentations` by translating objects from and to the Freyd category via the functions `obj_in_freyd`, `obj_from_freyd` etc.

For sensible performance I enable the compilation of `HomomorphismStructureOnMorphismsWithGivenObjects` for the `AdditiveClosure` (in my applications, the other operations are no bottlenecks). For this, some logic templates and functions are needed. Finally, I trigger a "pre-compilation" with small objects and morphisms (see https://github.com/homalg-project/CAP_project/pull/551/files#diff-49bbf7e69ea4577f86f29ad6778d7f19R2208).

So yes, we can also put the compilation logic in `FreydCategories` or `AdditiveCategories` and make users use `FreydCategories` directly. The reasons why I added it here are the following:
1. I thought that's what people are using right now.
2. Logic templates are currently rather expensive with regard to compilation performance, so I wanted to add them at a point where I know they are useful.
3. I have not looked at any operations of `AdditiveClosure( RingAsCategory( homalg_ring ) )` except the ones belonging to the homomorphism structure. Thus, all other operations will probably be painfully slow.

All three points will probably change in the future, but I'm not sure when I will get to points 2. and 3.

zickgraf:
In theory, this is now ready for review. However, the issues regarding dependencies and the speed of the compiler still exist, and I actually have some more improvements for lifts. Thus, I will keep this PR as a draft for now.

PR #553, opened 2020-07-22
Do not format functions printed to streams in the derivation mechanism

zickgraf:
GAP breaks lines when printing strings longer than `SizeScreen`. This can also break the names of operations. Such operations are not detected by `CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION` anymore. This leads to wrong derivations being installed or correct derivations not being installed. To test this, execute `make test` in a narrow terminal (e.g. 40 characters) or execute `SizeScreen( [ 40 ] );` before loading CAP. The fix is to disable the formatting of the stream before calling `PrintTo`.

Note: There are more calls to `PrintTo` in `GroupRepresentationsForCAP/gap/AssociatorsForRepresentationCategoryOfGroup.gi` which I have not touched. They might need such a handling, too.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -13,7 +13,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2020.04.16", ## Sepp's version
   ## this line prevents merge conflicts
-  "2020.06.17", ## Fabian's version
+  "2020.07.22", ## Fabian's version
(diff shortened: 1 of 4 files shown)


sebastianpos:
Thanks for spotting this! I'm not sure how this affects the writing to files as it is done for the associators.

PR #561, opened 2020-08-10
Declare functions for LaTeXOutput

sebastianpos:
for cells and categories, and implementations in the case of QRows and Adelman.

Planned usage: pretty print in Jupyter notebooks.


mohamed-barakat:
Very good idea :)

PR #563, opened 2020-09-05
Add documentation of subpackages to README

zickgraf:
Note: the first commit is the one from #562.

This PR adds links to the documentations of all subpackages which are part of `make ci-test` to the README. If a package has HTML documentation, this is used. If not, its `manual.pdf` is added to the branch `doc` (which I will have to create) automatically by the CI and then used for the link. You can view the result here: https://github.com/zickgraf/CAP_project/blob/documentation/README.md

I'm not 100% satisfied with the layout, so if you like the basic idea feel free to suggest improvements. Also, the order of the packages is more or less random right now and can be adjusted.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+codecov:
+  disable_default_path_fixes: true
+fixes:
+  - "/home/gap/.gap/pkg/CAP_project/::"
+ignore:
+  - "home/"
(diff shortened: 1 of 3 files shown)


sebastianpos (approved):

sebastianpos:
Thanks a lot for your effort.
I'll let you know when I have any suggestions.

PR #565, opened 2020-09-08
Apply PackageJanitor to CAP

zickgraf:
This PR applies PackageJanitor to CAP.

Changes (in order of appearance):
* use the `.gitignore` file provided by PackageJanitor
* add LICENSE
* provide a consistent header across all files
* add a nice header to README
* rename Intros.autodoc -> Doc.autodoc
* use the `clean` file provided by PackageJanitor
* move files in `examples/testfiles` one directory up and rename `.gi` files to `.g` in this directory
* use the makedoc.g file provided by PackageJanitor (uses AutoDocs `extract_examples`)
* add a `makedoc_with_overfull_hbox_warnings.g` which allows to test for overfull hboxes in the CI (this test is not activated as it does not pass right now)
* use the `makefile` provided by PackageJanitor (this drops the make targets `archive` and `towww` which I assume are obsolete anyway?)
* provide files in `tst` for use with AutoDocs `extract_examples`
* update main README to not include the "_" in "CAP_project"

Note: If you do not like some of the changes I can of course add appropriate exceptions to PackageJanitor. However, of course the fewer exceptions we have, the easier the maintenance will be in the future.

Note 2: And if you do not like the idea at all, of course we do not have to use PackageJanitor for CAP, this is just a suggestion!


CAP/gap/PrintingFunctions.gi
```diff
@@ -1,8 +1,8 @@
...
-##
-##                                               CAP package
-##
-##  Copyright 2016, Sebastian Gutsche, TU Kaiserslautern
+#
+# CAP: Categories, Algorithms, Programming
+#
+# Implementations
+#
 ##                  Sebastian Posur,   RWTH Aachen
```
sebastianpos, line 6: Why did my name survive the cleanup here (and in the instances below)? ;-)

sebastianpos (approved):
Thanks for the nice work. I like the idea very much that we can manage our administration with the help of a single package.

Concerning my question: I'm also happy to merge this PR right away if you say so.

CAP/gap/PrintingFunctions.gi
zickgraf: Thanks for spotting this! I indeed seem to have stopped at some point in the alphabet and missed all files after this one, not sure why :D Should be fixed :-)

sebastianpos (approved):

PR #566, opened 2020-09-09
Create LaTeXOutput for Rows and Freyd

sebastianpos:


FreydCategoriesForCAP/gap/CategoryOfRows.gi
```diff
@@ -1235,6 +1235,36 @@ InstallMethod( Display,
     
 end );
 
+##
+InstallMethod( LaTeXOutput,
+               [ IsCategoryOfRowsObject ],
+               
+  function( obj )
+    
+    return Concatenation( "R^{1 \\times ", String( RankOfObject( obj ) ), "}" );
```
kamalsaleh, line 1244: I would have done this:

C := CapCategory( obj );

if IsBound( C!.ring_latex_output ) then
   R := C!.ring_latex_output;
else
   R := "R";
fi;
return Concatenation( "{", R, "}^{1 \\times ", String( RankOfObject( obj ) ), "}" );

In case one works over a field or the ring of integer and with more than one ring in the same session.

sebastianpos: What is `ring_latex_output`?

kamalsaleh: The name of the ring in latex-code: for example "\\mathbb{Z}", or  "\\mathbb{Q}"

sebastianpos: I think the appropriate place for such LaTeX code would be in `RingsForHomalg`, preferably callable by `LaTeXStringOp( R )`, where R is a homalg ring. If such an operation were available, I would be happy to use it :)

kamalsaleh: True, would be nice to have such mechanism, to give a ring a proper LaTeX-name, as already done for entries and matrices. Such a name could also be used for other purposes.

kamalsaleh: Also one thing, the blue color does not stop where it should be. Try for example the code:

```
ZZ := HomalgRingOfIntegers();
ZZrows := CategoryOfRows( ZZ );
Print( LaTeXOutput( IdentityMorphism( IdentityMorphism( 3/ZZrows ) / FreydCategory( ZZrows ) ) ) );
```

kamalsaleh: You can use `{\color{blue} math-code}` instead of `\color{blue}{math-code}`

sebastianpos: Thanks, I'll do that. It's curious that it works fine within a jupyter notebook for me.

sebastianpos: Fixed

kamalsaleh: I think the loaded packages and their order might interfere with each other and cause such effects.

sebastianpos: You mean LaTeX packages? Yes, maybe.

kamalsaleh: Of course :)

PR #569, opened 2020-09-09
Apply PackageJanitor to more packages

zickgraf:
Notes:

* MonoidalCategories only has a subset of the features which PackageJanitor would usually introduce since it has a rather special makedoc file and I do not want to create conflicts with PR #531.
* I had to include some warning in the Freyd tests. These warnings did already exist but seem to not have been caught by the old test system. Please check that the tests pass for you.
* Freyd now includes the documentation in `gap/GradedModulePresentationsByFreyd/` which I assume was missing in the old makedoc.g by accident.
* I have dropped some empty AutoDoc comments while editing the headers. As a result, some chapters which were empty anyway are not included in the documentation anymore.

PS: These are the packages we need for our proposal, so for the time being this will probably be the last PackageJanitor PR :D


zickgraf:
I will have to look into the dropping code coverage, this is not expected.

zickgraf:
I have restarted the CI and now the code coverage is unaffected as expected. I assume there was some intermittent error.

sebastianpos (approved):
Thanks a lot. The tests pass for me.

I was not aware of the warnings in Freyd, I'll have a look at them myself.

zickgraf:
> I was not aware of the warnings in Freyd, I'll have a look at them myself.

Perfect! Note that FreydCategories itself triggers two such warnings: https://github.com/homalg-project/CAP_project/blob/d3a8df4d6c37ead70c20d4d5abaaaf68f826eadb/FreydCategoriesForCAP/tst/000_LoadPackage.tst#L8-L10

PR #570, opened 2020-09-10
LaTeX strings for semisimple categories

zickgraf:
This consists of multiple commits:

1. Move the "x_" in the output of GIrreducibleObjects from the view methods to the string (so the string actually functions as a name). This is not necessarily required but I think it makes things clearer.
2. Add version of `RepresentationCategoryObject` which allows to set the string of the underlying GIrreducibleObject (for example, to "\chi" or "χ"). Also include a convenience method for calling this with Julia strings.
3. Add `LaTeXStringOp` for objects and morphisms of semisimple categories. Uses the helper function `LaTeXStringOfSemisimpleCategoryObjectList`. Also add `IsShowable` so the LaTeX string is used in Jupyter notebooks.

Notes:
1. If you like I can add tests but since they would depend on homalg's LaTeX strings I fear they would break regularly, at least in the near future where all of the LaTeX stuff is still in flux.
2. We now have an inconsistency of names in CAP/homalg: `LaTeXString(Op)` vs `LaTeXOutput`. I think the term `LaTeXString` is better suited because `Output` might hint to something being displayed immediately. However, I might be biased here as my code already uses `LaTeXString(Op)` :D

### DIFF
--- a/GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
+++ b/GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
@@ -329,3 +329,59 @@ InstallMethod( Display,
     
 end );
 
+##
+InstallMethod( LaTeXStringOp,
+        "for a morphism in a semisimple category",
+        [ IsSemisimpleCategoryMorphism ],
+        
+  function ( morphism )
+    local morphism_list, source_object_list, range_object_list, source_object_string, range_object_string, diag_entries, matrix, string, elem;
+    
+    morphism_list := SemisimpleCategoryMorphismList( morphism );
+    
+    morphism_list := Filtered( morphism_list, pair -> not( IsZeroForObjects( Source( pair[1] ) ) or IsZeroForObjects( Range( pair[1] ) ) ) );
+    
+    if IsEmpty( morphism_list ) then
+        
+        Display( "0" );
+        
+    else
+        
+        source_object_list := List( morphism_list, m -> [ NrRows( UnderlyingMatrix( m[1] ) ), m[2] ] );
+        range_object_list := List( morphism_list, m -> [ NrCols( UnderlyingMatrix( m[1] ) ), m[2] ] );
+        
+        source_object_string := LaTeXStringOfSemisimpleCategoryObjectList( source_object_list );
+        range_object_string := LaTeXStringOfSemisimpleCategoryObjectList( range_object_list );
+        
+        diag_entries := [ ];
+        
+        for elem in morphism_list do
+            
+            matrix := UnderlyingMatrix( elem[1] );
+            
+            if NrRows( matrix ) = 1 and NrCols( matrix ) = 1 then
+                
+                Add( diag_entries, LaTeXStringOp( matrix[1, 1] ) );
+                
+            else
+                
+                Add( diag_entries, LaTeXStringOp( matrix ) );
+                
+            fi;
+            
+        od;
+        
+        string := "";
+        Append( string, source_object_string );
+        Append( string, " \\xrightarrow{\\operatorname{diag}(" );
+        Append( string, JoinStringsWithSeparator( diag_entries, "," ) );
+        Append( string, ")} " );
+        Append( string, range_object_string );
+        
+        return string;
+        
+    fi;
+    
+end );
+
+MakeShowable( [ "text/latex", "application/x-latex" ], IsSemisimpleCategoryMorphism );
--- a/GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
+++ b/GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
@@ -315,7 +315,7 @@ InstallMethod( Display,
         
         for elem in morphism_list do
             
-            Print( Concatenation( "Component: (x_", String( elem[2] ), ")\n" ) );
+            Print( Concatenation( "Component: (", String( elem[2] ), ")\n" ) );
             
             Print( "\n" );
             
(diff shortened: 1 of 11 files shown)


zickgraf:
Thinking about it, I am not sure my implementation for morphisms is correct. Maybe it is only correct for the subset of cases we have in our examples in `InternalModules`. I will have to think about this.

zickgraf:
> I am not sure my implementation for morphisms is correct.

Indeed, it turns out I was only thinking about isomorphisms :D I hope that everything is fine now.

GroupRepresentationsForCAP/gap/Julia.gi
```diff
@@ -0,0 +1,16 @@
+#
+# GroupRepresentationsForCAP: Skeletal category of group representations for CAP
+#
+# Implementations
+#
+
+##
+InstallOtherMethod( RepresentationCategoryObject,
```
sebastianpos, line 8: Could you explain why you need this function (and thus the optional dependency on JuliaInterface)?

GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
```diff
@@ -329,3 +329,72 @@ InstallMethod( Display,
...
+        Append( string, range_object_string );
+        
+        return string;
+        
+    fi;
+    
+end );
+
+##
+InstallMethod( IsShowable,
```
sebastianpos, line 388: This looks to me like a generic function that you always want to be installed whenever you want a GAP object to be showable (you install exactly the same function for `IsSemisimpleCategoryObject` below). Maybe that should be taken as a hint for the LaTeX interface that you implicitly propagate with IsShowable/LaTeXStringOp at the moment. I would suggest to outsource this code to a function within `ToolsForHomalg` called something like `ActivateIsShowable( filter )`.

GroupRepresentationsForCAP/gap/Julia.gi
zickgraf: I think we only want this for convenience so we can write
```
RepresentationCategoryObject( chi, RepG, "\\chi" );
```
in a Jupyter Notebook instead of
```
RepresentationCategoryObject( chi, RepG, g"\\chi" );
```

@mohamed-barakat Is my statement correct or is there some other reason I have missed?

GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
zickgraf: You are correct that right now this is rather generic, since we know that we can always generate sensible LaTeX strings. However, in more complicated cases I expect that there will be further checks, for example if we only want to support LaTeX output if the underlying category supports LaTeX output, or only if the output is reasonable small. So I am not sure if a general function `ActivateIsShowable( mime_type, filter )` would be widely used. But of course it would not hurt introducing such a function.

@mohamed-barakat Do you have an opinion on this?

GroupRepresentationsForCAP/gap/Julia.gi
mohamed-barakat: > @mohamed-barakat Is my statement correct or is there some other reason I have missed?

Yes, your statement is correct.

GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
zickgraf: I have created a PR: homalg-project/homalg_project#353

GroupRepresentationsForCAP/gap/Julia.gi
sebastianpos: Then the problem appears to be more general and maybe it would be desirable to have a solution on the Julia side?

If not, we would need to use `JuliaInterface` as an optional package and a method like this one for each instance of a GAP command which uses a string and that we want to call via Julia.

zickgraf: I have just pushed a version without the Julia convenience so the discussion does not block the PR. I have no strong opinion on this and will discuss with Mohamed later.

mohamed-barakat: Thanks, merged.

GroupRepresentationsForCAP/gap/SemisimpleCategoryMorphisms.gi
mohamed-barakat: Thanks, merged.

GroupRepresentationsForCAP/gap/Julia.gi
mohamed-barakat: Like in the other packages I would suggest separate files `Julia.gd/Julia.gi` with the following line in `init.g/read.g`

```
if IsPackageMarkedForLoading( "JuliaInterface", ">= 0.2" ) then
    ReadPackage( "GroupRepresentationsForCAP", "gap/Julia.gi");
fi;
```
```
if IsPackageMarkedForLoading( "JuliaInterface", ">= 0.2" ) then
    ReadPackage( "GroupRepresentationsForCAP", "gap/Julia.gi");
fi;
```

These files hence won't affect the package when used within normal `GAP`, but will be read if the package is loaded using `GAP.jl`.

mohamed-barakat: Sorry, I was confused by just looking at the current state of the PR. It seems you have just deleted `Julia.gi` from the PR.

sebastianpos (approved):
I'm fine with these changes.

I also like the name `LaTeXString` better, but I avoided it when I chose `LaTeXOutput` due to the conflicts with other Gap pkgs. I propose to defer the discussion about the name to another thread.

GroupRepresentationsForCAP/gap/Julia.gi
zickgraf: For reference: We now handle this as a special case in HomalgProject.jl and will create an issue in GAP.jl to handle it in a more general way in the future.

PR #572, opened 2020-09-10
Declare the operation LaTeXOutputOnlyMorphismData

kamalsaleh:
Sometimes one wants to print nicely only the data of the morphism without its source and range. This is helpful at least in complexes categories and algebroids and additive closures categories.

### DIFF
--- a/CAP/gap/CAP.gd
+++ b/CAP/gap/CAP.gd
@@ -530,6 +530,15 @@ DeclareGlobalFunction( "DisableAddForCategoricalOperations" );
 DeclareOperation( "LaTeXOutput", [ IsCapCategoryCell ] );
 
 #! @Description
+#! The argument is a morphism $c$.
+#! The output is a LaTeX string $s$ (without enclosing dollar signs) that may be used to print out nicely
+#! the data which are needed to reconstruct $c$ (e.g., by passing it to an appropriate constructor,
+#! possibly together with its source and range).
+#! @Returns a string
+#! @Arguments c
+DeclareOperation( "LaTeXOutputOnlyMorphismData", [ IsCapCategoryMorphism ] );
+
+#! @Description
 #! The argument is a category $C$.
 #! The output is a LaTeX string $s$ (without enclosing dollar signs) that may be used to print out $C$ nicely.
 #! @Returns a string


sebastianpos:
I would rather suggest to use options for a single method `LaTeXOutput`, like I do here
https://github.com/homalg-project/CAP_project/blob/c88c13df8c38bf8442938d2a93afeccaffbe03d7/FreydCategoriesForCAP/gap/AdelmanCategory.gi#L1292
or there
https://github.com/homalg-project/CAP_project/blob/c88c13df8c38bf8442938d2a93afeccaffbe03d7/FreydCategoriesForCAP/gap/QuiverRows.gi#L1577

Since these are already two different options, and since I don't know how many customizations we will need in the future, maybe it's wise to keep it simple and not to enhance the interface.

If you agree, I'm happy to discuss some standards for names of such options. The option `OnlyDatum` appears to be exactly what you were aiming for with this PR, and so we should agree on a standard name for that.

In contrast, I believe the option `HighlightMiddle` appears to be not of general interest.

kamalsaleh:
I think it is also valid, but one should know how to use the options, they could lead to unexpected behaviors. Since the entered option-value will be passed to all operations until the output is returned, unless one changes the option-value inside the methods. It is enough for my needs if we take the option `OnlyDatum` into consideration in the next `LaTeXOutput` implementations. Maybe other option would be `WithColor`?

kamalsaleh:
```
##
InstallMethod( LaTeXOutput,
               [ IsFreydCategoryMorphism ],
  function( mor )
    local datum, OnlyDatum;

    datum := LaTeXOutput( MorphismDatum( mor ) : OnlyDatum := true );

    OnlyDatum := ValueOption( "OnlyDatum" );

    if OnlyDatum = true then

       return Concatenation(
        """{\color{blue}{""",
        datum,
        """}}"""
      );

    else

      return Concatenation(
        "{ \\tiny ", LaTeXOutput( Source( mor ) ), "}",
        """{\color{blue}{\xrightarrow{""",
        datum,
        """}}}""",
        "{ \\tiny ", LaTeXOutput( Range( mor ) ), "}"
      );

    fi;

end );

```

sebastianpos:
Yes, an option in this case should only be passed if the underlying method directly makes use of it.
I would be happy if you provided a PR with your application to `FreydCategories`.

Is it okay for you if I close this PR?

kamalsaleh:
I will prepare the PR in a minute, and of course you can close this PR.

PR #575, opened 2020-09-12
Preparation for #547

zickgraf:
I finally took some time to split up #547 (to get rid of too many branches to rebase :D ). I have decided to not only split it into commits but even into two separate PRs, so this PR serves as a preparation for #547.

This PR makes `AdditiveClosure` not depend on lists (of lists) as the underlying data structure of morphisms anymore (with regard to the original input of the user, results of operations are still stored as lists of list). So theoretically, after this PR one can use any data structure having the operation `[,]`. However, one can still not use homalg matrices for example, since their entries are not CAP morphisms. This will be tackled in #547.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -21,7 +21,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2020.04.16", ## Sepp's version
   ## this line prevents merge conflicts
-  "2020.07.22", ## Fabian's version
+  "2020.09.01", ## Fabian's version
   ## this line prevents merge conflicts
   "2020.08.01", ## Kamal's version
 ] ),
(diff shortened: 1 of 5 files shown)


sebastianpos (approved):
Thanks, I'm fine with the changes. Note that methods like `MorphismBetweenDirectSums` where primitively added for the sake of performance, but since the additive closure category will be subject to the compiler, I assume this won't matter in the future.

zickgraf:
> Note that methods like `MorphismBetweenDirectSums` where primitively added for the sake of performance, but since the additive closure category will be subject to the compiler, I assume this won't matter in the future.

Exactly! Compiling the derivation of `MorphismBetweenDirectSums` already works well as can be seen in this example: https://github.com/homalg-project/CAP_project/blob/master/CompilerForCAP/examples/LinearAlgebraForCAP.g

Compiling `ProjectionInFactorOfDirectSum` et al. should work equally well, I think.

PR #577, opened 2020-09-15
Apply PackageJanitor

zickgraf:
In detail:
* Set SOURCE_DATE_EPOCH so the generated PDFs are reproducible (i.e. bit-for-bit identical if nothing has changed except the system time) -> less disk usage for the `doc` branch
* Make sure all files have a newline at the end -> better git diffs and less conflicts when adding new code at the end of a file
* Add tests for missing newlines at the end of files
* Make GroupRepresentations write newlines at the end of files when writing the associator data.


sebastianpos (approved):
Thank you so much :)

PR #585, opened 2020-09-21
Avoid "method installed for /" ... in QuiverRows

sebastianpos:
`IsQuiverVertex` appears to be a specialization of `IsPath`, which is the reason for the info message while loading the Freyd package.

@mohamed-barakat  if you are okay with this solution, we may discard #579


mohamed-barakat:
Yes, this the better solution. Thanks for investigating. The other PR may be closed.

mohamed-barakat (dismissed):
Could you please bump the version?

PR #587, opened 2020-09-22
Some small changes to the CAP manual

zickgraf:


sebastianpos (approved):
Thanks again for your analysis.

Issue #588, opened 2020-09-23
FinalDerivations with ConditionsListComplete always have weight 1


zickgraf:
Since https://github.com/homalg-project/CAP_project/commit/f148ff9fff8253602f13f063bfbc543dc7c6b59a, when computing weights, final derivations only consider operations automatically discovered in this loop:

https://github.com/homalg-project/CAP_project/blob/f148ff9fff8253602f13f063bfbc543dc7c6b59a/CAP/gap/Finalize.gi#L85-L97

If `ConditionsListComplete` is set to true, this loop is not executed and `collected_list` remains empty. Thus, the derivation gets assigned weight 1.

The only derivation affected by this (in the current CAP code) seems to be `BasisOfExternalHom`.

zickgraf:
Fixed by c5ff6260c9dcafe818e4818a39dcdbfe3a729adf.

PR #589, opened 2020-09-23
Fix func being an operation in CAP_INTERNAL_FIND_APPEARANCE_OF_SYMBOL_IN_FUNCTION

zickgraf:
If func is an operation MyOperation, previously "MyOperation" was not detected as a symbol.

Thus, for example the following final derivation
```
AddFinalDerivation( IsEqualForMorphisms,
                    [ [ IsCongruentForMorphisms, 1 ] ],
                    [ IsEqualForMorphisms ],
                    
  IsCongruentForMorphisms : Description := "Use IsCongruentForMorphisms for IsEqualForMorphisms" );
```
was installed with weight 1 instead of the weight of `IsCongruentForMorphisms` + 1.

Notes:
1. This is NOT related to #588.
2. I have not tested this extensively. The worst-case scenario should be that results change in some packages due to other derivations being triggered.
3. The proper way to do this in recent GAP version would probably be making use of syntax trees.


sebastianpos (approved):
It's good that you became suspicious of a derivation with weight 1. I overlooked that issue. Thanks for the fix, and yes, it would be better to base this code on meta programming tools.

PR #592, opened 2020-09-23
Fall back to IsIdenticalObj for IsEqualForCacheForObjects/Morphisms

### LINKED ISSUE #583: IsEqualForCacheForMorphisms does not default to IsEqualForMorphismsOnMor in all cases
From the documentation of `AddIsEqualForCacheForMorphisms`:

> By default [`IsEqualForCacheForMorphisms`] is an alias for `IsEqualForMorphismsOnMor`, where fail is substituted by false.

However, the following example shows that this is not necessarily true:

```
LoadPackage( "CAP" );

cat := CreateCapCategory( );
myobj := SymmetricGroup( 3 );
AddObject( cat, myobj );
mymor1 := GroupHomomorphismByImages( myobj, myobj );
AddMorphism( cat, mymor1 );
mymor2 := GroupHomomorphismByImages( myobj, myobj );
AddMorphism( cat, mymor2 );

AddIsCongruentForMorphisms( cat, function( a, b )
	
	return true;

end );

Finalize( cat );

Display( IsEqualForMorphismsOnMor( mymor1, mymor2 ) );
Display( IsEqualForCacheForMorphisms( mymor1, mymor2 ) );
```

displays

```
true
false
```

The reason is the following: there is a derivation setting `IsEqualForCacheForMorphisms` to `IsIdenticalObj` if `IsCongruentForMorphisms` is installed and `IsEqualForMorphisms` is not (yet) installed. In the above example, this derivation is triggered with weight 1. Afterwards, `IsEqualForMorphisms(OnMor)` is derived from `IsCongruentForMorphisms` and the usual derivation of `IsEqualForCacheForMorphisms` from `IsEqualForMorphismsOnMor` is triggered, but with weight 203, so it is not installed.

I'm not sure if this is a bug, or if the behaviour is intended (in order to not use the possibly expensive `IsCongruentForMorphisms` for cache comparisons) and the documentation should be changed.

Edit: Another problem: the final derivations for `IsEqualForCacheForMorphisms` to not have `IsEqualForCacheForMorphisms` in their exclude list and thus overwrite user a user given `IsEqualForCacheForMorphisms`.

zickgraf:
Fixes #583.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gd
+++ b/CAP/gap/CategoryMorphisms.gd
@@ -1707,8 +1707,9 @@ DeclareOperation( "IsEqualForCacheForMorphisms",
 #! @Description
 #!  By default, CAP uses caches to store the values of Categorical operations.
 #!  To get a value out of the cache, one needs to compare the input of a basic operation
-#!  with its previous input. To compare morphisms in the category, IsEqualForCacheForMorphism is
-#!  used. By default this is an alias for IsEqualForMorphismsOnMor, where fail is substituted by false.
+#!  with its previous input. To compare morphisms in the category, IsEqualForCacheForMorphisms is
+#!  used. By default, IsEqualForCacheForMorphisms falls back to IsEqualForCache (see ToolsForHomalg),
+#!  which in turn defaults to recursive comparison for lists and `IsIdenticalObj` in all other cases.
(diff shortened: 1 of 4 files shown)


sebastianpos (approved):
Thanks.

PR #596, opened 2020-10-01
Use homalg matrices as default data structure for AdditiveClosure( RingAsCategory( homalg_ring ) )

zickgraf:
Thanks for the suggestion!


mohamed-barakat (approved):

sebastianpos (approved):
You're welcome and thank you :)

PR #601, opened 2020-10-09
Use Jekyll redirect plugin

zickgraf:
It has multiple advantages like telling search engines to not index the
redirecting page.


sebastianpos (approved):

sebastianpos:
Thanks a lot for your good work!

PR #609, opened 2020-10-16
Add AssociatorData for trivial group

zickgraf:
The trivial group has no nice group string, so handle it as an exception.

Use `fail` to signal that no associator data is available since the empty
list and the empty string are equal.

I hope the empty list is mathematically correct, Mohamed and I have created it using `WriteAssociatorDataToFile` with some hacks to work around issues arising with the trivial group.


sebastianpos:
Yes, mathematically this is correct since it boils down to the category of vector spaces.

Technically, the following issue arises at the moment:

If I create
`RepresentationCategory( 1, 1 );`
then the result does not have associators.

If I create
`RepresentationCategory( TrivialGroup() );`
then the result does have associators.

Could you take a look at it?

zickgraf:
The same happens with
`RepresentationCategory( 6, 1 );`
vs.
`SymmetricGroup( 3 );`,
so I'm not sure what the "correct" fix would be. Should I add another entry to the database with name `1,1` also pointing to `1Ass.g`?

sebastianpos:
No, that's okay. I don't have a clean solution at the moment.

Thanks for taking a look!

sebastianpos (approved):

PR #612, opened 2020-10-22
Allow IsEqualForObjects/Morphisms to return fail

### LINKED ISSUE #595: Allowing fail for IsEqualForObjects
Objective:
I would like to allow to return the value `fail` in `IsEqualForObjects` (and maybe also for `IsEqualForMorphisms` as a second step).

Reasoning:
In `GradedCategories` our objects are infinite lists which we cannot compare in finite time. Thus, we have to use `IsIdenticalObj` for comparing objects and crisp caching for fulfilling the specifications. However, I have noticed two things:
1. On a technical level, CAP uses `IsEqualForObjects` only for input sanity checks and a small number of derivations (e.g., IsEndomorphism, IsIdenticalToIdentityMorphism, ...). If I do neither need input sanity checks (of course, my code is always 100% perfect!!! :D) nor one of the derivations, on a technical level I do not need `IsEqualForObjects`.
2. On an abstract level I know that all of the algorithms give equal output for equal input, that is, for any finite sublist the specification is actually fulfilled.

Thus, since all results I consider only depend on finite sublists, I think I do not actually have to use crisp caching. However, if I disable it, of course the sanity checks fail. I have tested disabling both the caching and the input sanity checks and all tests pass.

I have thought about how one could properly support this in CAP (except simply disabling input sanity checks), and I think returning `fail` for `IsEqualForObjects` for two non-identical objects would express exactly what I want: "Sorry, I do not know how to compare these two objects in finite time, but you can assume that you can work with them".

Implementation:
I would adapt the input sanity checks to simply continue if `fail` is returned by `IsEqualForObjects`, and think about the derivations on a case by case basis.

Question:
Have I missed something why I would really need crisp caching in this case? If not, do you like the idea?

Note: This is not urgent right now for me, but I expect that I would like to have this in the future, for example when using `CompilerForCAP` on categories with `IsIdenticalObj` for the equality of objects.

zickgraf:
Fixes #595.

The first commit removes the final derivations falling back to `ReturnFail` as discussed in #595. In contrast to the discussion in #595 I decided not to install another fallback, as I think the user should always be clear on his/her notion of equality. Thus, I added suitable sanity checks in `IsFinalized` instead, forcing the user to provide a notion of equality. With this, I noticed that `Cat` and `TerminalCategory` actually relied on the previous behaviour, so I adjusted them.

The second commit relaxes the pre functions as discussed in #595, and adds suitable warnings to the documentation. I deliberately only changed the documentation of the `AddIsEqual...` functions and kept the documentation of the `IsEqual...` functions as is, because I think of this as an expert feature which the normal user reading what `IsEqual...` does should not have to think about.

The last commit uses the new capabilities to give a more intricate notion of equality to the (Co)ActionsCategory in the non-computable case, which allows to disable the caching.

### DIFF
--- a/ActionsForCAP/gap/ActionsCategory.gi
+++ b/ActionsForCAP/gap/ActionsCategory.gi
@@ -607,13 +607,46 @@ InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
     
     else
         
-        SetCachingOfCategoryCrisp( category );
+        ##
+        AddIsEqualForObjects( category,
+          function( action_object_1, action_object_2 )
+            local object_equality, structure_morphism_equality;
+            
+            object_equality := IsEqualForObjects( ActionDomain( action_object_1 ), ActionDomain( action_object_2 ) );
+            
+            if object_equality = false then
+                
+                return false;
+                
+            elif object_equality = fail then
+                
+                return fail;
+                
+            else # object_equality = true
+                
+                structure_morphism_equality := IsEqualForMorphisms( StructureMorphism( action_object_1 ), StructureMorphism( action_object_2 ) );
+                
+                if structure_morphism_equality = true then
+                    
+                    return true;
+                    
+                else
+                    
+                    return fail; # we would have to decide IsCongruentForMorphisms
+                    
+                fi;
+                
+            fi;
+            
+        end );
         
         ##
-        AddIsEqualForObjects( category, IsIdenticalObj );
-        
-        ##
-        AddIsEqualForMorphisms( category, IsIdenticalObj );
+        AddIsEqualForMorphisms( category,
+          function( morphism_1, morphism_2 )
+            
+            return IsEqualForMorphisms( UnderlyingCell( morphism_1 ), UnderlyingCell( morphism_2 ) );
+            
+        end );
         
         ## cannot AddIsCongruentForMorphisms
         category!.is_computable := false;
--- a/ActionsForCAP/gap/ActionsCategory.gi
+++ b/ActionsForCAP/gap/ActionsCategory.gi
@@ -578,33 +578,33 @@ InstallGlobalFunction( ADD_FUNCTIONS_FOR_LEFT_AND_RIGHT_ACTIONS_CATEGORY,
     underlying_category := UnderlyingCategory( category );
     
     if CanCompute( underlying_category, "IsCongruentForMorphisms" ) then
-    
-    ##
-    AddIsEqualForObjects( category,
-      function( action_object_1, action_object_2 )
         
-        return IsEqualForObjects( ActionDomain( action_object_1 ), ActionDomain( action_object_2 ) )
-               and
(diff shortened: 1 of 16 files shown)


zickgraf:
I have added two more commits:

1. There are even more categories relying on the old behaviour. Thus I have added a way to suppress the new warning.
2. I have fixed the indentation in (Co)ActionsCategory.

sebastianpos:
This is not quite what I expected. I imagined that not providing `IsEqualForObjects`, `IsEqualForMorphisms` means that
`CanCompute( IsEqualForObjects, cat )` yields `false`, and calling `IsEqualForObjects` on objects in `a,b` triggers a fallback given by `if IsIdentityObject(a,b) then return true; else return fail; fi`. The advantage of that design is that no derivations would be falsely triggered.

Is it really beneficial to have `IsEqualForObjects`, `IsEqualForMorphisms` as partial functions (with the downsides like partial derivations and suppression options), or would it suffice (in your applications) to simply not add them in the sense I explained above?

zickgraf:
> This is not quite what I expected.

Yes, as mentioned in the description I have deviated from what we have discussed in #595, because I think forcing the user to think about his/her notion of equality is a good thing, for example for uncovering categories relying on the removed final derivation.

> The advantage of that design is that no derivations would be falsely triggered.

I would not call them "falsely triggered", but "partial derivations". But of course I understand your concerns; this is certainly not "nice". That's the reason why I have not adjusted the documentation of `IsEqual...`, so it is clear, that doing what I want to do is not intended and beyond "safe waters".

> Is it really beneficial to have `IsEqualForObjects`, `IsEqualForMorphisms` as partial functions?

I think yes, as I can compare "infinite objects" up to, say, the first 100 degrees and get `false` if they don't match. So I can recover quite a bit of type checking of my calculations (which only involve finitely many degrees) by comparing enough degrees, even if I do not know the exact value of what "enough" is.

However: all I really need from this PR are the changes to the method record. The other changes (forcing the user to think about the notion of equality, adding at least a bit of documentation about this case, etc.) are, from my point of view, significant improvements over the current state (which is, having the "ugly" final derivation, having some categories relying on that without anyone knowing, etc.), but not strictly necessary for what I want to achieve. So if you do not like the other changes, I can of course revert them.

sebastianpos:
I agree with the following:
- having final derivations for equalities removed
- your adjustments of the method record

I'm still struggling with partial functions. What do you think about the following idea:
- `AddF` should only be called if `F` can be realized by a total function, i.e., a function that produces output for every input, and not on a subset of its specified input
- If you want to have `F` partially available in your category, use `InstallMethod` instead of `AddF`. By doing so, we will not tell `CAP` to trigger derivations that were not designed for dealing with partial functions.

In your example of infinite objects, that would mean to use `InstallMethod` on `IsEqualForObjects` and filters that you specified for your objects.

zickgraf:
That would be fine for me! To simplify the discussion, I will open separate PRs for the separate points (knowing that we would discuss this in such detail I would have split this into separate PRs from the beginning on, sorry).

PR #614, opened 2020-11-01
remove link to CAP_project subpage

mohamed-barakat:
as discussed with Fabian:

since the CAP_project subpage has now less information than README.md
it can be deleted for the time being


sebastianpos (approved):
Ok, we can do that.

PR #615, opened 2020-11-02
Apply PackageJanitor

zickgraf:
The result can be seen here: https://github.com/zickgraf/CAP_project/tree/PackageJanitor


sebastianpos (approved):
Looks good.

PR #619, opened 2020-11-06
Fix product in CapCat

zickgraf:
Regressed by 3523cc1baec93ace21ce734ae9f41730b4726a59 (and missed in 79a5a978d09125553f37d46893ac697ee741c89b).

This has been broken for six years, so nobody seems to use it :D But I came across it while preparing #612. I have not tested the universal morphism, there might be more bugs.


sebastianpos (approved):
Thanks, and yes, apparently, these functor operations are not used.

PR #620, opened 2020-11-06
Relax type checks in pre functions

zickgraf:
`fail` should simply skip the pre functions. Originally part of #612.


zickgraf:
The application which motivated this PR is dead, so I will close the PR.

PR #621, opened 2020-11-06
Remove final derivations regarding equality

zickgraf:
This should be the solution with a fallback described in https://github.com/homalg-project/CAP_project/pull/612#issuecomment-722307457.

I do not think this is a good idea (that's why I open it as a draft): if a novice user does not provide `IsEqualFor...`, now instead of giving a warning, we provide an implicit fallback to something which is not even covered by the specification of `IsEqualFor...`.

My suggestion would be forcing the user to give a notion of equality (with an option to override this restraint) as I did in #612. But as I'm not affected by this (I hope I will never forget to provide an equality :D ), we can also merge this "as is" if you like.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gi
+++ b/CAP/gap/CategoryMorphisms.gi
@@ -351,7 +351,17 @@ InstallMethod( IsEqualForMorphisms,
     if not IsIdenticalObj( CapCategory( morphism_1 ), CapCategory( morphism_2 ) ) then
         Error( Concatenation( "the morphism \"", String( morphism_1 ), "\" and the morphism \"", String( morphism_2 ), "\" do not belong to the same CAP category" ) );
     else
-        Error( Concatenation( "the morphism \"", String( morphism_1 ), "\" and the morphism \"", String( morphism_2 ), "\" belong to the same CAP category, but no specific method IsEqualForMorphisms is installed. Maybe you forgot to finalize the category?" ) );
+        
+        if IsIdenticalObj( morphism_1, morphism_2 ) then
+            
+            return true;
+            
+        else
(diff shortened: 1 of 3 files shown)


zickgraf:
This PR was part of a larger series which I do not need anymore. However, I think the idea here (respectively my counter suggestion) is still worth discussing.

Recap:

Currently, there are final derivations setting `IsEqualForObjects/Morphisms` and `IsCongruentForMorphisms` to `ReturnFail` if not provided (where `IsIdenticalObj` is handled by redirect functions). This is bad for multiple reasons:
1. These operations play a crucial role in the specification of all other CAP operations (equal input gives equal output), so just assuming some equality will probably not give consistent results (at least one would have to enable strict caching if this should have the chance to be correct).
2. `fail` itself is not even allowed in the specification of `IsEqualForObjects/Morphisms` and `IsCongruentForMorphisms`.
3. Some operations like `IsIdenticalToZeroMorphism` will be derived although the derivation cannot handle the `fail` case.
4. A user who does not know that one has to install the equalities will run into really weird errors (usually some pre function will check an equality and if the output is `fail`, the output check of the equality will fail).

@sebastianpos Has proposed to get rid of the final derivations with a global fallback to `ReturnFail`. This is implemented in this PR and resolves points 2. and 3. However, it does not address points 1. and 4. I would instead suggest my solution in b19b21c422711f7281e34183dd90a8c20a0b7a16 and 833263459a3aa351c369aa8c232e73cd22455fc1 which actively forces the user to think about the equalities.

zickgraf:
Superseded by https://github.com/homalg-project/CAP_project/pull/1376.

PR #629, opened 2021-02-11
Refactor sanity checks

zickgraf:

### DIFF
--- a/CAP/doc/AddFunctions.autodoc
+++ b/CAP/doc/AddFunctions.autodoc
@@ -121,8 +121,9 @@ The record can have the following components, used as described:
 
 * post_function (optional): A function which is used as the postfunction of the installed methods, as described above.
 
-* filter_list: A list containing the basic filters for the methods installed by the add methods. Possible are filters, or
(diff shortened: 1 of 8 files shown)


zickgraf:
This PR will possibly break categories returning `fail` for `IsEqualForObjects` etc. This should be handled as part of #621: If we do not allow `IsEqualForObjects` to return `fail` when installing it via `AddIsEqualForObjects`, this PR is actually fine. If we allow this, I will introduce another return type `bool_or_fail`.

sebastianpos:
I get the following error:
```
gap> LoadPackage( "Freyd" );
Error, the return_type of <current_rec> is not a filter and does not appear in CAP_INTERNAL_VALID_RETURN_TYPES
```

zickgraf:
Can you type `current_rec` to see which record is affected? I assume it's this issue: https://github.com/homalg-project/HigherHomologicalAlgebra/issues/25 Then you can simply update `HigherHomologicalAlgebra` :-)

sebastianpos (approved):
Works, thank you.

PR #630, opened 2021-02-12
Demo: Using typed lists

zickgraf:
First some context for @sebastianpos: The motivation behind this PR is simplifying things for CompilerForCAP. It is based on discussions with @mohamed-barakat. Feel free to chime in, but as I raise some points which I have not discussed with Mohamed, you can also wait until Mohamed and I agree on our preferred way forward if you like :D

This PR is a demo of how we could use typed lists with attributes (https://github.com/homalg-project/homalg_project/pull/391) to:
1. get rid of method selection objects/morphisms,
2. handle corner cases like empty limits (e.g. https://github.com/homalg-project/CAP_project/blob/master/FreydCategoriesForCAP/gap/AdditiveClosure.gi#L951) without if/else statements.

Both points would be especially useful for `CompilerForCAP`, but are also useful feats on their own, I think.

This PR shows how this could work out using `DirectProduct` as an example. See the last commit on https://github.com/zickgraf/FinSetsForCAP/tree/typed_lists for tests.

Notes:
1. This change would be completely transparent to the end user due to convenience functions in `UniversalObjects.gi`.
2. These convenience functions could be created automatically.
3. Since typed lists offer the same interface as normal lists, one would only have to touch existing categories if one really wants to support empty limits etc.
4. In the case of `DirectSum` and `DirectProduct` we have to wrap the existing GAP functions as they throw an error on empty lists.
5. Some derivations like deriving `DirectProduct` from `ProjectionInFactorOfDirectProduct` would not work anymore as they require a non-empty list.
6. On the other hand, we could now derive `ZeroObject` as the empty direct sum.
7. Example of the syntax for the universal morphism into the empty direct sum with given test object:
```
UniversalMorphismIntoDirectProductWithGivenDirectProduct( ListOfObjectsInCategory( [ ], cat ), ListOfMorphismsInCategory( [ ], cat, Source, test_object ), DirectProduct( ListOfObjectsInCategory( [ ], cat ) ) )
```

The syntax is obviously very verbose and I'm not sure this is what we actually want. Of course, mathematically we have a "list of morphisms with a common source `test_object`" in mind. But maybe we could make things simpler.

#### Idea 1
For method selection we do not really have to know if something is a list of objects/morphisms/lists etc. So we could only provide a single short constructor `ListInCat(egory?)` yielding a list of \<whatever\> in a given category.

#### Idea 2
We could only require that one passes typed lists when actually needed for method selection (e.g. not in the WithGiven case), and in this case only for the first argument (and of course for all arguments which need attributes).

#### Idea 3
Instead of using typed lists with attributes, we could
1. allow to pass the category as the first argument of any CAP operation,
2. pass the attributes as explicit arguments

(of course again transparent for the user via suitable convenience functions).
Then
```
UniversalMorphismIntoDirectProductWithGivenDirectProduct( cat, [ ], [ ], test_object, DirectProduct( cat, [ ] ) )
```
would be much more readable.

Passing the test_object as an additional argument could even make sense when combined with typed list:
1. The io_type could explicitely access the test_object.
2. The full specification simply involves the test_object, so we should pass it as an argument. (Context of this statement: There is a convenience function `UniversalMorphismIntoDirectProduct` where you do not have to pass the product diagram. But since the diagram is part of the specification, we only offer this as a convenience function and not for the actual implimentation. I think the same reasoning could be applied to the test_object).
3. This would work nicely with idea 2 (we would never need attributes and often the test_object would completely suffice as a method selection object).

I once discussed parts of idea 3 with Mohamed and I think he was against it, but I'm not sure we were aware of all the pros and cons of the concepts.

### DIFF
--- a/CAP/gap/CAP.gd
+++ b/CAP/gap/CAP.gd
@@ -96,6 +96,38 @@ DeclareCategory( "IsCapCategoryMorphism",
 DeclareCategory( "IsCapCategoryTwoCell",
                  IsCapCategoryCell );
 
+#! @Description
+#! The GAP category of lists of CAP category cells.
+#! Every typed list of objects, morphisms, or $2$-cells
+#! of a CAP category lies in this GAP category.
+#! @Arguments object
+DeclareCategory( "IsListOfCapCategoryCells",
+                 IsListWithAttributes );
+
+#! @Description
+#! The GAP category of lists of CAP category objects.
+#! Every typed list of objects of a CAP category lies in
+#! this GAP category.
+#! @Arguments object
+DeclareCategory( "IsListOfCapCategoryObjects",
+                 IsListOfCapCategoryCells );
+
+#! @Description
+#! The GAP category of lists of CAP category morphisms.
+#! Every typed list of morphisms of a CAP category lies in
+#! this GAP category.
+#! @Arguments object
+DeclareCategory( "IsListOfCapCategoryMorphisms",
+                 IsListOfCapCategoryCells );
+
+#! @Description
+#! The GAP category of lists of CAP category $2$-cells.
+#! Every typed list of $2$-cells of a CAP category lies in
+#! this GAP category.
+#! @Arguments object
+DeclareCategory( "IsListOfCapCategoryTwoCells",
+                 IsListOfCapCategoryCells );
+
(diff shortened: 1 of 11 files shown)


zickgraf:
We have decided against using typed lists for this. I have created a new proposal in #643.

PR #631, opened 2021-02-19
Remove installation of Lift and Colift from additive closure

zickgraf:
These are exactly the existing derivations from SolveLinearSystemInAbCategory.


sebastianpos (approved):
Yes, you are right.

Thank you!

PR #632, opened 2021-02-19
Add test covering IsEqualForObjects in VectorSpacesAllMethods.g

zickgraf:
Due to caching and garbage collection, the line is sometimes covered but not
always. Make sure it is always covered to prevent unexpected code coverage
drops.

For an example of an unexpected code coverage drop see #631.


sebastianpos (approved):
Thanks for the example.

Issue #594, opened 2020-09-25
LinearAlgebraForCAP and CategoryOfRows/Columns do not fulfil specifications


zickgraf:
Consider the following code:
```
LoadPackage("LinearAlgebra");
LoadPackage("RingsForHomalg");
QQ := HomalgFieldOfRationalsInMAGMA();
M := HomalgMatrix( "[0,0,0,0]", 2, 2, QQ );
N := HomalgMatrix( "[0,0,0,0]", 2, 2, QQ );
IsZero(N);
phi := AsVectorSpaceMorphism( M );
emb_phi := KernelEmbedding( phi );
psi := AsVectorSpaceMorphism( N );
emb_psi := KernelEmbedding( psi );

Display( IsEqualForMorphisms( phi, psi ) );
Display( IsEqualForMorphisms( emb_phi, emb_psi ) );
```
outputs
```
true
false
```

The same happens for CategoryOfRows.

The reason is that in MAGMA, SyzygiesOfRows of a zero matrix returns a matrix with ones on the counter-diagonal. However, if homalg knows that a matrix is 0, it returns the identity matrix for SyzygiesOfRows.

Notes:
1. The same could happen any time for Singular, for example after a change as in https://github.com/homalg-project/homalg_project/pull/318 For example, if we would set the option `returnSB` in Singular (which applies `std` (=BasisOfColumns) to the output of the `syz` (=SyzygiesOfColums) command automatically), this would happen for Singular as well, since `BasisOfRows( IdentityMatrix )` returns a matrix with ones on the counter-diagonal since the PR linked above was merged.
2. I think fixing this with caching would need the "really crisp" caching we talked about in the other issue.

sebastianpos:
Wow, thanks for that brilliant example. 

Yes, really crisp caches would help here. However, I personally would not like that solution very much, since Rows/Cols/MatrixCategory should be regarded as workhorse categories, and a really crisp caching does not comply with this point of view. But in order to provide a really save version of these categories, I also don't see another solution.

[...]

PR #634, opened 2021-02-22
Remove superfluous special cases in CategoryOfRows/Columns

zickgraf:
ConvertMatrixFrom/ToRow/Column should be able to handle empty matrices without problems.

@mohamed-barakat Can you confirm this statement?

### DIFF
--- a/FreydCategoriesForCAP/gap/CategoryOfColumns.gi
+++ b/FreydCategoriesForCAP/gap/CategoryOfColumns.gi
@@ -786,21 +786,11 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_CATEGORY_OF_COLUMNS,
         ##
         AddInterpretMorphismAsMorphismFromDistinguishedObjectToHomomorphismStructure( category,
           function( alpha )
-            local underlying_matrix, nr_columns;
+            local underlying_matrix;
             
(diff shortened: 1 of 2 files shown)


mohamed-barakat (approved):
Yes, I had these corner cases in mind when I wrote the Convert* commands and tested them all. So I assume they still work.

sebastianpos (approved):
This code is much nicer and appears to work :)

mohamed-barakat:
I now remember why I did not do this simplification, although they would behave correctly worked (sorry, this was not present in my mind when I saw the PR). I wanted to go through the cache of `UniversalMorphismIntoZeroObject` and `ZeroMorphism` in the corner cases to avoid constructing new morphisms. Maybe these premature optimization ideas are slowly becoming obsolete with `CompilerForCAP`.

zickgraf:
> I wanted to go through the cache of `UniversalMorphismIntoZeroObject` and `ZeroMorphism` in the corner cases to avoid constructing new morphisms.

According to https://github.com/homalg-project/CAP_project/issues/594#issuecomment-698805777, Rows/Cols/MatrixCategory should be workhorse categories not relying on caching extensively. So if one wants/needs the caching, one probably has to adapt the category anyway, I assume.

mohamed-barakat:
I am happy with this :)

PR #636, opened 2021-02-23
fixed bug in membership test_function

mohamed-barakat:
the assumption made on HilbertPolynomial in the code is wrong

I understood from @HereAround that he has transferred some of the functionality of `ToricSheaves`. In this case, these fixes should also be relevant to his code.


HereAround:
> the assumption made on HilbertPolynomial in the code is wrong
> 
> I understood from @HereAround that he has transferred some of the functionality of `ToricSheaves`. In this case, these fixes should also be relevant to his code.

@mohamed-barakat Thank you for the pointer. Your critical change has already been fixed in https://github.com/homalg-project/ToricVarieties_project/blob/7dc7c5af027e4d9e52b3032ac7d77cb97a7e8a2a/CoherentSheavesOnToricVarieties/gap/CoherentSheaves.gi#L29.

Note that this line is identical to your change, except that `degree_matrix_as_list_list` has been renamed into `degrees`. Also `comes_from_smooth_variety` is identical to testing if the underlying variety is smooth.

sebastianpos:
Just to be sure, are you confident that the Hilbert polynomial works even under the new hypotheses?

The following example makes me doubt it, but maybe I get something wrong:
If you remove from the projective space the hyperplane x_0 = 0, you get a smooth toric variety with the same Cox ring as the projective space, but enlarged irrelevant locus. The Hilbert polynomial should not be able to detect that.

mohamed-barakat:
> Just to be sure, are you confident that the Hilbert polynomial works even under the new hypotheses?

One also needs that the irrelevant locus is zero-dimensional. The Hilbert polynomial invoked by the code is the classical Hilbertpolynomiall assuming all degrees equal to 1.

> The following example makes me doubt it, but maybe I get something wrong:
> If you remove from the projective space the hyperplane x_0 = 0, you get a smooth toric variety with the same Cox ring as the projective space, but enlarged irrelevant locus. The Hilbert polynomial should not be able to detect that.

I probably misunderstand your example. If I delete x_0 = 0 I get the affine space with one ray-generator less (so different Cox ring) and trivial degree group. What is the fan in you example?

sebastianpos:
> One also needs that the irrelevant locus is zero-dimensional. The Hilbert polynomial invoked by the code is the classical Hilbertpolynomiall assuming all degrees equal to 1.

Does this comment imply that the new hypotheses are still insufficient? 

What other cases beside projective spaces can be covered by the Hilbert polynomial? If there are none, then maybe the if clause should be replaced by checking whether we simply have a projective space?

I'm not familiar with the intended framework of this package. Does every modelled toric variety needs to come from a fan? Or is it sufficient to supply a Cox ring together with an irrelevant ideal? The second framework is more general, since it allows my given counter-example (the affine space modelled in an usual way).

mohamed-barakat:
> Does this comment imply that the new hypotheses are still insufficient?

Yes.

> What other cases beside projective spaces can be covered by the Hilbert polynomial? If there are none, then maybe the if clause should be replaced by checking whether we simply have a projective space?

For example the blowup of A^2 at the origin. The resulting normal toric variety is smooth with class group Z and irrelevant locus <x,y>.

> I'm not familiar with the intended framework of this package. Does every modelled toric variety needs to come from a fan? Or is it sufficient to supply a Cox ring together with an irrelevant ideal? The second framework is more general, since it allows my given counter-example (the affine space modelled in an usual way).

Since this package is based on `ToricVarieties` I assumed it requires normality.

sebastianpos:
> 
> 
>     Does this comment imply that the new hypotheses are still insufficient?
> 
> Yes.

Do you want this PR to be merged anyway? Or do you intend to correct it?

> For example the blowup of A^2 at the origin. The resulting normal toric variety is smooth with class group Z and irrelevant locus <x,y>.

The Cox ring in this example is k[t,x,y] with deg(t) = -1, deg(x) = deg(y) = 1.
The current if-clause excludes this example since one of the degrees is negative, if I'm not mistaken.
My question remains: can this if-clause simply be reduced to checking whether we have a projective space?

mohamed-barakat:
You are right, the easiest is to restrict to the projective space and generalize at a later point with an appropriate notion of a multigraded Hilbert polynomial. I will adapt the PR.

mohamed-barakat:
For the record:
* The input of method for [`CategoryOfToricSheaves`](https://github.com/homalg-project/CAP_project/blob/8614fee4e9e94c4b00bc354d55eba79e4e5c859f/ToricSheaves/gap/ToricSheaves.gi#L19) is (`graded_ring`, `irrelevant_ideal_generators`, and a boolean `comes_from_smooth_variety`).
* This method is called from [another method](https://github.com/homalg-project/CAP_project/blob/8614fee4e9e94c4b00bc354d55eba79e4e5c859f/ToricSheaves/gap/ToricSheavesForToricVarieties.gi#L10) for the same operation which takes a normal toric variety as input. This is the only way how the above method is invoked in the examples.
* The package `ToricVarieties` has a method [`IsIsomorphicToProjectiveSpace`](https://github.com/homalg-project/ToricVarieties_project/blob/3b36aeff83427c9c889de42b7193168fd61793f1/ToricVarieties/gap/ProjectiveToricVarieties.gi#L127) which additionally checks `IsProjective( variety )` and otherwise differs from the condition altered in this PR by requiring that all weights are = 1.

PR #638, opened 2021-02-24
Set IsProjective for free objects in various categories

zickgraf:
This helps FreydCategory lift homomorphism structures for which the distinguished object is a free object in one of the modified categories.

### DIFF
--- a/CAP/examples/Homology.g
+++ b/CAP/examples/Homology.g
@@ -11,10 +11,8 @@ LoadPackage( "RingsForHomalg" );
 #! @Example
 ZZ := HomalgRingOfIntegersInSingular( );
 #! Z
-C1 := FreeLeftPresentation( 1, ZZ );
-#! <An object in Category of left presentations of Z>
-C2 := FreeLeftPresentation( 2, ZZ );
-#! <An object in Category of left presentations of Z>
+C1 := FreeLeftPresentation( 1, ZZ );;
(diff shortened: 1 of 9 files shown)


kamalsaleh:
It would be also helpful to analogously adjust GradedRows/Columns.

sebastianpos:
We haven't actually specified what we mean by a projective object in CAP.
This makes it tricky to mark an object in `CategoryOfRows` as projective.

For me, without any additional context, an object P in an arbitrary category C is called projective if Hom_C(P,-): C -> Set maps epis to surjective maps.

This definition makes 0 the only projective object in `CategoryOfRows` for almost all rings (the exception being at least skew-fields). E.g., for the ring of integers Z, the morphism given by multiplication by 2: Z -> Z is an epi in `CategoryOfRows`, but applying Hom(Z,-) is the same as applying the forgetful functor, which yields a non-surjective map.

The same applies to `CategoryOfColumns`.

For all the other categories in this PR, the change is fine.

zickgraf:
> This definition makes 0 the only projective object in `CategoryOfRows` for almost all rings (the exception being at least skew-fields)

Ah, thanks, I was thinking set-theoretically (i.e. about `CategoryOfRows` embedded into `LeftPresentations` as in my commutative algebra lecture). I have pushed a new version of the PR only setting `IsProjective` in the field/abelian case.

sebastianpos (approved):
Yes, this is a good solution, thank you.

PR #639, opened 2021-02-25
added dual_operation to MultiplyWithElementOfCommutativeRingForMorphisms

mohamed-barakat:


sebastianpos (approved):
Yes, that should be correct. Thanks.

PR #640, opened 2021-02-25
added Display method for objects in the opposite category

mohamed-barakat:


sebastianpos (commented):
Good idea. Could you also provide the corresponding function for morphisms?

mohamed-barakat:
Good idea, I will :)

sebastianpos (approved):
Thanks for the additional change.

PR #641, opened 2021-02-26
Fix DistinguishedObjectOfHomomorphismStructure in FreydCategory

zickgraf:
I think this was not noticed before because until now in all examples relevant for `FreydCategory` we had
* `underlying_category = range_category` or
* `RangeCategoryOfHomomorphismStructure( range_category ) = range_category`

and in both cases we have
```
DistinguishedObjectOfHomomorphismStructure( range_category ) = DistinguishedObjectOfHomomorphismStructure( underlying_category )
```

That's also the reason why I could not add a test for this (I have an example in my code but it does not make sense to add that code as a test).


sebastianpos (approved):
Thanks for spotting and fixing this bug!

PR #648, opened 2021-03-16
Preparations for things to come

zickgraf:
See commit messages for details.

Issue #586, opened 2020-09-21
Freyd: method installed for IsInjective matches more than one declaration


sebastianpos:
During my investigation of this problem, I produced the (so far) following minimal example:
```
LoadPackage( "GradedRingForHomalg" );
LoadPackage( "FinS" );
CategoryOfSkeletalFinSets( : overhead := false );
#I  method installed for IsInjective matches more than one declaration
```

If I omit the overhead option, then the info will not be printed. This hints at a problem of the overhead option.

sebastianpos:
This bug is very strange.

The new minimal example is:
```
LoadPackage( "homalg" );
LoadPackage( "FinS" );
CategoryOfSkeletalFinSets( : overhead := false );
```
and I suspected a conflict with `IsInjective` for `IsHomalgObject`.

The warning is triggered when `FinSets` calls `AddIsInjective` which in turn triggers in `CAP`
https://github.com/homalg-project/CAP_project/blob/6c95daf14d869d56589758313f8e088bb0162c00/CAP/gap/InstallAdds.gi#L516
the installation of `IsInjective` with the filter
`[ <Category "(SkeletalFinSetsObjectFilter and IsCapCategoryObject)"> ]`.

`IsHomalgObject` does not conflict with the filter above, at least calling `IsHomalgObject` on the initial object of the output category yields `false`. Nevertheless, the warning occurs at this point. So maybe I'm not aware of the correct conflict here.

If I omit the overhead option,  `FinSets` calls `AddIsInjective` which in turn triggers in `CAP`
https://github.com/homalg-project/CAP_project/blob/6c95daf14d869d56589758313f8e088bb0162c00/CAP/gap/InstallAdds.gi#L438
the installation of `IsInjective`. The filter at that point is the same as above.
If I manually interrupt at that point and do the installation manually with this filter, no warning is triggered.

I'm puzzled.

zickgraf:
I think the difference is that `install_method` in line 438 is `InstallMethodWithCache`, whereas line 516 simply uses `InstallMethod`.

zickgraf:
Further findings: The issue seems to be related to line: https://github.com/homalg-project/CAP_project/blob/6c95daf14d869d56589758313f8e088bb0162c00/CAP/gap/InstallAdds.gi#L430

If I exclude this line for `IsProjective`, the error does not happen anymore.

sebastianpos:
[...]

PR #649, opened 2021-03-16
Allow to pass the category as the first argument to any CAP operation

zickgraf:
~The first few commits are outsourced to #648.~

Technically, this PR is ready for review, but I would first like to discuss some choices I made / I want to make:

1. This PR is 100% backwards compatible for people using existing categories or writing their own categories.
2. It is not backwards compatible for people writing deriviations: They have to adapt their derivations to accept the category as the first argument (which should not be too hard, I think).
3. I have decided to add the category as the first argument "artificially" in the sense that it is prepended to the filter list in `InstallAdds.gi` and not in the method record. This might change, see below. (The motivation behind this approach was that I could change things step by step instead of having to change all pre functions, post functions, redirect functions, derivations etc. at once.)
4. In the current implementation, the category is passed on to all derivations and additionally to all primitively added functions of categories with `category!.category_as_first_argument` set to true.
5. None of this is documented yet. Of course I will write documentation as soon as everything is settled.
6. Question: Do we want to add the category argument to the method record (see point 3), or should we view this as an implementation detail for method selection? I vote for adding it to the method record as that fits the type theoretic approach, I think.
7. Question: Do we want to deprecate categories without `category!.category_as_first_argument` set to true (see point 4) in the future? I would vote yes, as adapting existing categories is not hard (just add `cat` as the first argument to all added functions).


zickgraf:
Sorry, I forgot one question:
8. There are some convenience methods not part of the method record (e.g. functorials without WithGiven, `AdditiveInverse`, `PreCompose` for lists, etc.), so they do not automatically get versions with the category as first argument (e.g., one currently cannot write `PreCompose( cat, [ mor1, mor2 ] )`). Question: Should the convenience methods also get versions with the category as first argument? I would vote no, with the following argument: Either you want convenience, then you probaly do not want to pass the category, or you want speed/correctness/compilation, then you should not use any convenience. This would also mean that convenience methods should not be used in derivations. However, some things like `PreCompose` for lists could also simply be added to the method record, e.g. as `PreComposeList`, if needed.

zickgraf:
Note: I'm still pushing to this branch because I have further changes based on the original PR. Once the discussion above is solved, I will of course split the PR up to make it reviewable.

sebastianpos:
I agree with your votings w.r.t. questions 6-8.

zickgraf:
This PR is now ready for review. Please review commits separately. See commit messages for details.

#### Structure of the PR

1. The first commit allows to call any CAP operation with the category as the first argument by artificially adding an appropriate filter in `InstallAdds.gi`.
2. This allows to remove method selection objects, `argument_list`, and many ops in subsequent commits.
3. The fifth commit actually adds the category as the first arguments to the method record, and passes it to all functions (pre, post, redirect, added). For backwards compatibility, primitively added functions are wrapped if category!.category_as_first_argument is not set to true.
4. The remaining commits should be self-explanatory by the commit message.

It make sense to always review the files on the following order: `AddFunctions.autodoc`, `MethodRecord.gi` (starting with `CAP_INTERNAL_ENHANCE_NAME_RECORD`), `InstallAdds.gi`, remaining files.

#### Documentation

I have only added a minimum of documentation as I expect things to change further. Since backwards compatibility is kept in most cases, this should not be a problem for ordinary users. I will open an issue for discussing some points regarding the documentation.

#### Notes

* The creation of the convenience methods without the category is very crude at the moment. I plan to update and generalize this in a future PR.
* Most derived methods still call operations without the category as the first argument. Changing this will be part of a future PR.
* I have not touched the io_types, so CategoryConstructor does not need adjustments regarding that point. @mohamed-barakat: I am open for discussions if the io_types should include the category or not.
* I will not be available for the next two weeks, so please do not merge until until then (or don't blame me for broken things :D ).

#### Compatibility

Packages have to be updated if they are
* using automatic post functions, redirect functions, or WithGiven derivations (suitable warnings will be raised when loading the package),
* adding derivations for methods in one of the updated method records (i.e. CAP, Freyd, MonoidalCategories; errors about a wrong number of arguments will be raised when loading the package),
* using prepare functions (I only know of SkeletalFinSets using this),
* relying on `argument_list` in a method record (suitable warnings will be raised when loading the package),
* loading FreydCategoriesForCAP in their tests, because #586 is finally fixed :-)

I will create PRs for FinSetsForCAP, Toposes, HigherHomologicalAlgebra and CategoryConstructor.

sebastianpos (approved):

zickgraf:
I'm available again, so feel free to merge this PR any time you like :-)

sebastianpos:
Okay, so as discussed 2 weeks ago, I'll merge this PR now :)

PR #650, opened 2021-03-17
Partially revert "Fix some pre functions"

zickgraf:
This partially reverts commit ef4ca57ea17a9c15b5251cb088916eec266020dc.

The automatic derivation of pre functions for WithGiven pairs does not take
method selection objects into account and thus generates pre functions with
the wrong number of arguments in these two cases. The old version works
because the WithGiven object is used as the method selection object in
(Co)Equalizer.pre_function, which breaks semantics but technically works.

I have not noticed this problem before because I only tested this as part
of #649 where method selection objects are gone and thus the automatic
derivation of pre functions for WithGiven pairs is actually correct.


sebastianpos (approved):
Okay, so we see this as semantically wrong, technically okay, and superfluous in the very near future.

PR #653, opened 2021-04-13
Miscellaneous changes

zickgraf:

@sebastianpos There is nothing complex or controversial in this PR (except the deprecations which we have already talked about) and nothing should affect existing code (expect of course if one runs into a deprecation or a new plausibility check) -> I will use your permission and merge this PR without review later today (or should I keep a longer grace period?), except if you complain of course.


sebastianpos:
Please, go ahead :)

PR #654, opened 2021-04-13
Deprecate setting the installation name manually

zickgraf:

### Rationale

The installation name is a GAP-specific implementation detail which can be automatically chosen by the following rule: if the function name is an operation, use this, else append "Op".

### Structure of the PR

1. The first commit is preparatory.
2. The succeeding commits align several exceptions which do not follow the rule mentioned above.
3. The second to last commit deprecates setting the installation name manually to a value different than the one given by the rule.
4. The last commit finally removes `installation_name` from the method records (which also implies adaptions to the limit validation code).

### Backwards compatibility

Packages have to be updated if they
1. use installation names different than the one given by the rule (a suitable warning is raised),
2. explicitly call `MorphismBetweenDirectSumsOp`, `PushoutOp` or `CoproductOp` (I assume that this should be rare),
3. look for `InverseImmutable` in the method record (this is replaced by `InverseForMorphisms`, see the second commit).

@sebastianpos: The commits are more or less straight forward, but since they are not backwards compatible I do not want to proceed without your OK.


sebastianpos (approved):
You get my OK. Your rule for choosing a function name is very reasonable.

PR #667, opened 2021-05-16
Support empty limits in FreydCategory

zickgraf:
@sebastianpos That's the commit where I think a review would be good because mathematical nonsense could happen. The relevant part is just two lines :-)


sebastianpos (approved):
Yes, that looks mathematically sound.

PR #674, opened 2021-05-27
Make OppositeCategory compilable

zickgraf:
Major changes:

1. Pass the category as first argument to the dual preprocessor func.
2. Use `ObjectConstructor` and `MorphismConstructor`.
3. Do not rely on immediate methods. Instead, manually call `SetOpposite` where appropriate.
4. Create the added functions by building a string (see https://github.com/homalg-project/CAP_project/commit/6398eb826bfe77b9efc87cdd177caa6ac7c200df#diff-03476bc533805918f3fbc18e61f7d3e49bd101886a4f2ca292f6adf4d7b8a161R202 ff.) and applying `EvalString`. This makes the added functions readable and compilable. For example, the function created for `KernelLift` reads:
```
            function ( cat, alpha, T, tau )
              local dual_preprocessor_func, prep_arg, result, dual_postprocessor_func;
                
                
                
                result := KernelLift( Opposite( cat ), Opposite( alpha ), Opposite( T ), Opposite( tau ) );
                
                
                
                return MorphismConstructor( cat, ObjectConstructor( cat, Range( result ) ), result, ObjectConstructor( cat, Source( result ) ) );
                
            end
```
(the weird spacing and unused local variables are expected).

Note: The precompiled code in `OppositeOfMatrixCategoryPrecompiled.gi` is not meant to be used as is, but simply there for regression testing. In an upcoming PR, I will use `WrapperCategories` to generate usable precompiled code of opposite categories, e.g. CategoryOfColumns from CategoryOfRows.

@sebastianpos: Because the code is very generic, I don't think a review is required, but I would like to have your OK for the ideas explained above.

### DIFF
--- a/CAP/gap/InstallAdds.gi
+++ b/CAP/gap/InstallAdds.gi
@@ -400,6 +400,10 @@ InstallGlobalFunction( CapInternalInstallAdd,
             output_sanity_check_function := function( result )
                 CAP_INTERNAL_ASSERT_IS_MORPHISM_OF_CATEGORY( result, false, output_human_readable_identifier_getter );
             end;
+        elif record.return_type = "list_of_morphisms" then
+            output_sanity_check_function := function( result )
+                CAP_INTERNAL_ASSERT_IS_LIST_OF_MORPHISMS_OF_CATEGORY( result, category, output_human_readable_identifier_getter );
+            end;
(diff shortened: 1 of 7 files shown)


sebastianpos:
You have my OK :)

Issue #672, opened 2021-05-27
dual_postprocessor_func of BasisOfExternalHom


zickgraf:
The `dual_postprocessor_func` of `BasisOfExternalHom` is currently set to `IdFunc`. According to the documentation, the result of `BasisOfExternalHom` should be a list of morphisms in the same category as the input. Thus, I think choosing `IdFunc` for `dual_postprocessor_func` is wrong and the correct value should be `CAP_INTERNAL_OPPOSITE_RECURSIVE` (which is the default). Is this correct?

cc @mohamed-barakat who is the author of that code according to git-blame.

mohamed-barakat:
@kamalsaleh: Could you please have a look at it?

kamalsaleh:
Fabian is right, it is a bug. The above PR should fix it.

mohamed-barakat:
Thanks @zickgraf for reporting and @kamalsaleh for fixing 😊

PR #675, opened 2021-05-27
Fix bug in dual_postprocessor_func for BasisOfExternalHom record

kamalsaleh:
(Reported by Fabian in https://github.com/homalg-project/CAP_project/issues/672)


zickgraf (approved):

zickgraf:
Thanks for the PR! I will merge this without review from @sebastianpos because the change is straightfoward.

PR #678, opened 2021-06-01
Add Lift/Colift/SolveLinearSystemInAbCategoryOrFail

### LINKED ISSUE #673: Proposal: Disallow `fail` for Lift, Colift and SolveLinearSystemInAbCategory
As far as I am aware, `Lift`, `Colift` and `SolveLinearSystemInAbCategory` are the only remaining CAP operations, where `if ... else ...` constructions occur in the implementation, due to possibly returning fail. This currently is a major problem for `CompilerForCAP`, because in towers of categories this leads to nested `if ... else ...` statements which are hard to simplify in a generic way.

According to @mohamed-barakat, all of his applications fall into one of the following two cases: Either one knows a priori (mathematically) that the lift etc. exists, or one wants to decide if the lift exists (without actually computing a lift). The second case is already covered by `IsLiftable` etc.

Thus, I propose we tighten the specification of `Lift` etc. to assume that the lift exists. Of course it would be nice to do this in a backwards compatible manner, but that would mean we need a new operation `LiftWithGivenExistence` which would be weird when compared to `LiftAlongMonomorphism`, for example.

Should the need arise that one wants to decide if a lift exists and compute one at the same time, we could reintroduce the existing behavior via a operation `LiftOrFail` which would be ignored by the compiler.

zickgraf:
Changes contained in the PR:
1. ~The first commit is preparatory (we use `DecideZeroRows/Columns` in the second commit which would cause errors in the compiler).~ Merged as #681.
2. Lift/Colift/SolveLinearSystemInAbCategory now assume that a lift/colift/solution exists.
3. The operation `MereExistenceOfWitnessPairForBeingCongruentToZero` is added for Adelman categories.
4. I have looked at all places where Lift/Colift/SolveLinearSystemInAbCategory are called and have adjusted them.
5. I have not changed all existing implementations in categories (i.e. `AddLift` etc.), because these are still correct: If the input is valid according to the new specification, these implementations do not return `fail` in the old specification and thus everything is fine. However, I have adjusted some implementations where the tests need `IsLiftable` etc. (which cannot be derived from `Lift` etc. anymore and thus have to be added primitively).

Fixes #673.

@sebastianpos I think it would be a good idea if you could review the changes touching `FreydCategoriesForCAP`. All the "new" code is copied from somewhere else (e.g. the code in `IsLiftable` in `GradedCategoryOfRows` is copied from `CategoryOfRows`), so there is no need for an in-depth review, but I think it would be good if you could check that I have copied things from the correct place (i.e. it's easy to swap `Lift` and `Colift` by accident :D).


sebastianpos (approved):
I checked the changes to the Freyd package. They all look fine. As we discussed yesterday, I'll merge this PR.

Issue #403, opened 2019-10-14
Split tests for Freyd category (w.r.t. QPA 2)


sebastianpos:
See discussion in #401.

zickgraf:
I have just talked to Max and he has pointed me to a new syntax in GAP 4.11 which allows to insert `#@ if ...` in tests to conditionally execute a test. This syntax is backwards compatible in the sense that that GAP versions prior to 4.11 simply ignore it. So maybe we can proceed with this issue once GAP 4.11 is released.

sebastianpos:
This sounds promising. Okay.

zickgraf:
Now that GAP 4.11 is released, implementing this should be easy:
1. The tests needing QPA2 can be enclosed in `#@ if IsPackageMarkedForLoading ...`, see https://github.com/zickgraf/FinGSetsForCAP/blob/master/examples/ReconstructGroup.g#L10 and the closing `fi` in line 18.
2. The tests can be added to the CI analogous to #376.

There is one caveat: GAP 4.10 and earlier simply ignore `#@ if`, so if someone uses GAP <= 4.10 without having QPA2 installed there will be errors.

PR #682, opened 2021-06-07
Add CategoryOfColumnsAsOppositeOfCategoryOfRows

zickgraf:
This PR includes #678.

The main commit is the third to last commit: it adds the operation `CategoryOfColumnsAsOppositeOfCategoryOfRows` which does what its name says. By making use of ObjectConstructor, ObjectDatum, MorphismConstructor and MorphismDatum (see below), this is a drop-in replacement for `CategoryOfColumns` (with some additional attributes set, e.g. `Opposite` points to a category of rows). We still have to manually add some operations which are not covered by `OppositeCategory`, e.g. the tensor structure. The last commit compiles both versions of `CategoryOfColumns`. If one compares the output, one can see that the differences are negligible.

For all of this we need some preparation, see the commit messages for details. In particular, `OppositeCategory` now uses the information stored in the `io_type` to get source and range of a morphism directly from the input. For this, I add two new method record entries `output_source_getter` and `output_range_getter` which contain the result of `CAP_INTERNAL_GET_CORRESPONDING_OUTPUT_OBJECTS` (and should ultimately replace the `io_type` and `CAP_INTERNAL_GET_CORRESPONDING_OUTPUT_OBJECTS`).

Regarding the use of ObjectConstructor, ObjectDatum, MorphismConstructor and MorphismDatum: If the implementation of a category only uses these operations to construct objects and morphism and to access their data, one can now change the data structure of a category visible to the outside world ad libitum. For an example, see `CategoryOfColumnsAsOppositeOfCategoryOfRows.gi`: when creating `Opposite( CategoryOfRows )` without modifications, an object in this category is constructed from a CAP object in `CategoryOfRows` by wrapping it in an GAP object. Now, instead of wrapping such a CAP object we can simply wrap its rank (this is done in `ObjectConstructor`), and when the code tries to access the object datum of an object created in this way, we create a CAP object in `CategoryOfRows` again (this is done in `ObjectDatum`). By doing this, the user only sees GAP objects with the attribute `RankOfObject`, just as he/she would expect when working with a category of columns. I think this concept is very powerful and flexible. The only thing to keep in mind is what I have written above: this only works if the implementation of a category only uses the mentioned operations to construct objects and morphism and to access their data. In the case of the opposite category, this also includes `dual_pre/postprocessor_func`. But in general, this restriction could also be circumvented by wrapping the category using https://github.com/homalg-project/WrapperCategories (as soon as I have created a PR there). (Sorry if this paragraph is difficult to understand, I would need a blackboard to visualize this concept :D )


zickgraf:
The code coverage is dropping so much because the tests of `CompilerForCAP` now load `FreydCategoriesForCAP`, but the Freyd tests are not executed, so more or less everything is uncovered. We could live with this (because effectively nothing has changed) or we could fix #403 (which should at least improve things, although I don't know how much).

sebastianpos (approved):
As we discussed yesterday: since #678 is merged, I'll merge this PR.

PR #685, opened 2021-07-14
print missing newline in Display method

mohamed-barakat:
In Jupyter notebooks the missing newline causes the output to be delayed to the next cell


zickgraf:
The Codecov report above is broken. Actually, the code coverage does not change (as expected): https://codecov.io/gh/homalg-project/CAP_project/commit/01e7163fc5940a962f4cdec432260e466fda5736

zickgraf (approved):

PR #686, opened 2021-07-20
Add Object/MorphismConstructor/Datum to LinearAlgebraForCAP and CategoryOfRows/Columns

zickgraf:
The changes in this PR are straightforward, so I will merge this PR without additional review if and when the CI succeeds.


zickgraf:
With this PR, `CategoryOfColumnsAsOppositeOfCategoryOfRows` is not a drop-in replacement for `CategoryOfColumns` anymore, because `CategoryOfColumnsObject` now calls `ObjectConstructor`, and the constructors of  `CategoryOfColumnsAsOppositeOfCategoryOfRows` and `CategoryOfColumns` differ. I will have to think about a clean solution for this problem.

PR #687, opened 2021-07-21
Use Object/MorphismConstructor to simplify hom structure in QuiverRows

zickgraf:
At @kamalsaleh's request we keep MatrixCategory as the default Hom-structure of Quiver Rows.

----

Old description: Always use CategoryOfRows as range of the Hom-structure of QuiverRows

and allow to override it using the option `range_of_HomStructure`.

As a preparation, I introduce InterpretMorphismAsMorphismFromDistinguishedObjectToHomomorphismStructure**WithGivenObjects**.

@sebastianpos: I think the changes are straightforward but I would wait for a general "okay" from you :-)

### DIFF
--- a/FreydCategoriesForCAP/gap/QuiverRows.gi
+++ b/FreydCategoriesForCAP/gap/QuiverRows.gi
@@ -469,8 +469,8 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_QUIVER_ROWS,
   function( category, over_Z )
     local algebra, quiver, zero, IDENTITY_MATRIX_QUIVER_ROWS, ZERO_MATRIX_QUIVER_ROWS,
           vertices, basis, basis_paths_by_vertex_index, path, MATRIX_FOR_ALGEBROID_HOMSTRUCTURE, hom_structure_algebroid,
-          object_constructor, ring, morphism_constructor, hom_structure_range_category, hom_structure_on_morphisms_for_pure_components,
-          distinguished_object, representative_func;
+          ring, default_range_of_HomStructure, hom_structure_range_category, hom_structure_on_morphisms_for_pure_components,
+          representative_func;
     
     algebra := UnderlyingQuiverAlgebra( category );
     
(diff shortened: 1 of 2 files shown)


FreydCategoriesForCAP/gap/QuiverRows.gi
```diff
@@ -1335,7 +1291,15 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_QUIVER_ROWS,
             
         od;
         
-        return morphism_constructor( distinguished_object, row, object_constructor( Size( row ) ) );
+        if IsEmpty( row ) then
+            
+            return ZeroMorphism( hom_structure_range_category, source, range );
```
sebastianpos, line 1296: Is this line the reason why you needed to introduce the new WithGiven version of this function?

zickgraf: Technically, introducing the new WithGiven version was not a "must", but yes, that's where it comes in handy. I already thought about this yesterday when writing the following lines of code: https://github.com/homalg-project/Algebroids/blob/6eb2a9884facf904e3b76d5f07123f320c7174f5/gap/Algebroids.gi#L731-L732 And I think it's natural to have the WithGiven version, or is there anything that distinguishes this case from e.g. `HomomorphismStructureOnMorphismsWithGivenObjects`?

zickgraf:
(I just pushed a quick fixup because I forgot to call the operations in the derivations with the category as the first argument, argh....)

kamalsaleh:
I like this step toward making `CategoryOfRows(k)` the constructor of `k-vec` instead of `MatrixCategory(k)`.

However, I have one concern regarding this: Previously, every field `k` has its `k-vec` category as an attribute which I can use everywhere in Algebroids, functor categories, quiver representations, graded rows, quiver rows, etc. Currently, `CategoryOfRows(k)` is not an attribute, hence leads to multiple `k-vec`'s for the same field `k`.

Sometimes, lifting Hom-Structure  to a higher category requires applying a constructor on the range of that Hom-Structure (e.g., complexes categories) and having different `k-vec`'s would mean the constructor must be applied on all of them.

Here is my suggestion: Introduce an attribute: `CategoryOfRowsAttr(k)` or `Rows(k)`(https://github.com/homalg-project/CAP_project/pull/689); and use that whenever a generic constructor needs `k-vec`. The user can still use the option: `range_of_HomStructure` if different instances for `k-vec` are needed.

If it was up to me, I would go further and make `CategoryOfRows(k)` an attribute and introduce another constructor to create different instances. There is definitely a reason for the current behavior, but I can see/remeber it.

zickgraf:
> Here is my suggestion: Introduce an attribute: `CategoryOfRowsAttr(k)` or `Rows(k)`(#689); and use that whenever a generic constructor needs `k-vec`. The user can still use the option: `range_of_HomStructure` if different instances for `k-vec` are needed.
> 
> If it was up to me, I would go further and make `CategoryOfRows(k)` an attribute and introduce another constructor to create different instances.

Both suggestions sound find to me! Only two things we should be aware of:
1. Attributes do not respect options, so calling `CategoryOfRowsAttr(k)` and then `CategoryOfRowsAttr(k : enable_compilation := true )` returns the same instance (with compilation disabled). This might be confusing for users.
2. Imagine a user is using `CategoryOfRowsAttr(k)` and `QuiverRows` of a quiver over the same `k` at the same time. Then turning caching/logic on and off for `CategoryOfRowsAttr(k)` also affects computations in `QuiverRows`, which again might be very confusing and hard to debug if the user is unaware of the internals of `QuiverRows`.

Usually, I would say that we should avoid these issues, following the "principle of least surprise". However, a heavy GAP user might just as well expect all of this and would be surprised if something like this was not an attribute, so... :D

zickgraf:
The scope of this PR has changed: At @kamalsaleh's request we keep MatrixCategory as the default Hom-structure of Quiver Rows. So this PR only simplifies some things -> I will merge it if and when the CI succeeds without additional review.

PR #689, opened 2021-07-21
d&i CategoryOf(Rows/Columns)Attr as option to avoid constructing CategoryOfRows/Columns multiple times

kamalsaleh:


kamalsaleh:
No one, including me, needs this anymore. Hence, I will close it.

PR #696, opened 2021-07-30
Fix bug in caching mechanism and add Object/MorphismConstructor to non-cached operations

### LINKED ISSUE #695: The default caching for Object/MorphismConstructors slows computation
Currently, the `ObjectConstructor` and `MorphismConstructor` in a category C are basic operations, so they inherit the same caching of C. In  `LinearAlgebraForCAP` all basic methods create their output using `ObjectConstructor` and `MorphismConstructor` which are now also cached. This causes a major performance regression.

In some categories it might be helpful in `Algebroids` but in `LinearAlgebraForCAP` not that much :/ 

```
LoadPackage( "DerivedCategories" );
EnhanceAllPackages();
Q := HomalgFieldOfRationals( );
A := RandomQuiverAlgebraWhoseIndecProjectiveRepsAreStrongExceptionalCollection( Q, 5, 10, 5 );

Q_mat := MatrixCategory( Q );
A_oid := Algebroid( A : range_of_HomStructure := Q_mat );
K_A_oid_plus := HomotopyCategory( AdditiveClosure( A_oid ) );

o := RandomObject( K_A_oid_plus, [-4,4,3] );

# (1)
DeactivateCachingObject( Q_mat!.caches.MorphismConstructor );
B := CallFuncListWithTime( BasisOfExternalHom, [o,o] );;
# 20.967779 Secs.
Size(B);
# 16
Q_mat!.caches.MorphismConstructor;
# <A cache with 0 hits and 0 misses>


# (2)
B := CallFuncListWithTime( BasisOfExternalHom, [o,o] );;
# 39.058862 Secs.
Size(B);
# 16
Q_mat!.caches.MorphismConstructor;
# <A cache with 95 hits and 10676 misses>
```

zickgraf:
Previously, calling `SetCachingOfCategoryWeak/Crisp` would enable the caches of IsWellDefinedFor* and Random* due to those not being included in the list in `SetCachingOfCategory`.

Fixes #695.

@kamalsaleh I have noticed that one can actually override the caching even if it is set to "never". In fact, there was a bug based on this, see above. So now I have chosen option 3 in #695 with an explicit remark in the documentation. Maybe you can report if this indeed resolves your issue :-)

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -21,7 +21,7 @@ Version := Maximum( [
   ## this line prevents merge conflicts
   "2020.04-16", ## Sepp's version
   ## this line prevents merge conflicts
-  "2021.07-01", ## Fabian's version
+  "2021.07-02", ## Fabian's version
   ## this line prevents merge conflicts
   "2021.05-04", ## Kamal's version
 ] ),
--- a/CAP/gap/CAP.gi
+++ b/CAP/gap/CAP.gi
@@ -27,9 +27,40 @@ InstallTrueMethod( IsPreAbelianCategory, IsAbelianCategory );
 
 ##
 InstallValue( CAP_INTERNAL,
-              rec(
-                   name_counter := 0,
-                   default_cache_type := "weak",
+                rec(
+                    name_counter := 0,
+                    default_cache_type := "weak",
+                    operation_names_with_cache_disabled_by_default := [
+                        # the following operations are needed for comparison in caches
+                        "IsEqualForObjects",
+                        "IsEqualForMorphisms",
+                        "IsEqualForMorphismsOnMor",
+                        "IsEqualForCacheForObjects",
+                        "IsEqualForCacheForMorphisms",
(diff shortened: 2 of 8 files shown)


kamalsaleh:
Thanks for the quick PR. I think it is a sensible default behavior. However, in `LinearAlgebraForCAP` caching objects significantly improves the performance. See my comment on the issue.

zickgraf:
Ah, sorry, this PR was "mit heißer Nadel gestrickt" (a German idiom :D ). Can you try again with the newly added commits?

kamalsaleh:
Now the default behavior is much better:
```
# (5)
B := CallFuncListWithTime( BasisOfExternalHom, [o,o] );;
# 11.771241 Secs.
Size(B);
# 16
Q_mat!.caches.ObjectConstructor;
<A cache with 15849 hits and 31 misses>
```

zickgraf:
Perfect! Then I will merge this PR without additional review, because the new state is certainly better than the old state.

PR #700, opened 2021-08-25
Make `CategoryOfColumns` return a precompiled version of `CategoryOfColumnsAsOppositeOfCategoryOfRows`

zickgraf:
The remaining difference between the two can be seen by diffing
https://github.com/homalg-project/CAP_project/blob/83cccba77af7f754c36289312dbaeb9e6008d171/FreydCategoriesForCAP/gap/precompiled_categories/CategoryOfColumnsOfFieldPrecompiled.gi
and
https://github.com/homalg-project/CAP_project/blob/83cccba77af7f754c36289312dbaeb9e6008d171/FreydCategoriesForCAP/gap/precompiled_categories/OppositeOfCategoryOfRowsOfFieldPrecompiled.gi

There are only two actual differences:
1. the order of the `BasisOfExternalHom` is different,
2. Object/MorphismConstructor/Datum are different, as expected.

Constructing CategoryOfColumns via `CategoryOfColumnsAsOppositeOfCategoryOfRows` is slower by about a factor of 2, because two categories have to be created. However, the algorithms have the same speed due to compilation.

One can get an uncompiled version by passing the option `no_precompiled_code` to `CategoryOfColumns`.

In conjunction with this, I also declare the compiler as "ready for production" use. I have not found any instances where the compiler has produced wrong code in the last weeks and months, and with people actually using compiled code after this PR, I am quite confident that any problems would show quickly.

@sebastianpos: I will await your okay. No general review is needed, but maybe you can look at `CategoryOfColumns.gi` and say if it matches the things you had in mind when we talked about this during our last meeting.


mohamed-barakat:
The merged commits produced conflicts in this PR.

zickgraf:
> The merged commits produced conflicts in this PR.

Yes, that's expected due to version bumps and changes to the method record. Since both can occur again any time, I will rebase this right before the merge.

sebastianpos (approved):

PR #702, opened 2021-08-29
added new *Constructor/*Datum operations to no_install_list

mohamed-barakat:
comply with CAP v2021.07-02


zickgraf (dismissed):
I cannot review this from a greater point of view, but I would agree, that this should not break things.

PR #703, opened 2021-08-30
Turn WithoutGiven monoidal operations into proper CAP operations

### LINKED ISSUE #694: Turning `Braiding` et al. into proper CAP operations
Compilation of monoidal categories faces the following problem: For most of the operations in MonoidalCategories, only the WithGiven variant is a proper CAP operation, while the without given variant is manually installed as a convenience operation. These convenience operations have to be registered to the compiler via `InstallMethodForCompilerForCAP`, so that the compiler knows these operations when (pre-)compiling without JIT arguments. In this case, the compiler does a manual method selection solely based on the number of arguments. Unfortunately, there are cases, where this is not sufficient, for example for the braiding:

https://github.com/homalg-project/CAP_project/blob/a57615fe4cd202894542f54ed7346c80990510b5/MonoidalCategories/gap/BraidedMonoidalCategories.gi#L18-L19

Here, a special method for skeletal categories is installed, overriding the generic one with the same number of arguments.

If `Braiding` would be a proper CAP operation, this method would be a derivation with a special category filter. When the compiler does method selection for CAP operations, it has the category at hand and looks at the installations of the operation in that category. Thus, it would only see this special method if it was actually installed by the derivation mechanism, i.e. passing the category filter.

Until some days ago I thought turning `Braiding` et al. into proper CAP operations would be cumbersome, because I would have to write suitable post functions, but now that I know that the purpose of post functions is limited to performance reasons. Since post functions are ignored anyway when using `overhead := false` or the compiler, I would suggest that I simply turn `Braiding` et al. into CAP operations without writing suitable post functions.

@sebastianpos, @kamalsaleh, @mohamed-barakat: Could you give a quick thumbs-up or thumbs-down or "I'm not sure and would have to think about this" regarding this idea? I don't want to start this effort and realize at the end that this kills someone's performance :D

zickgraf:
Also includes the following changes:

1. Many operations have special installations for the skeletal case. These are kept as derivations (with a suitable CategoryFilter).
2. Fix a bug with the arguments of `MorphismFromCoBidual` and `MorphismFromCoBidualWithGivenCoBidual`.
3. Change the arguments of `TensorProductInternalHomCompatibilityMorphism(WithGiven...)`: For compatibility with the mechanism which automatically installs WithGiven derivations, source and range must be the second and last argument of a operation (recall, the first argument is the category). To comply with GAP's limit on the number of arguments of an operation, the remaining arguments are now packed in a list.
4. The documentation for Add functions is now generated automatically (see the ".autogen.gd" files). One can still change the content of these files for debugging purposes, but warnings like the following will be printed:
```
WARNING: The file AdditiveMonoidalCategories.autogen.gd in package MonoidalCategories differs from the automatically generated one. You can view the automatically generated file at the following path: /tmp/gaptempdirAF6Aw1/AdditiveMonoidalCategories.autogen.gd
```

@mohamed-barakat No in-depth review is necessary, but it would be nice of you could review the changes conceptionally.

Edit: Fixes #694.


mohamed-barakat (approved):
Thanks for the offline discussion. All dev branches in my packages now work as well. I approve and will merge.

PR #723, opened 2021-10-13
Fix the finalize mess

zickgraf:
1. Turn `IsFinalized` into a mutable attribute (properties cannot be
   mutable).
2. Do not finalize a category automatically when asking `IsFinalized`.
3. Set `IsFinalized` to `false` by default and to `true` when finalizing.
4. With 2. and 3. it's possible to get rid of `HasIsFinalized`.
5. Respect the option `FinalizeCategory` in `Finalize`.
6. With 5. we can get rid of the checks regarding `FinalizeCategory`
   everywhere else in the code.
7. Add a warning when calling an operation in an unfinalized category.

@sebastianpos Do you agree with the points above? I do not think a review of the code is necessary because any problem with this should show up very clearly in tests.

### DIFF
--- a/ActionsForCAP/gap/ActionsCategory.gi
+++ b/ActionsForCAP/gap/ActionsCategory.gi
@@ -36,7 +36,7 @@ InstallMethod( LeftActionsCategory,
   function( acting_object, name, context_filter_list )
     local underlying_monoidal_category, preconditions, category_weight_list, i,
           structure_record, object_constructor, morphism_constructor, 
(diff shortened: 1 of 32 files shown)


zickgraf:
Good point! It worked implicitly: the code actually ran a second time when calling `Finalize` two times, but it would usually simply detect that no new final derivations could be installed. However, this could have lead to errors when installing new final derivations before calling `Finalize` a second time.

I have now added an explicit check, so now everything should be fine and I will merge the PR if and when the CI succeeds.

PR #725, opened 2021-10-15
Miscellaneous changes around code coverage

zickgraf:


zickgraf:
The drop in code coverage is expected because GAP master counts continued lines in a different way than GAP stable.

PR #737, opened 2021-11-04
Compile all operations of MatrixCategory and use compiled code by default

zickgraf:
The first two commits are preparatory.

@sebastianpos This PR does not require a review but I would like your okay regarding the following points/decisions:

1. The first commit deprecates `AddWithGivenDerivationPairToCAP` (except a single case which is the only case actually used, namely passing a CAP operation and two functions) and `AddDerivationPair`. The reason: `AddDerivationPair` searches for CAP operations in both functions and simply takes the union of the results. This introduces unneeded dependencies to such derivation pairs and makes them seem more expensive than they actually are. Instead of fixing this, I would like to simply get rid of `AddDerivationPair` because it's not really used except for the special case mentioned above, which I now handle explicitly. Are you okay with the deprecation?
2. The second commit makes all functorials with-given pairs. In particular, it introduces `ZeroObjectFunctorialWithGivenZeroObject` (and the same for terminal and initial objects). For consistency with the other functorials, this operations gets the zero object twice, once as the source and once as the range. This might be unexpected for users, but since the WithGiven versions are mostly used in automated contexts (e.g. CategoryConstructor) and performance should not play a role here, I have decided to keep things simple and consistent on a technical level.
3. The second to last commit makes `MatrixCategory` use the compiled code by default.


zickgraf:
Ad 3.: Note that using the compiled code leads to less logic being applied, because the logic does not "push trough" derivations anymore if they are compiled. See the second to last commit for two examples of this.

sebastianpos:
I'm okay with 1). If I remember correctly, you already simplified the derivation of With-Given-operations earlier, so I wouldn't expect inconsistencies.

I'm also okay with 2). No user would use that operation within a computing session anyway, so it's okay to be technical.

Last, I'm fine with 3). It complies with my philosophy of the MatrixCategory being a workhorse category. In particular, no harm is done if no logic is applied.

zickgraf:
Perfect, thanks for your comments! Then I will merge once the CI is green after the last rebase.

PR #741, opened 2021-11-09
Two small changes in FreydCategory for better compilability

zickgraf:
@sebastianpos Could you review these two small changes? They should be straightforward, but I am very careful when touching "mathematics" in FreydCategory :D

### DIFF
--- a/FreydCategoriesForCAP/gap/FreydCategory.gi
+++ b/FreydCategoriesForCAP/gap/FreydCategory.gi
@@ -1390,9 +1390,13 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_FREYD_CATEGORY,
             function( cat, source, a, b, c, range )
                 local mor;
                 
-                mor := AssociatorLeftToRight( Range( ObjectDatum( cat, a ) ),
-                                              Range( ObjectDatum( cat, b ) ),
-                                              Range( ObjectDatum( cat, c ) ) );
+                mor := AssociatorLeftToRightWithGivenTensorProducts( underlying_category,
+                    Range( ObjectDatum( cat, source ) ),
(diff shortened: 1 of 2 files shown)


FreydCategoriesForCAP/gap/FreydCategory.gi
```diff
@@ -1344,7 +1344,7 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_FREYD_CATEGORY,
                 sink := [ mor1, mor2 ];
                 
                 # compute the universal morphism
-                uni := UniversalMorphismFromDirectSum( underlying_category, diagram, sink );
+                uni := UniversalMorphismFromDirectSum( underlying_category, diagram, range, sink );
```
sebastianpos, line 1347: If the 3rd argument of `UniversalMorphismFromDirectSum` now is the range of the result, then this line is correct.

Is this version of `UniversalMorphismFromDirectSum` documented?

sebastianpos (dismissed):
Up to my comment this looks correct.

FreydCategoriesForCAP/gap/FreydCategory.gi
zickgraf: > If the 3rd argument of `UniversalMorphismFromDirectSum` now is the range of the result, then this line is correct.

It is: it's the test object, i.e. common range of the morphisms in `sink`, and thus the range of the universal morphism.

> Is this version of `UniversalMorphismFromDirectSum` documented?

Yes, it is. In case you looked for a four argument version and could not find it: CAP operations are not documented with the category as the first argument, so in this case one has to look for the three argument version.

zickgraf:
> Up to my comment this looks correct.

Thanks for the review! I will rebase and merge.

PR #744, opened 2021-11-10
Introduce Left/RightPresentationsAsFreydCategoryOfCategoryOfRows/Columns

zickgraf:
Use those to generate (i.e. precompile) some operations for `Left/RightPresentations`, so we can drop the code in `ModulePresentationsForCAP`, see `ADD_FUNCTIONS_FOR_LEFT_PRESENTATION`. The optional dependency on `FreydCategoriesForCAP` is only needed for precompilation.

Note: I know that `ModulePresentationsForCAP` should be replaced by `FreydCategory` anyway. With this PR we can see and check what is actually missing for achieving this.

@sebastianpos Again, this does not need a review, but your general approval of the overall idea :-)

### DIFF
--- a/CompilerForCAP/examples/PrecompileModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.g
+++ b/CompilerForCAP/examples/PrecompileModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.g
@@ -1,0 +1,151 @@
+#! @Chapter Examples and tests
+
+#! @Section Tests
+
+#! @Example
+
+LoadPackage( "ModulePresentationsForCAP", false );
+#! true
+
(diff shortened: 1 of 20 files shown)


zickgraf:
Since I have merged compiled code multiple times in the last weeks, I will merge this PR without additional approval.

The first commit of this PR now renames the morphism attribute in FreydCategory from `MorphismDatum` to `UnderlyingMorphism`, because otherwise the compiler confuses the morphism attribute with the CAP operation of the same name (strictly speaking, the compiler can handle this, but the string-based consistency checks cannot, which leads to warnings). @sebastianpos If you do not like this change, it's easy to revert: We only have to change very few lines, because the CAP operation `MorphismDatum` is based in the morphism attribute getter anyway, so this is fully backwards compatible for the existing implementation.

PR #747, opened 2021-11-11
Fix regression of 0df6824a in GradedModulePresentationsByFreyd

### LINKED ISSUE #746: Many failure for FPGradedModules in ToricVarieties_project
Currently, there are massive failures in the tests of at least four packages in `ToricVarieties_project`. This appears to be linked to a change which leads to f.p. graded modules no longer satisfying the filter `IsFpGradedLeftOrRightModulesObject`. Here is an example:

> Q := HomalgFieldOfRationalsInSingular();
> S := GradedRing( Q * "x_1, x_2" );
> SetWeightsOfIndeterminates( S, [[1],[1]] );
> Module := AsFreydCategoryObject( GradedRow( [[[1],1]], S ) );
> IsFpGradedLeftOrRightModulesObject( Module );

The last line should return `true` but appears to return `false` since at most 48 hours. This seems to be the root cause for the current massive failures.

Why is this? Would it be possible to revert this change?

I have temporarily disabled the corresponding tests in https://github.com/homalg-project/ToricVarieties_project/pull/167/commits/a131a178c4c5e06a5c3d76ee5ce23e85ca75ab52.

zickgraf:
Do not hijack FreydCategoryObject/Morphism.

Fixes #746.

@HereAround Please check if this fixes your issues. Feel free to create a PR with a test which prevents such regressions in the future.

### DIFF
--- a/FreydCategoriesForCAP/PackageInfo.g
+++ b/FreydCategoriesForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "FreydCategoriesForCAP",
 Subtitle := "Freyd categories - Formal (co)kernels for additive categories",
-Version := "2021.11-02",
+Version := "2021.11-03",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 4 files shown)


HereAround:
@zickgraf Thank you. Yes, a first quick check confirms that this fixes the issue.

I have opened a PR here https://github.com/homalg-project/CAP_project/pull/751 which adds tests to detect these failures in the future.

Once merged, I hope that then all tests in https://github.com/homalg-project/ToricVarieties_project/pull/168 pass again.

zickgraf:
Thanks for checking! I will not merge this PR until the CI in #751 fails to make sure we actually catch the regression.

mohamed-barakat:
#751 fails :)

PR #751, opened 2021-11-11
Add tests to prevent https://github.com/homalg-project/CAP_project/is…

HereAround:
Add tests to prevent https://github.com/homalg-project/CAP_project/issues/746.


zickgraf:
The CI succeeds, so this does not really seem to test the issue.

zickgraf:
The difference is: Your example uses `AsFreydCategoryMorphism`, the example here uses `FreydCategoryMorphism`.

HereAround:
Yes, it appears that this only affects objects and morphisms that were created via `AsFreydCategoryObject` and `AsFreydCategoryMorphism`. I just extended the tests accordingly.

HereAround:
There we go. Now we have the expected failures.

zickgraf:
Thanks! I will now retrigger the CI and merge if it turns green.

PR #756, opened 2021-11-13
Typo in MonoidalCategories

TKuh:


mohamed-barakat:
Funnily, I found this typo today as well :)

PR #757, opened 2021-11-14
Direct product op

mohamed-barakat:


zickgraf:
This change is not compatible with the compiler: Compiling code calling `DirectSum( cat, obj1, obj2, ...)` with at least two objects will give a warning that `DirectSum` cannot be resolved because it does not match a CAP operation. And worse, `DirectSum( cat, obj1 )` will assume that you are calling the CAP operation, but since the CAP operation expects a list, this will not work and result in very weird errors.

Bottom line: When using CAP as a calculator, we don't need the category and can use the convenience `DirectSum( obj1, obj2, ...)`. When writing code which should later be compiled, we have to be more formal and add the category and the brackets for the list.

mohamed-barakat:
Made obsolete by https://github.com/homalg-project/Toposes/commit/099ac0475bba6269c9261a032dc9188ecf0abdc8.

PR #759, opened 2021-11-15
Fix several issues with properties of opposite categories and their objects/morphisms

zickgraf:
1. The properties were collected only once, thus not taking into account
   properties added in other packages.
2. If a categorical property was registered without giving a dual, it was
   assumed to be self-dual. This behavior is dangerous. Now one must always
   explicitly pass the dual or `fail` if the dual is not implemented.
3. The synchronization of categorical properties was completely broken
   because ToDoListEntryToMaintainFollowingAttributes was called with
   arguments having the wrong format. This might be due to the
   documentation of ToDoListEntryToMaintainFollowingAttributes which seems
   to be wrong.
4. The reverse version of one implication was missing.

@mohamed-barakat @TKuh It would be nice of one of you could check that I have not messed up in the parts touching `MonoidalCategories`.


TKuh:
Well, this is actually something I wanted to discuss tomorrow. :)
By now, I believe that this is less obvious than it seems to be.

zickgraf:
I have excluded the problematic properties and will merge if and when the CI succeeds.

PR #760, opened 2021-11-15
Try harder to choose correct reference in automatic documentation

zickgraf:
Also make sure that the documentation of packages we might want to
reference are built.


zickgraf:
The code coverage is increasing because the `manual.six` files are now read in the CI.

PR #764, opened 2021-11-22
Add CategoryConstructor

zickgraf:
The first commit enforces that all CAP operations must now accept the category as the first argument.


zickgraf:
Currently, the newly added code is unused, so the code coverage is very low, but this will change soon.

PR #765, opened 2021-11-22
Import WrapperCategories into CAP

zickgraf:


zickgraf:
I have pushed the result directly (i.e. without clicking "merge") to master to make sure this a proper subtree merge for the "lines of code" script.

PR #772, opened 2021-11-27
AddDerivationToCAP( IsDominating / IsCodominating )

mohamed-barakat:


mohamed-barakat:
How can I give a derivation a low priority?

zickgraf:
> How can I give a derivation a low priority?

I assume you want a derivation not to be triggered in a certain case where it would be more expensive than another derivation? In this case, you should state this expensiveness in the weights of the primitive operations used by the different derivations.

mohamed-barakat:
> I assume you want a derivation not to be triggered in a certain case where it would be more expensive than another derivation? In this case, you should state this expensiveness in the weights of the primitive operations used by the different derivations.

OK, for me this means I have to change weights of primitive operations in some categories provided by CAP_project. I have to discuss the details with you, maybe on Tuesday.

A second point is that the failure of this PR reveals some weird behavior. The GAP operation `RangeCategoryOfHomomorphismStructure` starts appearing in the file:
```
FreydCategoriesForCAP/gap/precompiled_categories/CategoryOfColumnsAsOppositeOfCategoryOfRowsOfArbitraryRingPrecompiled.gi
```
which will cause the compiled `CategoryOfColumns` to fail for rings without `RangeCategoryOfHomomorphismStructure`. I believe, the ring used is not "arbitrary" enough.

Another thing in this context: I see in the compiled lines like
```
if not IS_IDENTICAL_OBJ( cat_1, cat_1 ) then
```
```
if not IS_IDENTICAL_OBJ( deduped_1_1, deduped_1_1 ) then
```
which can probably be skipped.

zickgraf:
> OK, for me this means I have to change weights of primitive operations in some categories provided by CAP_project.

I have now looked at the output and I think everything is fine: MatrixCategory now uses `DecideZeroRows/Columns` instead of `SyzygiesOfRows/Columns`, and the former should be much faster in Singular.

> A second point is that the failure of this PR reveals some weird behavior. The GAP operation `RangeCategoryOfHomomorphismStructure` starts appearing in the file:
> 
> ```
> FreydCategoriesForCAP/gap/precompiled_categories/CategoryOfColumnsAsOppositeOfCategoryOfRowsOfArbitraryRingPrecompiled.gi
> ```

I don't see this, neither locally nor in the CI. Only the `autogen.gd` files change, which is expected.

> Another thing in this context: I see in the compiled lines like
> 
> ```
> if not IS_IDENTICAL_OBJ( cat_1, cat_1 ) then
> ```
> 
> ```
> if not IS_IDENTICAL_OBJ( deduped_1_1, deduped_1_1 ) then
> ```
> 
> which can probably be skipped.

Yes, that's expected/known and only occurs in the `IsWellDefined*` operations. The compiler currently resolved calls to `CapCategory` and simplifies this to `cat_1` in the first case. Resolving `CapCategory` will not be required anymore once we have the type system.

zickgraf (changes requested):
I would suggest using `LiftAlongMonomorphism` and `ColiftAlongEpimorphism` to get the most special case (e.g. in Freyd this makes a huge difference as far as I know). Also, could you put the closing bracket in the commit message? Reason: https://xkcd.com/859/ :D

mohamed-barakat:
According to the documentation, the two arguments of `IsDominating` are not necessarily monos and the two arguments of `IsCodominating` are not necessarily epis. This is in line with what I expect.

zickgraf:
The documentation says

> The arguments are two subobjects...

and "subobjects" are defined in the documentation as

> Subobjects of an object c are monomorphisms with range c...

Also, in the diagram alpha and beta are written as monos. Should the definition of "subobject" be generalized?

mohamed-barakat:
No, you are right.

zickgraf:
Sorry, I just realized that my suggestion does not make sense because we do not have `IsLiftableAlongMonomorphism` (because `LiftAlongMonomorphism` always assumes that a lift exists).

mohamed-barakat:
> Sorry, I just realized that my suggestion does not make sense because we do not have `IsLiftableAlongMonomorphism` (because `LiftAlongMonomorphism` always assumes that a lift exists).

I don't understand your comment.

zickgraf:
Ah, I was wrong thinking I was wrong :D Now I see we actually have `IsLiftableAlongMonomorphism`. I expected we do not have this for the reason mentioned above. Grepping shows that we do not actually implement `IsLiftableAlongMonomorphism` anywhere, and I'm not sure there exists a case where an implementation would be faster than `IsLiftable`. But if we have it, we should use it. We just have to be careful with the order of the arguments (I think the arguments are swapped compared to `IsLiftable`).

mohamed-barakat:
> Ah, I was wrong thinking I was wrong :D Now I see we actually have `IsLiftableAlongMonomorphism`. I expected we do not have this for the reason mentioned above. Grepping shows that we do not actually implement `IsLiftableAlongMonomorphism` anywhere, and I'm not sure there exists a case where an implementation would be faster than `IsLiftable`.

If I remember correctly: For `(Skeletal)FinSets` there is a case distinction that can be dropped for `ColiftAlongEpimorphism` which has to be treated in `Colift`. Apart from that I don't have an example at hand.

> But if we have it, we should use it. We just have to be careful with the order of the arguments (I think the arguments are swapped compared to `IsLiftable`).

Ah, right, this thing again :(

zickgraf:
> If I remember correctly: For `(Skeletal)FinSets` there is a case distinction that can be dropped for `ColiftAlongEpimorphism` which has to be treated in `Colift`. Apart from that I don't have an example at hand.

What I meant is: I know a few places where `(Co)LiftAlongMono/Epimorphism` is faster than `(Co)Lift`, but I could not find an instance where we have `Is(Co)LiftableAlongMono/Epimorpism` installed.

CAP/gap/DerivedMethods.gi
```diff
@@ -1121,7 +1131,17 @@ AddDerivationToCAP( IsDominating,
...
+
+##
+AddDerivationToCAP( IsCodominating,
+                    [ [ IsColiftable, 1 ] ],
+
+  function( cat, factor1, factor2 )
+    
+    return IsColiftable( cat, factor2, factor1 );
+    
+end : Description := "IsCodominating using IsColiftable" );
```
zickgraf, line 1144:
```suggestion
    return IsColiftableAlongEpimorphism( cat, factor2, factor1 );
    
end : Description := "IsCodominating using IsColiftableAlongEpimorphism" );
```

zickgraf (changes requested):

CAP/gap/DerivedMethods.gi
mohamed-barakat: Sorry, was gone after several rebasing.

CAP/gap/DerivedMethods.gi
```diff
@@ -1090,6 +1090,16 @@ AddDerivationToCAP( IsEqualAsFactorobjects,
     
 end : Description := "IsEqualAsFactorobjects(factor1, factor2) if factor1 dominates factor2 and vice versa" );
 
+##
+AddDerivationToCAP( IsDominating,
+                    [ [ IsLiftable, 1 ] ],
```
zickgraf, line 1095:
```suggestion
```

CAP/gap/DerivedMethods.gi
```diff
@@ -1121,7 +1131,17 @@ AddDerivationToCAP( IsDominating,
     
     return IsZeroForMorphisms( cat, composition );
     
-end : Description := "IsDominating(sub1, sub2) by deciding if sub1 composed with CokernelProjection( cat,sub2) is zero" );
+end : Description := "IsDominating(sub1, sub2) by deciding if sub1 composed with CokernelProjection(cat, sub2) is zero" );
+
+##
+AddDerivationToCAP( IsCodominating,
+                    [ [ IsColiftable, 1 ] ],
```
zickgraf, line 1138:
```suggestion
```

zickgraf (changes requested):

LinearAlgebraForCAP/gap/precompiled_categories/MatrixCategoryPrecompiled.gi
```diff
@@ -1867,11 +1867,11 @@ function ( cat_1, arg2_1, arg3_1 )
     local deduped_1_1, deduped_2_1;
     deduped_2_1 := UnderlyingMatrix( arg3_1 );
     deduped_1_1 := UnderlyingMatrix( arg2_1 );
-    return IsZero( deduped_1_1 * SyzygiesOfColumns( deduped_2_1 ) ) and IsZero( deduped_2_1 * SyzygiesOfColumns( deduped_1_1 ) );
+    return IsZero( DecideZeroRows( deduped_1_1, deduped_2_1 ) ) and IsZero( DecideZeroRows( deduped_2_1, deduped_1_1 ) );
```
mohamed-barakat, line 1870: For the record: I can think of a context in which this change would pose a regression. This would require an implementation of `SyzygiesOfColumns`, which is cheaper than `BasisOfRowModule` which is called by `DecideZeroRows`.

zickgraf: For the record 2: We have decided to not tackle this point for now, because

1. When using Singular, `BasisOfRowModule` is cheaper than `SyzygiesOfColumns`,
2. probably there are many more of such cases in `LinearAlgebraForCAP`, so one should go over the whole code at once,
3. if we make the primitive installation of `IsLiftable` more expensive, than it will be derived from the homomorphism structure leading to an infinite recursion, because the derivation mechanism currently does not take the weights in the range category into account.

PR #783, opened 2021-12-07
Various improvements

zickgraf:
* Distinguish final derivations and precompiled functions from primitive operations (ping @kamalsaleh)
* Remove a trailing new line (ping @mohamed-barakat )
* Add tests for ComponentOfMorphismInto/FromDirectSum in AdditiveClosure

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2021.12-01",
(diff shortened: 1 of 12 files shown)


mohamed-barakat (approved):

zickgraf:
Note for future reference: With this PR, adding precompiled functions does not affect `ListPrimitivelyInstalledOperationsOfCategory`. This might be the desired result in some cases (e.g. when looking at `InfoOfInstalledOperationsOfCategory`), but not in others (e.g. in the case of the homomorphism structure for CategoryOfRows over an exterior algebra) -> We will have to refine this based on concrete applications in the future.

PR #791, opened 2021-12-13
Fix coclosed method records

TKuh:


mohamed-barakat:
This does not yet fix the bug reported in https://github.com/homalg-project/FinSetsForCAP/pull/73#issuecomment-991710363.

PR #794, opened 2021-12-14
Add some checks and warnings

zickgraf:


zickgraf:
The CI only fails because Codecov has issues -> I will merge.

PR #796, opened 2021-12-16
added opposite categorical properties

mohamed-barakat:
see https://github.com/homalg-project/Toposes/commit/695b2758a4f4529fe584e10f9763794993ad1056#commitcomment-62011137

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2021.12-07",
+Version := "2021.12-08",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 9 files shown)


GradedModulePresentationsForCAP/examples/CohP1.g
```diff
@@ -15,7 +15,7 @@ S := GradedRing( Q["x,y"] );
 Sgrmod := GradedLeftPresentations( S );
 #! The category of graded left f.p. modules over Q[x,y] (with weights [ 1, 1 ])
 InfoOfInstalledOperationsOfCategory( Sgrmod );
-#! 40 primitive operations were used to derive 232 operations for this category which
+#! 40 primitive operations were used to derive 246 operations for this category which
```
zickgraf, line 18: I think these numbers should not change, because they have nothing to do with `Opposite` as far as I can tell. Do you have these numbers from the CI or from your local computer? In the latter case, probably `Toposes` increases this number.

mohamed-barakat: Yes, right. I keep forgetting that I need to test `CAP_project` without `Toposes`.

mohamed-barakat: ... `HigherHomologicalAlgebra`.

Issue #780, opened 2021-12-03
The role of `ObjectConstructor` and `MorphismConsctructor`


zickgraf:
Currently, `ObjectConstructor` and `MorphismConsctructor` have two different applications:

1. They are used to access different categories with a common interface, e.g. `MatrixCategory` and `CategoryOfRows` [here](https://github.com/homalg-project/Algebroids/blob/9d89c1725374bf970ea111b7c8575549786998a0/gap/Algebroids.gi#L767).
2. They are used to decouple the "external" (i.e., user visible) data structure from the "internal" data structure (the one used in the implementation) in combination with `WrapperCategory`, e.g. for `CategoryOfRowsAsAdditiveClosureOfRingAsCategory` [here](https://github.com/homalg-project/CAP_project/blob/fb0a13d8ca2a6095bf812d79ae9100be247bed44/FreydCategoriesForCAP/gap/CategoryOfRowsAsAdditiveClosureOfRingAsCategory.gi#L24).

These two applications are mutually exclusive: For the second application, the `ObjectConstructor` of `CategoryOfRowsAsAdditiveClosureOfRingAsCategory` has to be different from the `ObjectConstructor` of `CategoryOfRows`, while for the first they should be the same. But both applications can also be solved in other ways (the first one by using the "old" constructors as before, the second one by introducing new options for `WrapperCategory`).

We should decide what the role of `ObjectConstructor` and `MorphismConsctructor` should actually be. I still have to think about this, ideas and suggestions are welcome!

mohamed-barakat:
I like the names.

mohamed-barakat:
I forgot what the next step is.

zickgraf:
> I forgot what the next step is.

There is no next step for you: You can continue using the existing system until I find the time to introduce the new system.

zickgraf:
This is more complicated than expected (I also have to touch `CategoryConstructor` for this....) But independent of that there is a further question: Assume that we are in the setting of modeling the category of columns as the opposite of the category of rows. Then `modeling_to_modeled_object` would be `obj -> CategoryOfColumnsObject( cat, RankOfObject( Opposite( obj ) ) )` and `object_constructor` would be `rank -> CategoryOfColumnsObject( cat, rank )`. Note that in this way we specify `CategoryOfColumnsObject` twice. This has two disadvantages in general:
[...]

PR #800, opened 2021-12-27
AddObjectDatum/AddMorphismDatum in AdditiveClosure

mohamed-barakat:
I need this in LazyCategories for a uniform behavior of the categories in the tower

### DIFF
--- a/FreydCategoriesForCAP/PackageInfo.g
+++ b/FreydCategoriesForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "FreydCategoriesForCAP",
 Subtitle := "Freyd categories - Formal (co)kernels for additive categories",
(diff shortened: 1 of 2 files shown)


mohamed-barakat:
`Algebroids/examples/PrecompileAdditiveClosureOfAlgebroid.g` fails with the PR.

zickgraf:
> I need this in LazyCategories for a uniform behavior of the categories in the tower

Before using `ObjectConstructor` et al. for uniform behavior we should discuss https://github.com/homalg-project/CAP_project/issues/780.

In any case, when adding `*Datum` one should also add `*Constructor`, as otherwise there is not really a specification.

zickgraf:
@mohamed-barakat Now that #780 is solved, this PR can be merged after rebasing and adding the constructors.

zickgraf (approved):
Thanks!

PR #805, opened 2022-01-20
Remove additional filters from filter lists of IsEqualAsSub/Factorobjects and Is(Co)Dominating

zickgraf:
I'm not sure why those were added in the first place. If the reason was type checking: this is not done anywhere else, for example `LiftAlongMonomorphism` does not check that it gets a monomorphism.

I want to get rid of this special case (and in a follow up remove CAP's support for this) because it complicates any code handling `filter_list`s.

@sebastianpos Couly you quickly check if you see any problems with this?

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2021.12-09",
+Version := "2022.01-01",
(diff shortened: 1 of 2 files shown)


sebastianpos:
It was added as a type check. You may delete it, but the semantics should not be altered (i.e., if the inputs are not subobject, and if there is no other implementation of the type check, then the output can be garbage).

zickgraf:
> but the semantics should not be altered

Of course, the specification will not change: the documentation explicitly requires subobjects resp. factorobjects. And with that we have "garbage in, garbage out" :D

Thanks for the fast reply!

PR #808, opened 2022-01-24
Miscellaneous changes

zickgraf:
See commit message for details.

Note: The object and morphism filters of the range category of a homomorphism structure are now used in method installation if the range category is known. Thus, `SetRangeCategoryOfHomomorphismStructure` should be called before calling `AddDistinguishedObject` et al. A warning is displayed if this is not the case.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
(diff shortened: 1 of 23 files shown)


mohamed-barakat (commented):
Is it possible to put the changes in MonoidalCategories in a separated PR?

zickgraf:
> Is it possible to put the changes in MonoidalCategories in a separated PR?

Yes: https://github.com/homalg-project/CAP_project/pull/809

PR #812, opened 2022-01-25
added WithGivenObjects-objects variants

mohamed-barakat:
* IsomorphismFromDualToInternalHomWithGivenObjects
* IsomorphismFromInternalCoHomToCoDualWithGivenObjects

### DIFF
--- a/MonoidalCategories/gap/ClosedMonoidalCategories.gd
+++ b/MonoidalCategories/gap/ClosedMonoidalCategories.gd
@@ -278,6 +278,18 @@ DeclareAttribute( "IsomorphismFromDualToInternalHom",
                   IsCapCategoryObject );
 
 #! @Description
+#! The arguments are an object $s = a^{\vee}$,
+#! an object $a$,
+#! and an object $r = \mathrm{\underline{Hom}}(a,1)$.
+#! The argument is an object $a$.
+#! The output is the isomorphism
+#! $\mathrm{IsomorphismFromDualToInternalHom}_{a}: a^{\vee} \rightarrow \mathrm{\underline{Hom}}(a,1)$.
+#! @Returns a morphism in $\mathrm{Hom}(a^{\vee}, \mathrm{\underline{Hom}}(a,1))$.
+#! @Arguments s,a,r
+DeclareOperation( "IsomorphismFromDualToInternalHomWithGivenObjects",
+                  [ IsCapCategoryObject, IsCapCategoryObject, IsCapCategoryObject ] );
+
+#! @Description
 #! The argument is an object $a$.
 #! The output is the isomorphism
 #! $\mathrm{IsomorphismFromInternalHomToDual}_{a}: \mathrm{\underline{Hom}}(a,1) \rightarrow a^{\vee}$.
--- a/MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
+++ b/MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
@@ -907,3 +907,15 @@ AddDerivationToCAP( MonoidalPostComposeMorphismWithGivenObjects,
     
 end : CategoryFilter := IsSymmetricClosedMonoidalCategory and IsStrictMonoidalCategory,
       Description := "MonoidalPostComposeMorphismWithGivenObjects using evaluation, and tensor hom adjunction" );
+
+##
+AddDerivationToCAP( IsomorphismFromDualToInternalHom,
+        
+  function( cat, a )
+    
+    return PreCompose( cat,
+                   RightUnitorInverseWithGivenTensorProduct( cat, a ),
+                   MorphismFromTensorProductToInternalHom( cat, a, TensorUnit( cat ) ) );
+    
+end : CategoryFilter := IsSymmetricClosedMonoidalCategory,
+      Description := "IsomorphismFromDualToInternalHom using MorphismFromTensorProductToInternalHom" );
--- a/MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
+++ b/MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
@@ -919,3 +919,15 @@ AddDerivationToCAP( IsomorphismFromDualToInternalHom,
     
 end : CategoryFilter := IsSymmetricClosedMonoidalCategory,
       Description := "IsomorphismFromDualToInternalHom using MorphismFromTensorProductToInternalHom" );
+
+##
+AddDerivationToCAP( IsomorphismFromDualToInternalHomWithGivenObjects,
+        
+  function( cat, source, a, range )
+    
+    return PreCompose( cat,
+                   RightUnitorInverseWithGivenTensorProduct( cat, a, source ),
+                   MorphismFromTensorProductToInternalHomWithGivenObjects( cat, source, a, TensorUnit( cat ), range ) );
+    
+end : CategoryFilter := IsSymmetricClosedMonoidalCategory,
+      Description := "IsomorphismFromDualToInternalHomWithGivenObjects using MorphismFromTensorProductToInternalHomWithGivenObjects" );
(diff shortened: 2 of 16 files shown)


MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -907,3 +907,27 @@ AddDerivationToCAP( MonoidalPostComposeMorphismWithGivenObjects,
     
 end : CategoryFilter := IsSymmetricClosedMonoidalCategory and IsStrictMonoidalCategory,
       Description := "MonoidalPostComposeMorphismWithGivenObjects using evaluation, and tensor hom adjunction" );
+
+##
+AddDerivationToCAP( IsomorphismFromDualToInternalHom,
+        
+  function( cat, a )
```
TKuh, line 914: I would prefer a diagram for easier reading. That would also be consistent with the other derivations.

MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -907,3 +907,27 @@ AddDerivationToCAP( MonoidalPostComposeMorphismWithGivenObjects,
...
+                   MorphismFromTensorProductToInternalHom( cat, a, TensorUnit( cat ) ) );
+    
+end : CategoryFilter := IsSymmetricClosedMonoidalCategory,
+      Description := "IsomorphismFromDualToInternalHom using MorphismFromTensorProductToInternalHom" );
+
+##
+AddDerivationToCAP( IsomorphismFromDualToInternalHomWithGivenObjects,
+        
+  function( cat, source, a, range )
+
```
TKuh, line 927: Diagram.

MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -907,3 +907,27 @@ AddDerivationToCAP( MonoidalPostComposeMorphismWithGivenObjects,
...
+      Description := "IsomorphismFromDualToInternalHom using MorphismFromTensorProductToInternalHom" );
+
+##
+AddDerivationToCAP( IsomorphismFromDualToInternalHomWithGivenObjects,
+        
+  function( cat, source, a, range )
+    
+    return PreCompose( cat,
+                   RightUnitorInverseWithGivenTensorProduct( cat, a, source ),
+                   MorphismFromTensorProductToInternalHomWithGivenObjects( cat, source, a, TensorUnit( cat ), range ) );
```
TKuh, line 930: We have `source = a^v` but `MorphismFromTensorProductToInternalHomWithGivenObjects` expects the second argument to be `a^v x 1`.

MonoidalCategories/gap/SymmetricCoclosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -919,3 +919,27 @@ AddDerivationToCAP( MonoidalPostCoComposeMorphismWithGivenObjects,
     
 end : CategoryFilter := IsSymmetricCoclosedMonoidalCategory and IsStrictMonoidalCategory,
       Description := "MonoidalPostCoComposeMorphismWithGivenObjects using coclosed evaluation, and cohom tensor adjunction" );
+
+##
+AddDerivationToCAP( IsomorphismFromInternalCoHomToCoDual,
+        
+  function( cat, a )
+
```
TKuh, line 927: Diagram.

MonoidalCategories/gap/SymmetricCoclosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -919,3 +919,27 @@ AddDerivationToCAP( MonoidalPostCoComposeMorphismWithGivenObjects,
...
+                   RightUnitorWithGivenTensorProduct( cat, a ) );
+    
+end : CategoryFilter := IsSymmetricCoclosedMonoidalCategory,
+      Description := "IsomorphismFromInternalCoHomToCoDual using MorphismFromInternalCoHomToTensorProduct" );
+
+##
+AddDerivationToCAP( IsomorphismFromInternalCoHomToCoDualWithGivenObjects,
+        
+  function( cat, source, a, range )
+
```
TKuh, line 939: Diagram.

MonoidalCategories/gap/SymmetricCoclosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -919,3 +919,27 @@ AddDerivationToCAP( MonoidalPostCoComposeMorphismWithGivenObjects,
...
+      Description := "IsomorphismFromInternalCoHomToCoDual using MorphismFromInternalCoHomToTensorProduct" );
+
+##
+AddDerivationToCAP( IsomorphismFromInternalCoHomToCoDualWithGivenObjects,
+        
+  function( cat, source, a, range )
+    
+    return PreCompose( cat,
+                   MorphismFromInternalCoHomToTensorProductWithGivenObjects( cat, source, TensorUnit( cat ), a, range ),
+                   RightUnitorWithGivenTensorProduct( cat, a, range ) );
```
TKuh, line 942: `RightUnitorWithGivenTensorProduct` expects the third argument to be `a_v x 1` but we have `range = a_v`.

TKuh (changes requested):
In addition to my other remarks, I have the following general questions (for the future) on how to handle situations (I would be especially interested in @zickgraf's opinion):

1. Right now, I noticed, that many of the derivations in `Symmetric*` don't actually need the symmetric structure and could be mathematically placed in `MonoidalClosed` or `MonoidalCoclosed`. What are the reasons that these are in the symmetric files? (E.g. the very first derivation in `SymmetricClosedMonoidalCategoriesDerivedMethods.gi`)
2. One could now also add a `WithGiven` function and derivations for `IsomorphismFromInternalHomToDual` and `IsomorphismFromCoDualToInternalCoHom` (in the rigid case) which would be very similar. Would that be considered nice-to-have, or is it better to only implement things one actually uses? 
3. The derivations in this PR could also additionally have a strict version. Again: nice-to-have or better not? There are also not many derivations which have a strict counterpart, even though they could.

zickgraf:
ad 2.: As far as I know, the primary purpose why all those isomorphisms were added was to derive things in a consistent way e.g. `DualOnObjects` from `InternalHom` etc. (more details next Tuesday). In this setting, having `WithGiven` operations is not useful, because for calling those I must already be able to compute the objects. We have another mechanism for deriving things consistently now (final derivations with exclusion lists), so we could even get rid of the isomorphisms altogether.

ad 3.: Thinking about this, having optimizations for the strict case at all sound dangerous to me, again for reasons of consistency. However, those were added by Sebastian and I'm sure he had this in mind. In any case, the compiler makes these additional derivations obsolete, at least in all categories where composing with a identity cancels on the implementation level.

MonoidalCategories/gap/ClosedMonoidalCategories.gd
```diff
@@ -277,6 +277,18 @@ DeclareOperation( "MorphismFromTensorProductToInternalHomWithGivenObjects",
 DeclareAttribute( "IsomorphismFromDualToInternalHom",
                   IsCapCategoryObject );
 
+#! @Description
+#! The arguments are an object $s = a^{\vee}$,
+#! an object $a$,
+#! and an object $r = \mathrm{\underline{Hom}}(a,1)$.
+#! The argument is an object $a$.
```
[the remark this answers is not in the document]
mohamed-barakat: Yes, a copy-paste mistake :)

MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
mohamed-barakat: Very good, this revealed a mistake :)

MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
mohamed-barakat: Right.

MonoidalCategories/gap/SymmetricCoclosedMonoidalCategoriesDerivedMethods.gi
mohamed-barakat: Right.

mohamed-barakat:
> ad 2.: As far as I know, the primary purpose why all those isomorphisms were added was to derive things in a consistent way e.g. `DualOnObjects` from `InternalHom` etc. (more details next Tuesday). In this setting, having `WithGiven` operations is not useful, because for calling those I must already be able to compute the objects. We have another mechanism for deriving things consistently now (final derivations with exclusion lists), so we could even get rid of the isomorphisms altogether.

I agree. And since I don't need this PR anymore, I will close it.

PR #787, opened 2021-12-09
Coclosed structure for CategoryOfRows/LinearAlgebraForCap

TKuh:

PR #817, opened 2022-01-26
Disable compatibility in tests

TKuh:
To increase the speed, see https://github.com/homalg-project/CAP_project/pull/787#issuecomment-1022272044
This is only until I refine the tests.

The two tests in LinearAlgebra still take roughly double the time than those of Freyd.


zickgraf:
Thanks for the quick reaction! Now the CI time is back to normal :-)

PR #823, opened 2022-01-31
Monoidal test functions

TKuh:
Some questions:

1. I created new files for the tests, however, they could also be placed in `MonoidalCategories.gd`, `ClosedMonoidalCategories.gd`, etc. what is preferable? (Or a different place?)
2. I'm playing with the idea to add `verbose` and `exclude` options. `verbose` could output the method currently tested and with `exclude` one could specify methods to be excluded from a test call. This could be useful when one wants to test some methods with larger input, but wants to prevent other methods from being called with this specific input (e.g. because it might take too long). How do you feel about that? 
3. In the old test I had this check: https://github.com/homalg-project/CAP_project/blob/master/FreydCategoriesForCAP/tst/TensorCorrespondenceRowsAndOppositeOfRows.tst#L1053-L1060 
(it checks if the output of the `adjunction maps` agree with the output of `EvaluationMorphism` on certain iput, etc.) and something similar here: https://github.com/homalg-project/CAP_project/blob/master/FreydCategoriesForCAP/tst/TensorCorrespondenceRowsAndOppositeOfRows.tst#L1625-L1641 (checks if the general `compatibility` agrees with `duality compatibility` on certain input).
These do not really fit into the new test files. Should we remove it, or can it be reused, for example in a separate check? There also exists a function `CheckBraiding`.


FreydCategoriesForCAP/tst/TensorCorrespondenceRowsAndOppositeOfRows.tst
```diff
@@ -4,2842 +4,216 @@
 
 gap> START_TEST("TensorCorrespondenceRowsAndOppositeOfRows.tst");
 
-gap> LoadPackage( "FreydCategoriesForCAP", false );
+gap> LoadPackage("FreydCategoriesForCap", false);
```
zickgraf: Please keep the spaces and the capitalization of `CAP`.

MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gd
```diff
@@ -0,0 +1,24 @@
...
+#! @Chapter Examples and Tests
+
+#! @Section Test functions
+
+#! @Description
+#! The arguments are
+#! * a CAP category $cat$
+#! * an assertion level $assertlevel$
+#! * objects $a, b$
+#! This function checks for every object and morphism
```
zickgraf, line 16:
```suggestion
#! This function checks for every operation
```
With "object" and "morphism" we usually mean GAP objects which are CAP objects or morphisms.

MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gi
```diff
@@ -0,0 +1,47 @@
...
+
+        if CanCompute( cat, "Braiding" ) then 
+            
+            braiding_a_b := Braiding( a, b );;
+            braiding_b_a := Braiding( b, a );;
+            
+            braiding_inverse_a_b_op := BraidingInverse( a_op, b_op );;
+            braiding_inverse_b_a_op := BraidingInverse( b_op, a_op );;
+            
+            Assert( assertlevel, IsEqualForMorphisms( braiding_inverse_a_b_op, Opposite( braiding_a_b ) ) );
```
zickgraf, line 29:
```suggestion
            Assert( 0, IsEqualForMorphisms( braiding_inverse_a_b_op, Opposite( braiding_a_b ) ) );
```
What's the motivation behind `assertlevel`? I think we can always use 0, because we want the assert to always be checked.

zickgraf, line 29: I have not checked, but I think we have to use `IsCongruentForMorphisms` here.

zickgraf (changes requested):
Nice work :-) Just some random comments below.

One further idea/suggestion for discussion (feel free to dismiss the idea as "over-optimization" :D ):

The test functions are now "far away" from the method records, i.e. in another file. This means that probably things will get out-of-sync if more operations are added and one is not aware of the test functions. So a further enhancement would be a mechanism to "register" test functions. That is, one could call `RegisterTestFunction( "BraidedMonoidalCategories", "Braiding", { a, b } -> ... )` after the corresponding method record entry.

TKuh:
Interestingly, the new tests run a bit faster than the old one, even though the compatibility morphisms are now activated in the new tests with the same input that previously caused performance troubles in #817:

Input in old tests: 
- https://github.com/homalg-project/CAP_project/blob/master/FreydCategoriesForCAP/tst/TensorCorrespondenceRowsAndOppositeOfRows.tst#L20-L23
- https://github.com/homalg-project/CAP_project/blob/master/FreydCategoriesForCAP/tst/TensorCorrespondenceRowsAndOppositeOfRows.tst#L1503-L1513

Input in new tests: 
- https://github.com/homalg-project/CAP_project/blob/1e0b496eb4d2aa841abc91610804019c9e63fa99/FreydCategoriesForCAP/tst/TensorCorrespondenceRowsAndOppositeOfRows.tst#L85-L105
- https://github.com/homalg-project/CAP_project/blob/1e0b496eb4d2aa841abc91610804019c9e63fa99/MonoidalCategories/gap/ClosedMonoidalCategoriesTest.gi#L407-L416

Also, the tests in LinearAlg and Freyd are now roughly equally fast (for me at least):
`6687 ms (2462 ms GC) and 1.27GB allocated for TensorCorrespondenceMatrixCategoryAndOpposite.tst`
`6572 ms (2443 ms GC) and 1.23GB allocated for TensorCorrespondenceRowsAndOppositeOfRows.tst`

mohamed-barakat (approved):
Thank you very much Tom.

MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gi
```diff
@@ -0,0 +1,61 @@
...
+        a_op := Opposite( a );;
+        b_op := Opposite( b );;
+        
+        verbose := ValueOption( "verbose" ) = true;;
+        
+        if CanCompute( cat, "Braiding" ) then
+            
+            if verbose then
+                
+                Display( "Testing 'Braiding' ..." );
```
zickgraf, line 27: You can put a comment `# COVERAGE_IGNORE_NEXT_LINE` above lines which solely exist for debugging purposes to ignore them when gathering code coverage.

MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gi
```diff
@@ -0,0 +1,61 @@
...
+    function( cat, a, b )
+        
+        local opposite, verbose,
+              
+              a_op, braiding_a_b, braiding_a_b_op, braiding_inverse_a_b, braiding_inverse_a_b_op, 
+              b_op, braiding_b_a, braiding_b_a_op, braiding_inverse_b_a, braiding_inverse_b_a_op;
+        
+        opposite := Opposite( cat );;
+        
+        a_op := Opposite( a );;
```
zickgraf, line 18: The double semicolon is not needed anymore.

zickgraf (approved):
Just two comments which you can address or ignore, just as you like.

One general thing for possible future improvements: I was wondering how much of these tests could be generated automatically just by looking at `dual_operation` in the method record.

MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gi
TKuh: Oh, that's nice. Added it.

TKuh:
> One general thing for possible future improvements: I was wondering how much of these tests could be generated automatically just by looking at `dual_operation` in the method record.

I think the `CanCompute(...)` blocks and the local variables shouldn't be hard to auto-generate. What requires more thought are the parameters for the test-functions. Topic for Tuesday?

zickgraf (approved):

PR #841, opened 2022-02-13
Document `only_primitive_operations` for monoidal tests

TKuh:
As @zickgraf pointed out here: https://github.com/homalg-project/FinSetsForCAP/pull/91#issuecomment-1036257261, passing `only_primitive_operations` to `Opposite()` might trigger more derivations. This is currently not documented, but I think it might be helpful to let users know that this could (optionally) be also used for the tests. 

09ba50613cd0ea0f8144a8779bd6254a6c041ff2 fixes a typo in the documentation.

### DIFF
--- a/MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gd
+++ b/MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gd
@@ -19,6 +19,14 @@
 #! with the parameters given above and
 #! compared to the equivalent computation in
 #! the opposite category of $cat$.
-#! Pass the option 'verbose := true' to output more information.
+#! Pass the options
+#! * `verbose := true` to output more information.
+#! * `only_primitive_operations := true`,
+#!    which is passed on to Opposite(),
+#!    to only primitively install
+#!    dual operations for primitively
+#!    installed operations in $cat$.
+#!    The advantage is, that more derivations might be tested.
+#!    On the downside, this might test fewer dual_pre/postprocessor_funcs.
 #! @Arguments cat, a, b
 DeclareGlobalFunction( "BraidedMonoidalCategoriesTest" );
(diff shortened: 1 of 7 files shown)


MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gd
```diff
@@ -19,6 +19,10 @@
 #! with the parameters given above and
 #! compared to the equivalent computation in
 #! the opposite category of $cat$.
-#! Pass the option 'verbose := true' to output more information.
+#! Pass the options 
+#! * 'verbose := true' to output more information.
+#! * 'only_primitive_operations := true' to only use 
+#!    dual operations for operations primitively installed in $cat$.
```
zickgraf, line 25: For me the phrasing sounds a bit misleading because I read "use" as "test" in this context. Suggestion:
```suggestion
#! * 'only_primitive_operations := true' to only primitively install 
#!    dual operations for operations primitively installed in $cat$.
```

And I think it would be good to also mention the downsides, i.e. that fewer dual_pre/postprocessor_funcs might get tested.

zickgraf (changes requested):
Thanks!

MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gd
TKuh: > For me the phrasing sounds a bit misleading because I read "use" as "test" in this context. [...] And I think it would be good to also mention the downsides, i.e. that fewer dual_pre/postprocessor_funcs might get tested.

Yes, I wasn't satisfied with my wording as well. :/ I also don't like how complicated that sentence reads, maybe we can briefly talk about it tomorrow.

mohamed-barakat (dismissed):

zickgraf (approved):

PR #846, opened 2022-02-21
Add IsBijectiveObject

sebastianpos:
A bijective object is an object which is both projective and injective.

Unfortunately, I was not able to use `IsBijective`, since this appears to be an alias(?) of `(((IsSingleValued and IsTotal) and IsInjective) and IsSurjective)` in GAP. So, I opted for `IsBijectiveObject`.

### DIFF
--- a/CAP/gap/AddFunctions.autogen.gd
+++ b/CAP/gap/AddFunctions.autogen.gd
@@ -1532,6 +1532,25 @@ DeclareOperation( "AddIsAutomorphism",
 #! @Description
 #! The arguments are a category $C$ and a function $F$.
 #! This operation adds the given function $F$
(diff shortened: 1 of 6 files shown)


sebastianpos:
I need an explanation on what I did wrong w.r.t. CompilerForCAP.

zickgraf:
The new operation is automatically derived for `MatrixCategory`. You have to run the compiler tests (or simply `CompilerForCAP/examples/PrecompileLinearAlgebraForCAP.g`) to update the precompiled code, and include the resulting changes in the PR. In contrast to the autogen files, this cannot be done automatically on each startup because precompilation runs for seconds or even minutes.

sebastianpos:
Thanks for your explanation.

PR #851, opened 2022-02-26
Declare and install generic algorithms for projective and injective dimension

sebastianpos:
My introduction of the type `non_neg_integer_or_infinity` leads to deletions of precompiled code, which I think is a bad sign.

@zickgraf Did I forget something in the introduction of that type, or do you need to adjust the compiler?

### DIFF
--- a/CAP/gap/CategoryObjects.gd
+++ b/CAP/gap/CategoryObjects.gd
@@ -416,3 +416,25 @@ DeclareOperation( "SimplifyObject_IsoFromInputObject",
 #! @Arguments A, i
 DeclareOperation( "SimplifyObject_IsoToInputObject",
                   [ IsCapCategoryObject, IsObject ] );
+
+###################################
+##
+#! @Section Dimensions
+##
+###################################
+
+#! @Description
+#! The argument is an object $A$.
+#! The output is a the projective dimension of $A$.
+#! @Returns a nonnegative integer or infinity
+#! @Arguments A
+DeclareAttribute( "ProjectiveDimension",
+                  IsCapCategoryObject );
+
+#! @Description
+#! The argument is an object $A$.
+#! The output is a the injective dimension of $A$.
+#! @Returns a nonnegative integer or infinity
+#! @Arguments A
+DeclareAttribute( "InjectiveDimension",
+                  IsCapCategoryObject );
--- a/CAP/gap/AddFunctions.autogen.gd
+++ b/CAP/gap/AddFunctions.autogen.gd
@@ -1418,6 +1418,25 @@ DeclareOperation( "AddInjectiveColift",
 #! @Description
 #! The arguments are a category $C$ and a function $F$.
 #! This operation adds the given function $F$
+#! to the category for the basic operation `InjectiveDimension`.
+#! $F: ( arg2 ) \mapsto \mathtt{InjectiveDimension}(arg2)$.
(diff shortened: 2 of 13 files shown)


CAP/gap/CategoryObjects.gd
```diff
@@ -416,3 +416,25 @@ DeclareOperation( "SimplifyObject_IsoFromInputObject",
...
+###################################
+##
+#! @Section Dimensions
+##
+###################################
+
+#! @Description
+#! The argument is an object $A$.
+#! The output is a the projective dimension of $A$.
+#! @Returns a non negative integer or infinity
```
mohamed-barakat, line 429:
```suggestion
#! @Returns a nonnegative integer or infinity
```

mohamed-barakat, line 429:
```suggestion
#! @Returns a nonnegative integer or infinity
```

CAP/gap/CategoryObjects.gd
```diff
@@ -416,3 +416,25 @@ DeclareOperation( "SimplifyObject_IsoFromInputObject",
...
+#! @Arguments A
+DeclareAttribute( "ProjectiveDimension",
+                  IsCapCategoryObject );
+
+#! @Description
+#! The argument is an object $A$.
+#! The output is a the injective dimension of $A$.
+#! @Returns a non negative integer or infinity
```
mohamed-barakat, line 437:
```suggestion
#! @Returns a nonnegative integer or infinity
```

FreydCategoriesForCAP/gap/precompiled_categories/CategoryOfColumnsAsOppositeOfCategoryOfRowsOfFieldPrecompiled.gi
```diff
@@ -708,172 +708,6 @@ end
         
     , 100 );
     
-    ##
```
mohamed-barakat, line 711: Why did this code disappear?

zickgraf:
The code disappears because `CategoryOfColumns` uses `WrapperCategory` which uses the internal `CategoryConstructor` which should also be adapted here

https://github.com/homalg-project/CAP_project/blob/95ed52baa29546b260ecb8609f2a1200c967161f/CAP/gap/CategoryConstructor.gi#L247

and here

https://github.com/homalg-project/CAP_project/blob/95ed52baa29546b260ecb8609f2a1200c967161f/CAP/gap/CategoryConstructor.gi#L329

That's also the reason why the tests are failing. Sorry, I didn't think of this when we talked about it.

zickgraf:
Addendum: That `CategoryConstructor` does not fail with an error is a deliberate choice. This way, adding new filter types does not cause errors. The downside is that renaming filter types as in this PR silently skips operations which were previously installed. However, since this happens very rarely I see no need for action.

sebastianpos:
Thanks for spotting the typo and thanks for the explanation of the missing code. I replaced `IsCyclotomic` in CategoryConstructor by the new type (which I think is appropriate since `IsCyclotomic` was not a proper solution anyway).

zickgraf:
> I replaced `IsCyclotomic` in CategoryConstructor by the new type (which I think is appropriate since `IsCyclotomic` was not a proper solution anyway).

Yes!

PR #854, opened 2022-03-02
Install logical implications in Finalize

zickgraf:
This allows us to suspend/resume method reordering once instead of multiple times per category. The only regression is that this way all categorical properties have to be set before calling Finalize, but setting properties afterwards is a bad idea anyway because then derivations specific to this property cannot be triggered anymore.

This gives a speedup of some 100ms per category, which is particularly noticeable when categories are created during package loading (e.g. TerminalCategory, FinSets, SkeletalFinSets).

@sebastianpos Since the modified code is very old, I would like your general okay for this, but no deep technical review should be required.


sebastianpos:
You have my general ok.

PR #860, opened 2022-03-13
Implementation of Properties of adelman objects

sebastianpos:


zickgraf:
According to the code coverage the newly added methods (e.g. for `IsLeftExact`) are never executed. That's a bit unexpected: either

1. the added example is not executed by `make (ci-)test`,
2. some existing methods are triggered, or
3. the code coverage is broken.

zickgraf:
> Is it a problem that the provided test code is based on `QPA2`?

Aaaah, yes, I didn't notice the line
```
#! #@if IsPackageMarkedForLoading( "QPA", ">= 2.0" )
```

Then everything works as expected (as we do not install QPA2 in the CI), sorry for the false alarm.

sebastianpos:
Okay, thank you for the clarification.

PR #862, opened 2022-03-15
Move tests depending on non-deposited packages to respective packages

zickgraf:
@sebastianpos This moves two examples on spectral sequences appearing in the manual of `CAP` to the manual of `HomologicalAlgebraForCAP`. Is this okay, or are those explicitly wanted in the manual of `CAP`?

Background: https://github.com/homalg-project/homalg_project/issues/465#issuecomment-1066865080


fingolfin (approved):
That would be helpful, thank you!

sebastianpos:
Yes, this is okay. Thank you.

PR #864, opened 2022-03-19
Add new primitive: AdditiveGenerators

sebastianpos:
@zickgraf Unfortunately, this new primitive yields errors in `CompilerForCAP`. Maybe this is the case because `AdditiveGenerators` is the first primitive whose output is a list of objects?

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.04-01",
+Version := "2022.04-02",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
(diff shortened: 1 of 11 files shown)


zickgraf:
The error "this case is not handled yet" comes from `OppositeCategory` and also occurs in the Freyd tests. You probably also have to adjust `WrapperCategory` if this should be available for `CategoryOfColumns`.

All of this should be analogous to https://github.com/homalg-project/CAP_project/pull/851.

sebastianpos:
I fixed the opposite category, but I don't see what I need to do in order to make `AdditiveGenerators` work on `CategoryOfColumns`. `WrapperCategory` seems to rely on `CategoryConstructor`, which in turn should be able to handle `list_of_objects` as an output type?

zickgraf:
Ah, sorry, I confused input and output types. You are right, `CategoryConstructor` would have to handle `list_of_objects` for this. You can give it a try or I can do it when I touch `CategoryConstructor` the next time.

sebastianpos:
In that case: I'm happy if you do it :)

PR #868, opened 2022-03-31
Sum( L, f ) -> Sum( List( L, f ) ) for better compilation

mohamed-barakat:


zickgraf:
Thanks, the result looks much nicer! At some point we can think about converting `Sum(L,f)` to `Sum(List(L,f))` before the compilation and the other way round after the compilation to get the best runtime performance.

I have retriggered the CI because Codecov is having issues.

mohamed-barakat:
These few changes simplified the compiled `FunctorCategories` code a lot. @kamalsaleh.

mohamed-barakat:
See, e.g., [AddHomomorphismStructureOnObjects](https://github.com/mohamed-barakat/FunctorCategories/blob/no_cap_functors/gap/precompiled_categories/FunctorCategoryOfAlgebroidWithRelationsInMatrixCategoryPrecompiled.gi#L1326-L1403).

PR #873, opened 2022-04-06
Prefer the multi-argument version of Concatenation

zickgraf:
This is a preparation for outlining elements of literal lists.


zickgraf:
Codecov is not updating properly, the coverage actually increases.

PR #879, opened 2022-04-09
made IsAdditiveCategory include IsCartesianCategory and IsCocartesianCategory

mohamed-barakat:
This PR is meant for discussion. First of all, it is "nice" that `CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.property` includes a minimal list of operations. However, we use these lists in `__CAP_INTERNAL_RANK_CATEGORY_FROM_METHOD_RECORD` and hence in `InfoStringOfInstalledOperationsOfCategory` to decide which lists include which, and hence which doctrines imply which. But to do the latter properly, we instead need to use the lists of operations derivable from the former lists. So what I am suggesting in this PR is definitely not optimal, but would without bigger changes allow `IsAdditiveCategory` to "imply" `IsCartesianCategory` and `IsCocartesianCategory`. What do you think?


mohamed-barakat:
Together with PR #880:

```gap
gap> LoadPackage( "LinearAlgebra", false );;
gap> Qmat := MatrixCategory( Q );;
gap> InfoOfInstalledOperationsOfCategory( Qmat );
71 primitive operations were used to derive 404 operations for this category which
* IsEquippedWithHomomorphismStructure
* IsLinearCategoryOverCommutativeRing
* IsAbelianCategoryWithEnoughInjectives
* IsRigidSymmetricClosedMonoidalCategory
* IsRigidSymmetricCoclosedMonoidalCategory
```

zickgraf:
This first commit is #880, correct?

I think for the second commit, @sebastianpos has to decide. The clean solution would probably be adding a second argument `implications` to `AddCategoricalProperty` which would also allow us to get rid of all the manual `InstallTrueMethod`s. Maybe if we require the user to handle the transitivity (i.e. provide all implications, not only the immediate ones) this would not be too much work?

mohamed-barakat:
> I think for the second commit, @sebastianpos has to decide. The clean solution would probably be adding a second argument implications to AddCategoricalProperty which would also allow us to get rid of all the manual InstallTrueMethods. Maybe if we require the user to handle the transitivity (i.e. provide all implications, not only the immediate ones) this would not be too much work?

Do you know how `ShowImpliedFilters` works and if it captures the `InstallTrueMethod`-implications?

zickgraf:
> > I think for the second commit, @sebastianpos has to decide. The clean solution would probably be adding a second argument implications to AddCategoricalProperty which would also allow us to get rid of all the manual InstallTrueMethods. Maybe if we require the user to handle the transitivity (i.e. provide all implications, not only the immediate ones) this would not be too much work?
> 
> Do you know how `ShowImpliedFilters` works and if it captures the `InstallTrueMethod`-implications?

I don't, and looking at the code it looks complicated :/

mohamed-barakat:
Now obsolete with PR #894.

PR #880, opened 2022-04-09
simplified InfoStringOfInstalledOperationsOfCategory

mohamed-barakat:
using MaximalObjects and got rid of
* __CAP_INTERNAL_RANK_CATEGORY_FROM_METHOD_RECORD
* CAP_INTERNAL_CATEGORY_PROPERTY_RANK_AND_STRING

### DIFF
--- a/CAP/gap/PrintingFunctions.gi
+++ b/CAP/gap/PrintingFunctions.gi
@@ -3,133 +3,37 @@
 #
 # Implementations
 #
-InstallValue( CAP_INTERNAL_CATEGORY_PROPERTY_RANK_AND_STRING, rec( ) );
-
-BindGlobal( "__CAP_INTERNAL_RANK_CATEGORY_FROM_METHOD_RECORD",
-  function( )
-    local list_of_names, i, current, caps_positions, name_list, j, return_list, entries, added;
-    
-    list_of_names := ShallowCopy( RecNames( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD ) );
-    
-    Remove( list_of_names, Position( list_of_names, "EveryCategory" ) );
-    
-    name_list := [ ];
-    
-    for i in list_of_names do
-        
-        current := ShallowCopy( i );
-        
-        current := current{[ 3 .. Length( current ) ]};
-        
-        if EndsWith( current, "Category" ) then
-            current := current{[ 1 .. Length( current ) - 8 ]};
-        fi;
-        
-        caps_positions := Filtered( [ 1 .. Length( current ) ], j -> current[ j ] in "ABCDEFGHIJKLMNOPQRSTUVWXYZ" );
-        
-        if 1 in caps_positions then
-            Remove( caps_positions, Position( caps_positions, 1 ) );
-        fi;
-        
-        current := LowercaseString( current );
-        
-        for j in Reversed( caps_positions ) do
-            current := Concatenation( current{[ 1 .. j - 1 ]}, " ", current{[ j .. Length( current ) ]} );
-        od;
-        
-        Add( name_list, [ i, current ] );
-        
-    od;
-    
-    return_list := [ ];
-    
-    for i in name_list do
-        
-        entries := CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( i [ 1 ] );
-        
-        added := false;
-        
-        for j in return_list do
-            
-            if ForAll( j, k -> IsSubset( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( k[ 1 ] ), entries )
-                            or IsSubset( entries, CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( k[ 1 ] ) ) ) then
-                
-                Add( j, i );
-                added := true;
-                break;
-                
-            fi;
-            
-        od;
-        
-        if not added then
-            
-            Add( return_list, [ i ] );
-            
-        fi;
-        
-    od;
-    
-    for i in return_list do
-        Sort( i, function( j, k ) return IsSubset( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( j[ 1 ] ), CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( k[ 1 ] ) ); end );
-    od;
-    
-    # Some cosmetics for the lists:
-    # Write Abel with a capital, its the man ;).
-    # Change the order so that the list with Abelian is at the end.
-    
-    for i in return_list do
-        for j in i do
-            j[ 2 ] := ReplacedString( j[ 2 ], "abelian", "Abelian" );
-        od;
-    od;
-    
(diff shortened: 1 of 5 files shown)


mohamed-barakat:
This PR removes the redundancy in the output of `InfoOfInstalledOperationsOfCategory`:

```gap
gap> LoadPackage( "FinSets", false );;
gap> InfoOfInstalledOperationsOfCategory( SkeletalFinSets );
56 primitive operations were used to derive 264 operations for this category which
* IsEquippedWithHomomorphismStructure
* IsElementaryTopos
```

CAP/gap/PrintingFunctions.gi
```diff
@@ -3,133 +3,40 @@
...
     fi;
     
-    prim_operations := ListPrimitivelyInstalledOperationsOfCategory( category );
-    derived_operations := ListInstalledOperationsOfCategory( category );
+    prim_operations := Length( ListPrimitivelyInstalledOperationsOfCategory( category ) );
+    derived_operations := Length( ListInstalledOperationsOfCategory( category ) );
     
-    prim_operations := Length( prim_operations );
-    derived_operations := Length( derived_operations );
+    list_of_properties := ShallowCopy( RecNames( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD ) );
```
zickgraf, line 19:
```suggestion
    list_of_properties := Intersection( ListKnownCategoricalProperties( category ), RecNames( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD ) );
```

and get rid of the next line. (Strictly speaking, this is a bugfix independent of this PR, but I think it makes sense to include it here.)

The issue: Currently the displayed properties are chosen simply by looking at the installed operations. But CAP cannot distinguish for example between a pre-Abelian and an Abelian category simply by looking at the operations. So I think we should always restrict to known categorical properties.

CAP/gap/PrintingFunctions.gi
```diff
@@ -3,133 +3,40 @@
...
+    Sort( list_of_properties, { p1, p2 } -> Length( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p2 ) ) > CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p1 ) );
     
     string := ShallowCopy( String( prim_operations ) );
     Append( string, " primitive operations were used to derive " );
     Append( string, String( derived_operations ) );
-    Append( string, " operations for this category which" );
-    for current_adjective in adjectives do
+    Append( string, " operations for this category" );
+    if not IsEmpty( list_of_properties ) then
+        Append( string, " which" );
```
zickgraf, line 35:
```suggestion
        Append( string, " which constructively" );
```

to make clear that the displayed properties might differ from `ListKnownCategoricalProperties`.

CAP/gap/PrintingFunctions.gi
```diff
@@ -3,133 +3,40 @@
...
-                break;
-            fi;
-        od;
-    od;
+    list_of_properties := Filtered( list_of_properties, p -> CheckConstructivenessOfCategory( category, p ) = [ ] );
+    
+    list_of_properties := MaximalObjects( list_of_properties,
+                                  { p1, p2 } -> IsSubset( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p2 ), CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p1 ) ) );
+    
+    Sort( list_of_properties, { p1, p2 } -> Length( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p2 ) ) > CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p1 ) );
```
zickgraf, line 28:
```suggestion
    StableSortBy( list_of_properties, { p1, p2 } -> Length( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p2 ) ) > CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p1 ) );
```

I think it's better to not allow arbitrary reorderings of two properties with the same number of required operations (`Stable`), and the version with `By` should be faster (although it probably does not matter in this case).

zickgraf (changes requested):
See suggested changes.

CAP/gap/PrintingFunctions.gi
```diff
@@ -3,133 +3,40 @@
...
     
     string := ShallowCopy( String( prim_operations ) );
     Append( string, " primitive operations were used to derive " );
     Append( string, String( derived_operations ) );
```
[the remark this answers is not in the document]
mohamed-barakat: Done.

CAP/gap/PrintingFunctions.gi
mohamed-barakat: Done.

CAP/gap/PrintingFunctions.gi
mohamed-barakat: Done.

zickgraf (changes requested):

CAP/gap/PrintingFunctions.gi
```diff
@@ -3,133 +3,35 @@
...
     
-    for current_stack in __CAP_INTERNAL_RANK_CATEGORY_FROM_METHOD_RECORD() do
-        for current_property in current_stack do
-            if Length( CheckConstructivenessOfCategory( category, current_property[ 1 ] ) ) = 0 then
-                Add( adjectives, current_property[ 1 ] );
-                break;
-            fi;
-        od;
-    od;
+    StableSort( list_of_properties, { p1, p2 } -> Length( CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p2 ) ) > CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.( p1 ) );
```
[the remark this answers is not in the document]
mohamed-barakat: Oh, thanks, fixed.

CAP/gap/PrintingFunctions.gi
mohamed-barakat: This makes `IsEquippedWithHomomorphismStructure` disappear since it is not set by default. I will reset to the first behavior and we can take care of setting `IsEquippedWithHomomorphismStructure` in a later PR.

mohamed-barakat: The idea is that I want to show what can be computed, regardless of whether the property is set or not. Of Course, we can enforce setting the property if all defining operations are available.

mohamed-barakat:
Now all is clear from my side.

CAP/gap/PrintingFunctions.gi
zickgraf: The problem is that `IsAbelianCategory` cannot be checked in this way (and setting the property forcefully would be wrong in general), because there is no operation `IsomorphismFromCoimageToImage` or some equivalent operation which corresponds to being Abelian.

mohamed-barakat:
Is there more to do for this PR?

zickgraf (approved):
I like this PR, the result is so much simpler and better to understand than the previous code :-)

PR #882, opened 2022-04-19
Dual operations and tests for additive monoidal categories

TKuh:
I noticed that there is no mathematical definition of additive monoidal categories in the manual and there is also no doctrine for it. Maybe we can talk about that on Thursday. 

I already have the code for `CategoryConstructor`, `Toposes`, etc. to translate and use the tests, but I would like to wait with PR's until this is merged.

### DIFF
--- a/MonoidalCategories/PackageInfo.g
+++ b/MonoidalCategories/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "MonoidalCategories",
 Subtitle := "Monoidal and monoidal (co)closed categories",
-Version := "2022.03-02",
+Version := "2022.04-01",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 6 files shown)


mohamed-barakat (approved):

mohamed-barakat:
> I noticed that there is no mathematical definition of additive monoidal categories in the manual and there is also no doctrine for it. Maybe we can talk about that on Thursday.

As we just discussed `IsMonoidalCategory` and `IsAdditiveCategory` [is equivalent to](https://math.stackexchange.com/questions/419685/proving-that-tensor-distributes-over-biproduct-in-an-additive-monoidal-category) `IsMonoidalAdditiveCategory`.

zickgraf:
Wouldn't it make sense to use those tests also for `MatrixCategory` and `CategoryOfRows`?

zickgraf:
> > I noticed that there is no mathematical definition of additive monoidal categories in the manual and there is also no doctrine for it. Maybe we can talk about that on Thursday.
> 
> As we just discussed `IsMonoidalCategory` and `IsAdditiveCategory` [is equivalent to](https://math.stackexchange.com/questions/419685/proving-that-tensor-distributes-over-biproduct-in-an-additive-monoidal-category) `IsMonoidalAdditiveCategory`.

If I understand things correctly, this is not true. I assume that with "`IsMonoidalCategory` and `IsAdditiveCategory`" you mean that you do not impose any conditions on the tensor product, and with "`IsMonoidalAdditiveCategory`" you mean that the tensor product distributes over the direct sum? Then the linked post gives a [counterexample](https://math.stackexchange.com/questions/419685/proving-that-tensor-distributes-over-biproduct-in-an-additive-monoidal-category#comment897107_419689) (choose the tensor structure as the (co)cartesian structure e.g. in k-vec). Am I missing something?

mohamed-barakat:
Yes, I assumed that monoidal in Ab respects the Ab-structure.

zickgraf:
> Yes, I assumed that monoidal in Ab respects the Ab-structure.

I see, so to properly reflect this we would have to introduce `IsAbMonoidalCategory`. Probably you have already discussed this.

mohamed-barakat:
According to my assumption no, very much like a functor between additive categories is silently assumed to be additive. But that you both stumbled over this I now think my assumption isn't standard and we should indeed introduce the more general `IsAdditiveMonoidalCategory`.

Issue #886, opened 2022-04-25
Calling `MonoidalCategoriesTest` with `only_primitive_operations := true` fails


TKuh:
I'm trying to add the additive monoidal tests to CategoryOfRows as suggested by @zickgraf in https://github.com/homalg-project/CAP_project/pull/882#issuecomment-1108195977 and also wanted to add calls with the option `only_primitive_operations` as done in other variants of the tests (e.g. [here](https://github.com/homalg-project/FinSetsForCAP/blob/master/tst/CartesianClosedOpposite.tst#L25-L26)). However, when I put such calls with the option `only_primitive_operations := true` first, the following happens:

```gap
gap> LoadPackage( "FreydCategoriesForCAP" );
gap> Q := HomalgFieldOfRationals();;
gap> rows := CategoryOfRows( Q );;
gap> a := 2 / rows;;
gap> b := 3 / rows;;
gap> c := 6 / rows;;
gap> d := 4 / rows;;

gap> alpha := CategoryOfRowsMorphism( a, HomalgMatrix( [ 2 .. 7 ], RankOfObject( a ), RankOfObject( b ), Q ), b );;
gap> beta := CategoryOfRowsMorphism( c, HomalgMatrix( [ 8 .. 31 ], RankOfObject( c ), RankOfObject( d ), Q ), d );;

gap> MonoidalCategoriesTest( rows, a, b, c, alpha, beta : only_primitive_operations := true );;
gap> MonoidalCategoriesTest( rows, a, b, c, alpha, beta );;

gap> a := 0 / rows;;
gap> b := 3 / rows;;

gap> alpha := ZeroMorphism( a, b );;
gap> 
gap> MonoidalCategoriesTest( rows, a, b, c, alpha, beta : only_primitive_operations := true );;
Error, the number of rows has to be equal to the rank of the source at /home/tom/.gap/pkg/CAP_project/FreydCategoriesForCAP/gap/CategoryOfRows.gi:192 called from
CategoryOfRowsMorphism( cat, s, KroneckerMat( UnderlyingMatrix( alpha ), UnderlyingMatrix( beta ) ), r 
 ) at /home/tom/.gap/pkg/CAP_project/FreydCategoriesForCAP/gap/CategoryOfRows.gi:969 called from
CallFuncList( func_to_install, arg ) at /home/tom/.gap/pkg/CAP_project/CAP/gap/InstallAdds.gi:502 called from
[...]

PR #887, opened 2022-04-25
Fix dual_pre_processor_func for `TensorProductOnMorphisms`

TKuh:
Credit to @zickgraf. Was found via [here](https://github.com/homalg-project/CAP_project/issues/886).

### DIFF
--- a/LinearAlgebraForCAP/gap/precompiled_categories/OppositeOfMatrixCategoryPrecompiled.gi
+++ b/LinearAlgebraForCAP/gap/precompiled_categories/OppositeOfMatrixCategoryPrecompiled.gi
@@ -895,7 +895,7 @@ end
 function ( cat_1, s_1, alpha_1, beta_1, r_1 )
     return ObjectifyMorphismWithSourceAndRangeForCAPWithAttributes( rec(
            ), cat_1, s_1, r_1, Opposite, ObjectifyMorphismWithSourceAndRangeForCAPWithAttributes( rec(
-             ), OppositeCategory( cat_1 ), Opposite( s_1 ), Opposite( r_1 ), UnderlyingMatrix, KroneckerMat( UnderlyingMatrix( Opposite( alpha_1 ) ), UnderlyingMatrix( Opposite( beta_1 ) ) ) ) );
+             ), OppositeCategory( cat_1 ), Opposite( r_1 ), Opposite( s_1 ), UnderlyingMatrix, KroneckerMat( UnderlyingMatrix( Opposite( alpha_1 ) ), UnderlyingMatrix( Opposite( beta_1 ) ) ) ) );
 end
 ########
         
--- a/MonoidalCategories/PackageInfo.g
+++ b/MonoidalCategories/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "MonoidalCategories",
 Subtitle := "Monoidal and monoidal (co)closed categories",
-Version := "2022.04-03",
+Version := "2022.04-04",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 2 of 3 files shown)


zickgraf:
I think this pattern (keep the order of the actual arguments but swap the given source and range) occurs at many places, right? We could introduce a switch `dual_with_given_objects_reversed` (or even make this the default for operations with given source and range).

TKuh:
Oh yes, it occurs _very_ often. Such a switch would be very helpful (if it doesn't bloat the method records, it seems to me that they can become already quite chaotic?).

zickgraf:
> Oh yes, it occurs _very_ often. Such a switch would be very helpful (if it doesn't bloat the method records, it seems to me that they can become already quite chaotic?).

It's not more bloat than having two write a custom preprocessor function I think :D I suggest we implement this together during the next meeting, so you see the relevant code and can implement similar changes yourself in the future. Could you check if this actually occurs for *all* operations with `with_given_object_position := "both"` (both inside and outside of `MonoidalCategories`)? If yes, we could just make this the default.

And another thing: if the dual version of the tensor product would swap its arguments, we could simply use `dual_arguments_reversed`. What would be the implications of doing so? Would it make the code at other places (e.g. the internal hom adjunctions) easier or more complicated?

mohamed-barakat:
For me, this PR is ready for merge. Is there more discussion needed before merging?

zickgraf:
> For me, this PR is ready for merge. Is there more discussion needed before merging?

It's ready to merge, my suggestion should be part of a separate PR.

mohamed-barakat:
@TKuh: Is there anything analogous that must be done for `Toposes`?

TKuh:
> > Oh yes, it occurs _very_ often. Such a switch would be very helpful (if it doesn't bloat the method records, it seems to me that they can become already quite chaotic?).
> 
> It's not more bloat than having two write a custom preprocessor function I think :D 

Touché. :)

> I suggest we implement this together during the next meeting, so you see the relevant code and can implement similar changes yourself in the future. Could you check if this actually occurs for _all_ operations with `with_given_object_position := "both"` (both inside and outside of `MonoidalCategories`)? If yes, we could just make this the default.

Yes, that might be very helpful for the future. I'll look into it.

> And another thing: if the dual version of the tensor product would swap its arguments, we could simply use `dual_arguments_reversed`. What would be the implications of doing so? Would it make the code at other places (e.g. the internal hom adjunctions) easier or more complicated?

That would be mathematically wrong. :(
The dual of 
```
        f x g
a x b ----------> c x d
```
should be
```
          (f x g)^op
a x b <----------------- c x d
```
and not
```
           (g x f)^op
b x a <----------------- d x c
```
At least if we want `Opposite` to behave in the way one would expect it mathematically. And I think a couple of months ago when all of this dualization work started (left/right internal (co)homs etc.) we agreed, that opposite should work the way it does now.

TKuh:
> @TKuh: Is there anything analogous that must be done for `Toposes`?

Yes, the same mistake happens with `DirectProductOnMorphismsWithGivenDirectProducts` and `CoproductOnMorphismsWithGivenCoproducts`. I will fix it.

TKuh:
> > Oh yes, it occurs _very_ often. Such a switch would be very helpful (if it doesn't bloat the method records, it seems to me that they can become already quite chaotic?).
> 
> It's not more bloat than having two write a custom preprocessor function I think :D I suggest we implement this together during the next meeting, so you see the relevant code and can implement similar changes yourself in the future. Could you check if this actually occurs for _all_ operations with `with_given_object_position := "both"` (both inside and outside of `MonoidalCategories`)? If yes, we could just make this the default.

Ok, so I checked in the method records of `CAP/gap/MethodRecord.gi` and in `MonoidalCategories` how the operations with `'with_given_object_position := "both"` handle the dual operation and actually almost all the time `dual_arguments_reversed` is used (instead of a custom dual_pre/post_function).
For `MonoidalCategories` the only exceptions where dual_pre_functions are used, are

- TensorProductOnMorphisms
- The associators
- The compatibility morphisms (really not surprising, they are complicated)
- MorphismFromInternalHomToTensorProduct
- MorphismFromTensorProductToInternalCoHom

For CAP, the exceptions are

- HomomorphismStructureOnMorphisms
- InterpretMorphismAsMorphismFromDistinguishedObjectToHomomorphismStructure
- MorphismBetweenDirectSums

So I take back my initial enthusiasm; I don't think a switch `dual_with_given_objects_reversed` is necessary. In almost all cases one can use `dual_arguments_reversed`. The monoidal operations above are really only special cases, where reversing all arguments doesn't work.

PR #889, opened 2022-04-26
Only list known categorical properties in InfoOfInstalledOperationsOfCategory

zickgraf:
Otherwise any constructively pre-Abelian category which can compute LiftAlongMonomorphism and ColiftAlongEpimorphism will be displayed as Abelian, which is wrong in general.

Of course this means that we have to explicitly set the corresponding properties, but I see no way around this.


zickgraf:
As just discussed verbally, I will merge this and the properties will have to be set manually.

mohamed-barakat (approved):

PR #896, opened 2022-05-03
Apply PackageJanitor

zickgraf:


zickgraf:
The coverage drops because
1. Codecov is not updating properly and
2. the example `GradedModulePresentationsForCAP/examples/CohP1.g` runs without `FreydCategoriesForCAP`, which leads to different derivations.

PR #902, opened 2022-05-12
Simplify passing a type signature to CapJitCompiledFunction

zickgraf:


zickgraf:
Codecov is does update again....

PR #906, opened 2022-05-17
Rename IsIdenticalToIdentity/ZeroMorphism to IsEqualToIdentity/ZeroMorphism and deprecate CokernelFunctorial in favor of CokernelObjectFunctorial

zickgraf:
@sebastianpos Are those two changes okay? (You do not have to look at the code, there is nothing special.)

### DIFF
--- a/CAP/LogicForCategories/PropositionsForAbCategories.tex
+++ b/CAP/LogicForCategories/PropositionsForAbCategories.tex
@@ -6,7 +6,7 @@
 
 \begin{sequent}
 \begin{align*}
-   a: \Obj, b: \Obj ~|~ () \vdash \IsIdenticalToZeroMorphism \big( \ZeroMorphism( a, b ) \big)
+   a: \Obj, b: \Obj ~|~ () \vdash \IsEqualToZeroMorphism \big( \ZeroMorphism( a, b ) \big)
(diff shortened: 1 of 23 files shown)


sebastianpos:
Yes, I'm ok with the renaming. I cannot find a reason why it was `IsIdentical` instead of `IsEqual`

zickgraf:
> Yes, I'm ok with the renaming. I cannot find a reason why it was `IsIdentical` instead of `IsEqual`

I assume because `IsEqualForMorphisms` and `IsCongruentForMorphisms` were previously called `IsIdenticalForMorphisms` and `IsEqualForMorphisms`: 427e224b4c8e8ab71d7ae2de823c31e6ee2b66ee.

zickgraf:
Addendum: Thanks for the fast reply!

sebastianpos:
> I assume because `IsEqualForMorphisms` and `IsCongruentForMorphisms` were previously called `IsIdenticalForMorphisms` and `IsEqualForMorphisms`: [427e224](https://github.com/homalg-project/CAP_project/commit/427e224b4c8e8ab71d7ae2de823c31e6ee2b66ee).

This is a good explanation. Thanks.

PR #913, opened 2022-05-23
host TerminalCategoryWithMultipleObjects

mohamed-barakat:
which was outsourced from CategoryConstructor v2022.05-05

this constructor should be unified with CAP's terminal category

### DIFF
--- a/MonoidalCategories/PackageInfo.g
+++ b/MonoidalCategories/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "MonoidalCategories",
 Subtitle := "Monoidal and monoidal (co)closed categories",
-Version := "2022.05-03",
+Version := "2022.05-04",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
--- a/MonoidalCategories/gap/TerminalCategory.gd
+++ b/MonoidalCategories/gap/TerminalCategory.gd
@@ -1,0 +1,40 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+# CategoryConstructor: Construct categories out of given ones
+#
+# Declarations
+#
+
+#! @Chapter The category constructor
+
+####################################
+#
+#! @Section Example
+#
+####################################
+
+#! This is an example of a category which is created using <C>CategoryConstructor</C>
+#! out of no input.
+
+#! This category <Q>lies</Q> in all doctrines and can hence
+#! be used (in conjunction with <C>LazyCategory</C>)
+#! in order to check the type-correctness of the various derived methods
+#! provided by &CAP; or any &CAP;-based package.
+
+#! @Description
+#!  The &GAP; type of a terminal category with multiple objects.
+#! @Arguments T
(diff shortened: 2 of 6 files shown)


zickgraf:
From looking at the code coverage I have two questions:

1. Why do the existing tests (e.g. the ones using MatrixCategory/CategoryOfRows) not include `AdditiveMonoidalCategoriesTest`?
2. Why can't the terminal category compute `LeftDistributivityExpanding` et al.?

zickgraf:
> > 1. Why do the existing tests (e.g. the ones using MatrixCategory/CategoryOfRows) not include `AdditiveMonoidalCategoriesTest`?
> 
> I don't understand the question. `AdditiveMonoidalCategoriesTest` is currently only used in `MonoidalCategories/tst/TerminalCategoryMonoidal.tst` within the CAP_project repository.

Exactly, but I think it would be natural to also include `AdditiveMonoidalCategoriesTest` in the tests of MatrixCategory/CategoryOfRows, or am I missing something?

mohamed-barakat:
I don't see a problem. Maybe you can reference your suggestion in a new issue.

mohamed-barakat:
> 2. Why can't the terminal category compute `LeftDistributivityExpanding` et al.?

`TerminalCategoryWithMultipleObjects`() lies in the filter `IsAdditiveCategory`. So this is weird.

Note: The corresponding methods for [distributive cartesian categories](https://github.com/homalg-project/CAP_project/blob/master/CartesianCategories/gap/CartesianCategoriesDerivedMethods_extra.gi#L148-L164) and for [codistributive cocartesian categories](https://github.com/homalg-project/CAP_project/blob/master/CartesianCategories/gap/CocartesianCategoriesDerivedMethods_extra.gi#L148-L164) are installed.

zickgraf:
> > 2. Why can't the terminal category compute `LeftDistributivityExpanding` et al.?
> 
> `TerminalCategoryWithMultipleObjects`() lies in the filter `IsAdditiveCategory`. So this is weird.
> 
> Note: The corresponding methods for [distributive cartesian categories](https://github.com/homalg-project/CAP_project/blob/master/CartesianCategories/gap/CartesianCategoriesDerivedMethods_extra.gi#L148-L164) and for [codistributive cocartesian categories](https://github.com/homalg-project/CAP_project/blob/master/CartesianCategories/gap/CocartesianCategoriesDerivedMethods_extra.gi#L148-L164) are installed.

You analysis in https://github.com/homalg-project/CAP_project/pull/913#issuecomment-1134660221 is correct. The difference to the situation for cartesian categories is that instead of the non-WithGiven versions the WithGiven versions (including the object) are listed here: https://github.com/homalg-project/CAP_project/blob/001bb26c300ad07fae89fad42dbe48efb4f634b5/CartesianCategories/gap/CartesianCategoriesProperties.gd#L16-L18

The correct solution is the following: `TerminalCategoryWithMultipleObjects` has to replace non-WithGiven operations by WithGiven operations and the corresponding objects in `list_of_operations_to_install` here: https://github.com/mohamed-barakat/CAP_project/blob/accef9d6e869737ce900cdc84c35704a7943a1ae/MonoidalCategories/gap/TerminalCategory.gi#L80 Hopefully this is something @TKuh tackles anyway while unifying `TerminalCategory` and `TerminalCategoryWithMultipleObjects`.

PR #914, opened 2022-05-23
enhanced LaTeXOutput for IsFreydCategoryMorphism

mohamed-barakat:
to establish an analogy to the corresponding method for IsCategoryOfRowsMorphism


sebastianpos:
I'm confused. You describe this PR as if you were changing the LaTeX output for `IsFreydCategoryMorphism`, but what you do is to change the LaTeX output for `IsCategoryOfRowsMorphism`. Am I missing something?

mohamed-barakat:
I changed the wrong comment in the commit message and force-pushed.

sebastianpos:
I'm ok with the coloring.

PR #919, opened 2022-05-25
Add workaround for https://github.com/oscar-system/GAP.jl/issues/814

zickgraf:


mohamed-barakat:
Wonderful, many thanks.

PR #922, opened 2022-05-27
Introduce compatible_with_congruence_of_morphisms and require compatibility for some operations

zickgraf:
@sebastianpos This addresses some concerns from my email. The new method record component `compatible_with_congruence_of_morphisms` indicates if the operation is compatible with the congruence of morphisms, that is, if its output does not change with regard to `IsEqualForObjects` and `IsCongruentForMorphisms` if the input changes with regard to `IsEqualForObjects` and `IsCongruentForMorphisms`. If only compatible operations occur in some piece of code, I know that I can use transformations which are only valid up to congruence (instead of equality) as long as I'm only interested in the result up to congruence.

In the documentation, I now explicitly require addition, ring multiplication, and the tensor structure to be compatible with the congruence of morphisms. Are you okay with those requirements?

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.05-08",
+Version := "2022.05-09",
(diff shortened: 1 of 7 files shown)


sebastianpos:
Yes, that is ok.

PR #924, opened 2022-05-29
Import CartesianCategories into CAP

mohamed-barakat:


mohamed-barakat:
Once this is merged I can start a monorepo for the other packages. In particular, I need need to know I did the `git subtree add ...` correctly.

mohamed-barakat:
I now also added two test files:
* `TerminalCategoryCartesian.tst`
* `TerminalCategoryCocartesian.tst`

mohamed-barakat:
Currently, codecov doesn't seem to cover the subdirectory `CartesianCategories/gap`.

zickgraf:
> Once this is merged I can start a monorepo for the other packages. In particular, I need to know if I did the `git subtree add ...` correctly.

Looks fine, just keep in mind that we have to push the result directly instead of clicking "Merge pull request".



> Currently, codecov doesn't seem to cover the subdirectory `CartesianCategories/gap`:
> 
> https://app.codecov.io/gh/homalg-project/CAP_project/compare/924/tree/CartesianCategories

Yes, because the tests are no executed. You have to add suitable targets to `test` and `ci-test` in the main makefile manually. (The reason why I cannot do this automatically is that homalg_project has some special cases there.)

Some other small things:

1. Please sort the packages alphabetically, i.e. add `CartesianCategories` after `AttributeCategoryForCAP`, in PackageJanitor (the only exception is `CAP` as the central package).
2. In the changes to `Tests.yml` and `release-gap-package` you have to include the prefix `CAP_project/` before `CartesianCategories`.
3. Please remove all files inside `CartesianCategories` which are not needed anymore, i.e., `ls CartesianCategories` and `ls MonoidalCategories` should yield the same result.

mohamed-barakat:
Hopefully, all done.

zickgraf:
A note for future imports: One also has to commit and push the changes PackageJanitor does to gh-pages: https://github.com/homalg-project/CAP_project/commit/c710d2e7492acda8d0e5dcf9395ec93ebb6447a8

PR #932, opened 2022-06-12
Compile more operations in ModulePresentationsForCAP

TKuh:
For now, these are only the low-hanging fruits.

Some remarks:
- The compiled `AddBraiding` for right presentations inverts the permutation matrix, while the original code inverted the permutation directly before creating a matrix out of it.
- `TensorProductOnObjects` is compilable but the compiled verison makes the example `ProjectivityTest.g` fail.
- `PreCompose` has special cases eliminated when compiled, hence I did not include it for now.
- `IsEqualForObjects` has an additional check for the dimensions of the matrices, when compiled. Hence I did not include it for now.

These are easy to discuss and to add. They also don't impact the tests.

Everything else is very different when compiled. Especially the various lifts and the internal homs are complicated.


zickgraf:
Thanks for this start! I guess it's best we discuss this verbally during our next meeting.

zickgraf:
Results of our discussion:

>   The compiled `AddBraiding` for right presentations inverts the permutation matrix, while the original code inverted the permutation directly before creating a matrix out of it.

This is actually faster than the old version -> use it :-)

>   `TensorProductOnObjects` is compilable but the compiled verison makes the example `ProjectivityTest.g` fail.

Changing `ProjectivityTest.g` is okay, but we have to make sure the implementation is compatible with the old `InternalHom...`

>   `PreCompose` has special cases eliminated when compiled, hence I did not include it for now.

We can precompile the standard version and keep the special cases on top.

>   `IsEqualForObjects` has an additional check for the dimensions of the matrices, when compiled. Hence I did not include it for now.

This additional check should be removed in CategoryOfRows because any matrix comparison algorithm should do this anyway.

Notes for the internal hom:
1. `INTERNAL_HOM_EMBEDDING_IN_TENSOR_PRODUCT_LEFT` and `INTERNAL_HOM_EMBEDDING` seem to agree on a high-level (up to the change in `TensorProductOnObjects`?)
2. However, they both end with `KernelEmbedding`, which differs between ModulePresentations and FreydCategory.
3. The difference can be seen in `IntrinsicModules/gap/FreydCategory.gi`: `ProjectionOfBiased*Relative*WeakFiberProduct` is needed to obtain the old performance.

zickgraf:
Addendum (also already discussed):

1. The old code should be removed.
2. Please also test HigherHomologicalAlgebra and look for the timings.

TKuh:
Sorry, that this took such a long time. I'll take care of it now.

> > The compiled `AddBraiding` for right presentations inverts the permutation matrix, while the original code inverted the permutation directly before creating a matrix out of it.
> 
> This is actually faster than the old version -> use it :-)

Great. :)

> > `TensorProductOnObjects` is compilable but the compiled verison makes the example `ProjectivityTest.g` fail.
> 
> Changing `ProjectivityTest.g` is okay, but we have to make sure the implementation is compatible with the old `InternalHom...`

I'm having some trouble with this. I remember, that the compiled `TensorProductOnObjects` changed the order of a `UnionOfRows`, which is hence different, and since we use $\underline{\text{Hom}}(a,b) \:= a^\vee \otimes b$, we have to make sure, that `InternalHomOn*` is not badly affected by this. Is this the problem? If yes, what remains to be checked here? Would it be ok to just check, if the runtime and tests are not influenced by this (also of other packages)?

> > `PreCompose` has special cases eliminated when compiled, hence I did not include it for now.
>
> We can precompile the standard version and keep the special cases on top.
> 

I understand this as "remove only the first case in `AddPreCompose` but keep the rest". Is that correct?

> > `IsEqualForObjects` has an additional check for the dimensions of the matrices, when compiled. Hence I did not include it for now.
> 
> This additional check should be removed in CategoryOfRows because any matrix comparison algorithm should do this anyway.
> 

I searched CategoryOfRows (and FreydCategory) for such a check, but I can't find the place, where [this](https://github.com/homalg-project/CAP_project/blob/5145597947b3a8d7bc26d51796cf173a421e75ce/ModulePresentationsForCAP/gap/precompiled_categories/LeftPresentationsAsFreydCategoryOfCategoryOfRowsOfArbitraryRingPrecompiled.gi#L163) originates. :(

>     1. The old code should be removed.

I did that.

>     2. Please also test HigherHomologicalAlgebra and look for the timings.

Here are the runtimes I get with `HigherHomologicalAlgebra` using the master branch, the branch from this PR, and this PR plus a compiled `TensorProductOnObjects`. Doesn't look too bad to me (even though it's a bit slower), and the tests run fine. What is your opinion?

|  Package   | CAP master        | module_pres_compiled           | additionally compile `TensorProductOnObjects`  |
| ------------- |-------------| -----| ------|
| BBGG                                              | total      1341 ms (270 ms GC) and 507MB allocated| total      1392 ms (277 ms GC) and 513MB allocated |   total      1451 ms (288 ms GC) and 507MB allocated    |
| Bicomplexes                                     |   total      1301 ms (279 ms GC) and 466MB allocated  |   total      1311 ms (293 ms GC) and 467MB allocated |   total      1361 ms (281 ms GC) and 472MB allocated    |
| ComplexesCategories | total     23175 ms (2020 ms GC) and 4.07GB allocated |  total     25975 ms (2178 ms GC) and 4.06GB allocated |   total     27428 ms (2254 ms GC) and 4.08GB allocated    |
| DerivedCategories                            | total     64707 ms (5006 ms GC) and 8.70GB allocated | total     99906 ms (6492 ms GC) and 14.1GB allocated|   total     98703 ms (6517 ms GC) and 13.8GB allocated  |
| HomotopyCategories                        | total      2490 ms (399 ms GC) and 740MB allocated | total      2790 ms (439 ms GC) and 741MB allocated  |    total      3070 ms (481 ms GC) and 741MB allocated   |
| QuotientCategories                           | total      2505 ms (363 ms GC) and 541MB allocated | total      2193 ms (357 ms GC) and 536MB allocated |    total      2471 ms (400 ms GC) and 542MB allocated    |
| StableCategories                              | total     19967 ms (2837 ms GC) and 2.36GB allocated | total     19603 ms (2890 ms GC) and 2.33GB allocated |    total     21786 ms (3200 ms GC) and 2.33GB allocated   |
| ToolsForHigherHomologicalAlgebra | total      1454 ms (312 ms GC) and 459MB allocated  | total      1280 ms (275 ms GC) and 459MB allocated |   total      1585 ms (344 ms GC) and 459MB allocated    |
| TriangulatedCategories                     | total      6954 ms (697 ms GC) and 912MB allocated | total      6542 ms (707 ms GC) and 906MB allocated |   total      7651 ms (849 ms GC) and 912MB allocated    |

zickgraf:
> Sorry, that this took such a long time. I'll take care of it now.

No worries!

> > > `TensorProductOnObjects` is compilable but the compiled verison makes the example `ProjectivityTest.g` fail.
> > 
> > 
> > Changing `ProjectivityTest.g` is okay, but we have to make sure the implementation is compatible with the old `InternalHom...`
> 
> I'm having some trouble with this. I remember, that the compiled `TensorProductOnObjects` changed the order of a `UnionOfRows`, which is hence different, and since we use Hom―(a,b):=a∨⊗b, we have to make sure, that `InternalHomOn*` is not badly affected by this.

I don't think this is true: InternalHomOnObjects/Morphisms is primitively installed. So there is no way around checking that this implementation is compatible with changing the order of the `UnionOfRows` :/

> > > `PreCompose` has special cases eliminated when compiled, hence I did not include it for now.
> > 
> > 
> > We can precompile the standard version and keep the special cases on top.
> 
> I understand this as "remove only the first case in `AddPreCompose` but keep the rest". Is that correct?

Yes!
 
> > > `IsEqualForObjects` has an additional check for the dimensions of the matrices, when compiled. Hence I did not include it for now.
> > 
> > 
> > This additional check should be removed in CategoryOfRows because any matrix comparison algorithm should do this anyway.
> 
> I searched CategoryOfRows (and FreydCategory) for such a check, but I can't find the place, where [this](https://github.com/homalg-project/CAP_project/blob/5145597947b3a8d7bc26d51796cf173a421e75ce/ModulePresentationsForCAP/gap/precompiled_categories/LeftPresentationsAsFreydCategoryOfCategoryOfRowsOfArbitraryRingPrecompiled.gi#L163) originates. :(

This is unclear to me right now -> keep it in mind for now.

> > ```
> > 2. Please also test HigherHomologicalAlgebra and look for the timings.
> > ```
> 
> Here are the runtimes I get with `HigherHomologicalAlgebra` using the master branch, the branch from this PR, and this PR plus a compiled `TensorProductOnObjects`. Doesn't look too bad to me (even though it's a bit slower), and the tests run fine. What is your opinion?

Doesn't look bad, but there seems to be something systematically going on. Can you pinpoint which operations are slower? Either by commenting things in or out, or by using timing statistics? (search for "Timing statistics" in the documentation of CAP)

TKuh:
> > 
> > I'm having some trouble with this. I remember, that the compiled `TensorProductOnObjects` changed the order of a `UnionOfRows`, which is hence different, and since we use Hom―(a,b):=a∨⊗b, we have to make sure, that `InternalHomOn*` is not badly affected by this.
> 
> I don't think this is true: InternalHomOnObjects/Morphisms is primitively installed. So there is no way around checking that this implementation is compatible with changing the order of the `UnionOfRows` :/

There is one problem I can spot. [`InternalHomOnObjects`](https://github.com/homalg-project/CAP_project/blob/f6542df30aa5ad3bec2dc8fbca2ca875874eb2eb/ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi#L811-L817) mainly calls [`INTERNAL_HOM_EMBEDDING_IN_TENSOR_PRODUCT_LEFT`](https://github.com/homalg-project/CAP_project/blob/master/ModulePresentationsForCAP/gap/ModulePresentationObject.gi#L217-L245). Now, if we use the compiled `TensorProductOnObjects` (which secretly switches the order of the arguments of the internally used `UnionOfRows`), then the source and range of [this](https://github.com/homalg-project/CAP_project/blob/f6542df30aa5ad3bec2dc8fbca2ca875874eb2eb/ModulePresentationsForCAP/gap/ModulePresentationObject.gi#L238-L241) morphism actually change. Now, I don't know if this morphism is then still the correct one. Maybe the `KernelEmbedding` [called](https://github.com/homalg-project/CAP_project/blob/f6542df30aa5ad3bec2dc8fbca2ca875874eb2eb/ModulePresentationsForCAP/gap/ModulePresentationObject.gi#L243) afterwards doesn't mind this, but I don't see why.

If it does make a difference, then I guess one could either switch the arguments of the calls to `TensorProductOnObjects` or somehow change [this](https://github.com/homalg-project/CAP_project/blob/f6542df30aa5ad3bec2dc8fbca2ca875874eb2eb/ModulePresentationsForCAP/gap/ModulePresentationObject.gi#L230) Kronecker product?

`InternalHomOnMorphisms` also uses [`INTERNAL_HOM_EMBEDDING_IN_TENSOR_PRODUCT_LEFT`](https://github.com/homalg-project/CAP_project/blob/master/ModulePresentationsForCAP/gap/ModulePresentationsForCAP.gi#L843-L864),
so this might have a similar problem.

Are these algorithm written down somewhere? I couldn't find it in Sebastians PhD thesis. I would like to understand it mathematically, to see, if my talk is garbage. :)

zickgraf:
> If it does make a difference, then I guess one could either switch the arguments of the calls to `TensorProductOnObjects` or somehow change [this](https://github.com/homalg-project/CAP_project/blob/f6542df30aa5ad3bec2dc8fbca2ca875874eb2eb/ModulePresentationsForCAP/gap/ModulePresentationObject.gi#L230) Kronecker product?

Have you checked what the corresponding code in FreydCategory does? If it also differs exactly at this point, you can just port the change over and check that it does not affect the timings.

> Are these algorithm written down somewhere? I couldn't find it in Sebastians PhD thesis. I would like to understand it mathematically, to see, if my talk is garbage. :)

They should be here: https://arxiv.org/abs/1909.00172.

TKuh:
> Have you checked what the corresponding code in FreydCategory does? If it also differs exactly at this point, you can just port the change over and check that it does not affect the timings.

I can't prove if the code in `ModulePresentationsForCAP` is really equivalent to Freyd, but I'll explain the similarities I see.

There is this [comment](https://github.com/homalg-project/CAP_project/blob/99154c4d02ef76b9667caf6df3d314f0b92b7748/FreydCategoriesForCAP/gap/FreydCategory.gi#L1515-L1529) in Freyd which explains `InternalHomOnObjects`. The strategy is to construct a [source](https://github.com/homalg-project/CAP_project/blob/ab7d994ed26eb82dba50ba856b6f9945d620783e/FreydCategoriesForCAP/gap/FreydCategory.gi#L346-L348), a [range](https://github.com/homalg-project/CAP_project/blob/ab7d994ed26eb82dba50ba856b6f9945d620783e/FreydCategoriesForCAP/gap/FreydCategory.gi#L351-L353) and finally a [morphism](https://github.com/homalg-project/CAP_project/blob/ab7d994ed26eb82dba50ba856b6f9945d620783e/FreydCategoriesForCAP/gap/FreydCategory.gi#L356-L357) by using `InternalHomOnMorphisms` three times in the underlying category, i.e. in `CategoryOfRows` for our case. It seems like the same strategy is used in `ModulePresentationsForCAP`

Now, in `CategoryOfRows` the `InternalHomOnMorphisms(a,b)` is eventually derived as

```gap
TensorProductOnMorphisms( CategoryOfRowsMorphism( TransposedMatrix( UnderlyingMatrix( a ) ) ), b ), 
```
I purposely leave out some details here in case you dig through the derivations as well. I only want to emphasize the final matrix operations.

Now this in turn boils down to:

```gap
KroneckerMat( TransposedMatrix( UnderlyingMatrix( a ) ), UnderlyingMatrix( b ) )
```

This is probably what happens in the construction of `differential_matrix` here:
https://github.com/homalg-project/CAP_project/blob/ab7d994ed26eb82dba50ba856b6f9945d620783e/ModulePresentationsForCAP/gap/ModulePresentationObject.gi#L224-L230

And I guess it corresponds to this line in Freyd:
https://github.com/homalg-project/CAP_project/blob/ab7d994ed26eb82dba50ba856b6f9945d620783e/FreydCategoriesForCAP/gap/FreydCategory.gi#L356-L357

So this part seems fine.

On the other hand, the source and range of the constructed `differential` morphism in `ModulePresentationsForCAP` don't seem to be analogous to the source and range in Freyd:
https://github.com/homalg-project/CAP_project/blob/ab7d994ed26eb82dba50ba856b6f9945d620783e/ModulePresentationsForCAP/gap/ModulePresentationObject.gi#L234-L241 https://github.com/homalg-project/CAP_project/blob/ab7d994ed26eb82dba50ba856b6f9945d620783e/FreydCategoriesForCAP/gap/FreydCategory.gi#L346-L353

I don't understand why these `InternalHomOnMorphism`'s are translated into `TensorProductOnObjects`'s and not into 
 `TensorProductOnMorphism`'s. Without this understanding, I don't know if and how this needs to change, when the compiled tensor product is used...

mohamed-barakat:
As discussed here are the changes I made some time ago in `IntrinsicModules` to increase the similarity between `FreydCategory` and `ModulePresentation`:

https://github.com/homalg-project/IntrinsicModules/blob/master/gap/FreydCategory.gi

TKuh:
Some remarks after adding logic templates:

- The most part of `(Co)EvaluationMorphism` is now fairly readable. This was the last missing piece for the compilation of the closed structure. But it is still not easy to check, if it's truly compatible with the change of the tensor product.
- @mohamed-barakat remarked, that some logic templates are likely bad and need enhancement. Some are marked with `TODO`.
- I suspect I made a mistake somewhere in the logic templates, because the sudden change in e.g. [these numbers](https://github.com/homalg-project/CAP_project/pull/932/commits/86123704f5aa8d60047d3705e1ecc1357b984043#diff-a9a59588544cd782f888456a63312512e44f9c38478bb39590a4ad5775e42b16R40-R41) looks suspicious to me. :/

But other than that, this should be ready for a first review.

zickgraf:
Some points:

1.

> * The most part of `(Co)EvaluationMorphism` is now fairly readable. This was the last missing piece for the compilation of the closed structure. But it is still not easy to check, if it's truly compatible with the change of the tensor product.

Sorry, I don't understand this comment. With "the change of the tensor product" you mean using the Freyd version of the tensor product (compiled), right? If you also compile the closed structure from Freyd, this is certainly compatible with the Freyd tensor product, assuming that the Freyd implementation and the logic templates are correct. The only thing we should check for performance reasons is how the manual implementation of `(Co)EvaluationMorphism` compares to the compiled version. Have you tried doing that?

2.

> * I suspect I made a mistake somewhere in the logic templates, because the sudden change in e.g. [these numbers](https://github.com/homalg-project/CAP_project/pull/932/commits/86123704f5aa8d60047d3705e1ecc1357b984043#diff-a9a59588544cd782f888456a63312512e44f9c38478bb39590a4ad5775e42b16R40-R41) looks suspicious to me. :/ They appear after `LambdaElimination`, which eventually boils down to `EvaluationMorphismWithGivenSource`, and this uses a lot of the `LeftDivide` and `ReducedSyzygies` templates.

Do those changes disappear when compiling without the logic templates? If yes, you can bisect your logic templates to find the offending one.

3. You compile `KernelEmbedding` although it differs from the manual version. Is this deliberate?
4. You resolve `INTERNAL_HOM_EMBEDDING` (which uses `KernelEmbedding`), is this deliberate? This might explain the differing numbers.

TKuh:
> Sorry, I don't understand this comment. With "the change of the tensor product" you mean using the Freyd version of the tensor product (compiled), right?

Yes.

> If you also compile the closed structure from Freyd, this is certainly compatible with the Freyd tensor product, assuming that the Freyd implementation and the logic templates are correct. 

Ok, great!

> The only thing we should check for performance reasons is how the manual implementation of `(Co)EvaluationMorphism` compares to the compiled version. Have you tried doing that?
 
Compiled version:
```gap
gap> ZZ := HomalgRingOfIntegers();;
gap> rows := CategoryOfRows( ZZ );;
gap> lpres := LeftPresentations( ZZ );;

gap> mors := List([1 .. 6], i -> AsLeftPresentation( MorphismDatum( RandomMorphism( rows, 5 ) ) ) );;
gap> EnableTimingStatistics( lpres );;

gap> for pair in Tuples( [1..6], 2 ) do EvaluationMorphism( mors[pair[1]], mors[pair[2]] ); od;
gap> for pair in Tuples( [1..6], 2 ) do CoevaluationMorphism( mors[pair[1]], mors[pair[2]] ); od;

gap> DisplayTimingStatistics( lpres );
####
Timing statistics for the primitive operations of the category Category of left presentations of Z:
Total time spent in primitive operations of this category: 43671 ms
CoevaluationMorphismWithGivenRange was called 36 times with a total runtime of 7620 ms ( = 
211666 μs per execution)
EvaluationMorphismWithGivenSource was called 36 times with a total runtime of 1587 ms ( = 44083 μs per execution)
InternalHomOnObjects was called 72 times with a total runtime of 34441 ms ( = 478347 μs per execution)
TensorProductOnObjects was called 72 times with a total runtime of 23 ms ( = 319 μs per execution)
```

Manual version:
```gap
gap> ZZ := HomalgRingOfIntegers();;
gap> rows := CategoryOfRows( ZZ );;
gap> lpres := LeftPresentations( ZZ );;
gap> 
gap> mors := List([1 .. 6], i -> AsLeftPresentation( MorphismDatum( RandomMorphism( rows, 5 ) ) ) );;
gap> EnableTimingStatistics( lpres );;
gap> 
gap> for pair in Tuples( [1..6], 2 ) do EvaluationMorphism( mors[pair[1]], mors[pair[2]] ); od;
gap> for pair in Tuples( [1..6], 2 ) do CoevaluationMorphism( mors[pair[1]], mors[pair[2]] ); od;
gap> 
gap> DisplayTimingStatistics( lpres );
####
Timing statistics for the primWithoutitive operations of the category Category of left presentations of Z:
Total time spent in primitive operations of this category: 16715 ms
BraidingWithGivenTensorProducts was called 72 times with a total runtime of 24 ms ( = 333 μs per execution)
CoevaluationMorphismWithGivenRange was called 36 times with a total runtime of 1010 ms ( = 
28055 μs per execution)
EvaluationMorphismWithGivenSource was called 36 times with a total runtime of 108 ms ( = 3000 μs per execution)
IdentityMorphism was called 42 times with a total runtime of 14 ms ( = 333 μs per execution)
InternalHomOnObjects was called 72 times with a total runtime of 7292 ms ( = 101277 μs per execution)
IsEqualForObjects was called 72 times with a total runtime of 10 ms ( = 138 μs per execution)
KernelEmbedding was called 73 times with a total runtime of 7181 ms ( = 98369 μs per execution)
LiftAlongMonomorphism was called 36 times with a total runtime of 920 ms ( = 25555 μs per execution)
PreCompose was called 108 times with a total runtime of 20 ms ( = 185 μs per execution)
TensorProductOnMorphismsWithGivenTensorProducts was called 36 times with a total runtime of 7 ms ( = 
194 μs per execution)
TensorProductOnObjects was called 399 times with a total runtime of 129 ms ( = 323 μs per execution)
```

So the manually written ones are much faster. :( The morphisms were the same in both runs.

> > * I suspect I made a mistake somewhere in the logic templates, because the sudden change in e.g. [these numbers](https://github.com/homalg-project/CAP_project/pull/932/commits/86123704f5aa8d60047d3705e1ecc1357b984043#diff-a9a59588544cd782f888456a63312512e44f9c38478bb39590a4ad5775e42b16R40-R41) looks suspicious to me. :/ They appear after `LambdaElimination`, which eventually boils down to `EvaluationMorphismWithGivenSource`, and this uses a lot of the `LeftDivide` and `ReducedSyzygies` templates.
> 
> Do those changes disappear when compiling without the logic templates? If yes, you can bisect your logic templates to find the offending one.

Disabling my logic templates revealed, that it might be worse without. 
The `Expected Output` is what I get with all logic templates enabled, the `but found` is what I get without my logic templates.
```gap
testing: /home/tom/.gap/pkg/CAP_project/ModulePresentationsForCAP/tst/modulepresentationsforcap04.tst
########> Diff in /home/tom/.gap/pkg/CAP_project/ModulePresentationsForCAP/tst/modulepresentationsforcap04.ts\
t:43
# Input is:
Display( l );
# Expected output:
[ [  2406,     0 ],
  [   171,     0 ] ]

A morphism in Category of left presentations of Z
# But found:
[ [  236554,       0 ],
  [   78415,       0 ] ]

A morphism in Category of left presentations of Z
########

########> Diff in /home/tom/.gap/pkg/CAP_project/CAP/tst/cap16.tst:46
# Input is:
Display( UnderlyingMatrix( morphism_l1 ) );
# Expected output:
[ [  -84 ],
  [  -25 ] ]
# But found:
[ [   -84 ],
  [  -196 ] ]
########
########> Diff in /home/tom/.gap/pkg/CAP_project/CAP/tst/cap16.tst:53
# Input is:
Display( UnderlyingMatrix( morphism_l2 ) );
# Expected output:
[ [  -39 ],
  [  -55 ] ]
# But found:
[ [  -39 ],
  [  -91 ] ]
########
########> Diff in /home/tom/.gap/pkg/CAP_project/CAP/tst/cap16.tst:94
# Input is:
Display( UnderlyingMatrix( morphism_r1 ) );
# Expected output:
[ [  -84,  -25 ] ]
# But found:
[ [   -84,  -196 ] ]
########
########> Diff in /home/tom/.gap/pkg/CAP_project/CAP/tst/cap16.tst:100
# Input is:
Display( UnderlyingMatrix( morphism_r2 ) );
# Expected output:
[ [  -39 ],
  [  -55 ] ]
# But found:
[ [  -39 ],
  [  -91 ] ]
########
```
So maybe I was frightened for nothing and there might be nothing wrong with the logic templates.
I can also run the above timings to check how much the templates speed up computations, if you like.

>     3. You compile `KernelEmbedding` although it differs from the manual version. Is this deliberate?

No, it's an artifact.

>     4. You resolve `INTERNAL_HOM_EMBEDDING` (which uses `KernelEmbedding`), is this deliberate? This might explain the differing numbers.

Yes, I get a lot of warnings and an error otherwise:

```gap
WARNING: Could not find declaration of INTERNAL_HOM_EMBEDDING (current input: [ <Category "IsFreydCategory">, <Cat\
egory "IsFreydCategoryObject">, <Category "IsFreydCategoryObject"> ])
while compiling function with name "Function added to Freyd( Rows( Q ) ) for InternalHomOnObjects"
located at /home/tom/.gap/pkg/CAP_project/FreydCategoriesForCAP/gap/FreydCategory.gi:1534

WARNING: Could not find declaration of INTERNAL_HOM_EMBEDDING (current input: [ <Category "IsFreydCategory">, <Cat\
egory "IsFreydCategoryObject">, <Category "IsFreydCategoryObject"> ])
while compiling function with name "Function added to Freyd( Rows( Q ) ) for InternalHomOnMorphismsWithGivenIntern\
alHoms"
located at /home/tom/.gap/pkg/CAP_project/FreydCategoriesForCAP/gap/FreydCategory.gi:1576

WARNING: Could not find declaration of INTERNAL_HOM_EMBEDDING (current input: [ <Category "IsFreydCategory">, <Cat\
egory "IsFreydCategoryObject">, <Category "IsFreydCategoryObject"> ])
while compiling function with name "Derivation (first added to Rows( Q )) of InternalHomOnMorphisms"
located at stream:2

WARNING: Could not find declaration of INTERNAL_HOM_EMBEDDING (current input: [ <Category "IsFreydCategory">, <Cat\
egory "IsFreydCategoryObject">, <Category "IsFreydCategoryObject"> ])
while compiling function with name "Function added to Freyd( Rows( Q ) ) for CoevaluationMorphismWithGivenRange"
located at /home/tom/.gap/pkg/CAP_project/FreydCategoriesForCAP/gap/FreydCategory.gi:1682

WARNING: Could not find declaration of INTERNAL_HOM_EMBEDDING (current input: [ <Category "IsFreydCategory">, <Cat\
egory "IsFreydCategoryObject">, <Category "IsFreydCategoryObject"> ])
while compiling function with name "Function added to Category of left presentations of Q for CoevaluationMorphism\
WithGivenRange"
located at stream:2

Error, Could not get rid of all global variables, see <function_string>. You should use category_hints.category_attribute\
_names. at /home/tom/.gap/pkg/CAP_project/CompilerForCAP/gap/PrecompileCategory.gi:168 called from
CapJitPrecompileCategory( category_constructor, given_arguments, package_name, compiled_category_name 
 ); at /home/tom/.gap/pkg/CAP_project/CompilerForCAP/gap/PrecompileCategory.gi:362 called from
CapJitPrecompileCategoryAndCompareResult( function ( ring )
      return LeftPresentationsAsFreydCategoryOfCategoryOfRows( ring );
  end, [ ring ], "ModulePresentationsForCAP", Concatenation( "LeftPresentationsAsFreydCategoryOfCategoryOfRowsOf"
    , name, "Precompiled" ) 
 ); at examples/PrecompileModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.g:82 called from
<function "precompile_LeftPresentations">( <arguments> )
 called from read-eval loop at examples/PrecompileModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.g:86
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk> 
```

Huge thank you!

zickgraf:
> So maybe I was frightened for nothing and there might be nothing wrong with the logic templates.

Your fears are very sensible: Logic templates should not change results, otherwise it is very hard to judge whether the CAP operations are still consistent (and indeed, usually they are not, see for example https://github.com/homalg-project/CAP_project/pull/1194#discussion_r1042180284). So that the logic templates seem to "improve" the result is not a good thing :/ Can you find out which logic templates are responsible for these changes? I suspect (at least) some of the ones including `ReducedSyzygiesOfRows`: e.g. `ReducedSyzygiesOfRows( HomalgZeroMatrix( m, 0, ring ) )` is in general not equal to `HomalgIdentityMatrix( m, ring )`, for example it might also be a matrix with ones on the counter diagonal. Even worse, non of this is even 100% consistent on the matrix level, see for example https://github.com/homalg-project/CAP_project/issues/594. I hope that we are not running into issues like that one.

All other points are better discussed verbally, I think. Thanks for the tests!

TKuh:
> So that the logic templates seem to "improve" the result is not a good thing :/ Can you find out which logic templates are responsible for these changes? I suspect (at least) some of the ones including `ReducedSyzygiesOfRows`: e.g. `ReducedSyzygiesOfRows( HomalgZeroMatrix( m, 0, ring ) )` is in general not equal to `HomalgIdentityMatrix( m, ring )`, for example it might also be a matrix with ones on the counter diagonal. 

I have now tested `CAP` and `ModulePresentationsForCAP` with a compiled version which has
- all logic templates activated
- all logic templates deactivated
- all logic templates containing a `Reduced*` deactivated (the rest activated)

and in all three cases the tests return the same results and numbers. :)

I think the problem in https://github.com/homalg-project/CAP_project/pull/932#issuecomment-1357880243 was, that while I compiled with all logic templates, I didn't delete the manual code from `ModulePresentations` (wrong order of commits...). So the `expected output` in https://github.com/homalg-project/CAP_project/pull/932#issuecomment-1357880243 does not reflect running the tests with compiled code, but simply the result of running the manual version of `ModulePresentationsForCAP`.

So it does not seem, that the logic templates change results (at least for the tests).

zickgraf:
For future reference: I thought the code of `IsEqualForObjects` would change after https://github.com/homalg-project/CAP_project/pull/1259 but it doesn't.
Background: `IsEqualForObjects` in Freyd uses `IsEqualForMorphismsOnMor` in the underlying category of rows, which compares objects first. Thus, I somehow expected to see a comparison including `RankOfObject`. However, ModulePresentations does not explicitly store the dimensions of the matrices underlying the objects. Thus, the objects in the category of rows are created on the fly from the number of rows and columns of the matrices directly, and so only `NrRows` and `NrCols` appears in the compiled code of `IsEqualForObjects` instead of `RankOfObject`.

In any case, the use of `IsEqualForMorphismsOnMor` explains those additional checks on the matrix dimensions in the compiled code of `IsEqualForObjects`. We could of course avoid those checks with a very specific logic template, but I don't think this is necessary.

So to continue: Please extract the compilation of `PreCompose` and `IsEqualForObjects` (and dropping the uncompiled code) as well as the move of the operations in CompilerForCAP/examples/PrecompileModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.g from this PR. This new PR should not need any additional discussion and the PR here would get smaller. Then we can finally discuss the monoidal structure.

zickgraf:
As just discussed:

1. Please finish the work on translating INTERNAL_HOM_EMBEDDING from the Freyd to the ModulePresentations version (in particular: more type signatures have to be added, the logic templates should be made compatible with left and right versions, the expected number of objects/morphism has to be adjusted, some temporary changes have to be reverted).
2. Check if the logic templates for ReducedSyzygiesOfRows are required (actually I expect ReducedSyzygiesOfRows to disappear after 1.).
3. See if the logic templates for rewriting KroneckerMat to DiagMat really offer an improvement.
4. Separate the logic for MatrixCategory from the logic for ModulePresentations.
5. The changes to the compiled code of MatrixCategory should be in a separate PR.

We can meet again after 1. is finished (or if there are problems, of course), 2. - 5. can come afterwards. Thanks a lot!

PR #947, opened 2022-06-20
Derive DualOnMorphismsWithGivenDuals via coevaluation and evaluation for duals

zickgraf:
@sebastianpos This is the construction in Remark 3.29 of your thesis. Is there a reason why you did not add it as a derivation, or did you simply not need it?

Note: I think to make sure things are consistent, I have to show that if we set `DualOnObjects( a ) := InternalHomOnObjects( a, 1 )` the above derivation will be congruent to `InternalHomOnMorphisms( -, 1 )`. I hope this is true and not too hard :D

### DIFF
--- a/MonoidalCategories/PackageInfo.g
+++ b/MonoidalCategories/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "MonoidalCategories",
 Subtitle := "Monoidal and monoidal (co)closed categories",
-Version := "2022.06-04",
+Version := "2022.06-05",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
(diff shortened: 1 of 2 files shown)


zickgraf:
PS: You do not have to review the implementation, that should be straight-foward.

sebastianpos:
No, there is no reason why I left it out. It seems to be a harmless derivation.

zickgraf:
Perfect, thanks for the fast reply!

mohamed-barakat:
Thank you very much.

zickgraf:
Note for future reference: I have not shown that if we set `DualOnObjects( a ) := InternalHomOnObjects( a, 1 )` the above derivation will be congruent to `InternalHomOnMorphisms( -, 1 )`. However, `DualOnMorphisms` has no actual specification and is only used to derive `InternalHomOnMorphisms`. For this, it only has to form a contravariant functor (together with `DualOnObjects`), and that it does.

PR #125, opened 2018-01-30
Add missing cases for switching off logic

sebastianpos:

PR #953, opened 2022-06-22
Set categorical properties of opposite category even if logic is disabled

zickgraf:
@sebastianpos You have disabled this in #125 together with the logic for objects and morphisms. I don't know what the context of that PR was, but I assume it was for performance reasons. I think the logic for the categorical properties should not have a noticeable performance impact, and having the properties is important for derivations -> I think we should enable this logic again.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.06-04",
+Version := "2022.06-05",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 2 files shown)


zickgraf:
I have pushed another commit setting the categorical properties of the opposite category before installing operations so the properties are taken into account by the derivations.

sebastianpos:
I think the point was to really be able to switch off all logic. In this spirit, I would prefer a more fine grained solution.

If you don't have time for that, you can go ahead and merge this.

zickgraf:
> I think the point was to really be able to switch off all logic. In this spirit, I would prefer a more fine grained solution.

I see. I would distinguish two things here:

1. Setting the (dual) properties during creation of the opposite category. This is a thing that any doctrine-based category constructor does, so I would not call this "logic" (this also does not need immediate methods).
2. Propagating properties from the category to the opposite category (and back) if properties become known *after* the creation of the opposite category. (I have not come across code which actually sets properties after creation, but of course in principle this can happen.) This I would call some kind of logic.

Currently those things are coupled. I will push a commit decoupling the two things by setting the properties manually during creation, and hope that this will fit the spirit :-)

PR #967, opened 2022-08-08
set IsZeroForObjects/IsZeroForMorphisms in TerminalCategory

mohamed-barakat:


zickgraf (commented):
I think this could be solved in a cleaner way by putting the additional filters in the object and morphism representations, e.g. as in the case of the MatrixCategory in https://github.com/homalg-project/CAP_project/blob/0e6b279d45ba66972160e221418e05a231c3d72d/LinearAlgebraForCAP/gap/LinearAlgebraForCAP.gi#L52.

PR #970, opened 2022-08-09
CapJitAddTypeSignature( "Cartesian", [ IsList ] ... )

mohamed-barakat:


zickgraf (approved):
Note: The non-homogeneous case could in the future be handled by using the multiple-arguments version of `Cartesian` and returning a list of tuples.

Issue #977, opened 2022-08-11
Add more derivations for `...Inverse`


zickgraf:
CAP operations which return isomorphisms usually have an inverse version, e.g. `TensorProductInternalHomCompatibilityMorphism` and `TensorProductInternalHomCompatibilityMorphismInverse`. Often, the non-`Inverse` version has a proper derivation, while the `Inverse` version is simply derived from the non-`Inverse` version using `InverseForMorphisms`. I expect that in most cases a faster derivation for the `Inverse` version similar to the derivation of the non-`Inverse` version could be written.

Background: This shows up when taking `Opposite` categories and looking at the compiled code, or when comparing the compiled code of non-`Inverse` and `Inverse` operations.

PR #964, opened 2022-08-03
Add new options to WrapperCategory for explicit conversion of data structures

zickgraf:
Previously, this could only be done by overwriting ObjectConstructor et al.

PR #978, opened 2022-08-11
Disable post processing for syntax trees which are cached and later reused

zickgraf:
This PR has an interesting side-effect:

We have the equality (Rank-nullity theorem)
```
NumberRows( matrix ) - RowRankOfMatrix( matrix ) = NumberRows( SyzygiesOfRows( matrix ) )
```
which manifests itself in
```
KernelObject( mor ) = Source( KernelEmbedding( mor ) )
```
in the implementation of MatrixCategory. Previously, always the right-hand side of the equation was used, which in general is slower than the left-hand side. With this PR, the left-hand side is used. In cases where `SyzygiesOfRows` is computed anyway (e.g. `FiberProductFunctorial`) this is a performance regression. In cases where `SyzygiesOfRows` is not computed (e.g. `FiberProduct`) this is a performance improvement.

There are two further changes to compiled code in this PR:

1. `IsWellDefinedForMorphisms` for CategoryOfColumns is now correct (in particular, we don't actually need #964 for this).
2. Many derivations of `...Inverse` are now even less efficient than before. I have opened #977 to fix this on a different level.

### DIFF
--- a/CompilerForCAP/PackageInfo.g
+++ b/CompilerForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CompilerForCAP",
 Subtitle := "Speed up computations in CAP categories",
-Version := "2022.08-03",
+Version := "2022.08-04",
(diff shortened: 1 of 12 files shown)


zickgraf:
Closing and reopening to get a better Codecov report.

PR #983, opened 2022-08-15
Add CoFreydCategory(AsOppositeOfFreydCategoryOfOpposite)

zickgraf:
This does not include all the convenience functions which FreydCategory has, and it only compiles the basic operations available regardless of the concrete input category (e.g. not the tensor structure which is installed conditionally only if the concrete input category has a tensor structure).

@sebastianpos The code should be clean and stable, so even if this is not used at the moment, I would still like to merge it as a demo for CompilerForCAP. Is that okay? The only thing we should make sure to get right are the names: are `CoFreydCategory`, `CoRelationMorphism` and `UnderlyingMorphism` the names we want for the constructor and the attributes?

### DIFF
--- a/CompilerForCAP/PackageInfo.g
+++ b/CompilerForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CompilerForCAP",
 Subtitle := "Speed up computations in CAP categories",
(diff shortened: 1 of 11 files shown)


sebastianpos:
Yes, the names are all right. I haven't found a better name for the concept of a CoFreyd category yet.

PR #984, opened 2022-08-15
Make `/` a generic shorthand for `ObjectConstructor`

zickgraf:
@ all Do we want this?

Pros: Less code to write/think about. It should never cause problems because one can always install a method with more special filters if the generic method does not have the desired behavior.

Cons: Seeing a general installation of "/" for IsObject, IsCapCategory might be confusing.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.08-07",
+Version := "2022.08-08",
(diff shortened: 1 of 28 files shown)


mohamed-barakat:
I am in favor :)

zickgraf:
There is another downside: All other operations for "/" (e.g. for homalg matrices in MatrixCategory calling `MorphismConstructor`) have to be installed using `InstallOtherMethod` to prevent
```
#I  method installed for / matches more than one declaration
```

zickgraf:
There is another downside: All other operations for "/" (e.g. for homalg matrices in MatrixCategory calling `MorphismConstructor`) have to be installed using `InstallOtherMethod` to prevent
```
#I  method installed for / matches more than one declaration
```

mohamed-barakat (dismissed):

PR #988, opened 2022-08-16
Add basic support for bare categories

zickgraf:
A bare category is a category whose objects and morphisms are represented
by primitive data types, e.g. integers. See the added example for
background information.

I'm not sure how useful this actually is, so we might remove this again in the future, but I think at the moment it is a nice demo for the flexibility for WrapperCategory and CAP.

### DIFF
--- a/CAP/examples/unwrapped_MatrixCategory.g
+++ b/CAP/examples/unwrapped_MatrixCategory.g
@@ -1,0 +1,263 @@
+#! @Chapter Examples and Tests
+
+#! @Section WrapperCategory
+
+#! A priori, CAP only supports **wrapped** categories, that is, objects and morphisms in a category cannot be represented by primitive data types (like integers) but must be attribute storing GAP objects.
+#! In theory, CAP could support **bare** categories, that is, categories where objects and morphisms are represented by primitive data types. However, this would not be very convenient:
+#! * Type checking would be severely limited due to missing type information. For example, one could not distinguish between objects in a category of rows and objects in a category of columns if both are represented by plain integers.
+#! * In particular, one would always have to call CAP operations with the category as the first argument.
+#! * Morphisms could not store their source and range, so all CAP operations would have to be extended to explicitly include source and range of the morphisms in the input in addition to the primitive morphism data.
+#! * In particular, users would have to explicitly compute and take care of keeping records of the correct source and range of morphisms. In some cases this might actually lead to optimizations, but in most cases users would probably develop their own data structure for storing source and range together with the morphism datum, thus inventing the data structure we wanted to avoid in the first place.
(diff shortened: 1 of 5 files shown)


CAP/examples/unwrapped_MatrixCategory.g
```diff
@@ -0,0 +1,184 @@
...
+#! * In particular, one would always have to call CAP operations with the category as the first argument.
+#! * Morphisms could not store their source and range, so all CAP operations would have to be extended to explicitly include source and range of the morphisms in the input in addition to the primitive morphism data.
+#! * In particular, users would have to explicitly compute and take care of keeping records of the correct source and range of morphisms. In some cases this might actually lead to optimizations, but in most cases users would probably develop their own data structure for storing source and range together with the morphism datum, thus inventing the data structure we wanted to avoid in the first place.
+#! Despite these impracticalities in algorithms, when thinking about mathematics we often deal with categories where the objects and morphisms are of some primitive data type and the category is clear from the context.
+#! In particular, in principle we actually have to proof that our wrapping and unwrapping in CAP defines an equivalence of categories. Often this is trivial because wrapping and unwrapping just cancels.
```
mohamed-barakat, line 12:
```suggestion
#! In particular, in principle, we actually have to prove that our wrapping and unwrapping in CAP defines an equivalence of categories. Often this is trivial because wrapping and unwrapping just cancels.
```

zickgraf: Thanks!

zickgraf:
I think I am going to abandon the idea of bare categories. I thought this idea would be nice for didactic purposes and for modeling mathematics inside CAP in a more faithful way. However, the concept does not fulfill these expectations:

1. Didactic purposes: If a person is not interested in category theory and CAP at all, we can simply provide wrapper functions with primitive data types as inputs and output which will internally be mapped to a categorical structure. On the other hand, if a person is interested in CAP, than using bare categories would give a version of CAP with much of the convenience and many type checks removed -> this is not what we want for didactic purposes.
2. Modeling mathematics in a more faithful way: This works for objects but not really for morphisms. CAP simply expects `Source` and `Range` at many places, and if one starts endowing some primitive data structure with  source and range explicitly, one more or less simply reinvents `CreateCapCategoryMorphism`. A theoretical solution would be to enhance all categorical operations to accept all sources and ranges of morphisms in the input as explicit arguments, and provide convenience functions which automatically deduce these arguments if possible. This is how Agda models categories. But Agda has the advantage the deducing implicit arguments is built into the language: https://agda.readthedocs.io/en/v2.6.3/language/implicit-arguments.html So they can use this more or less for free. For CAP this would currently be a huge effort. A significant part of it could probably be automated but we would also run into GAP's limitation that an operation cannot get more than 6 arguments. Of course we could enhance GAP to support more arguments or write our own method handling, but all of this would be very time consuming and, I think, not worth the effort.

Issue #936, opened 2022-06-15
Compile AdelmanCategory( ... ) as Wrapper( Op( Freyd( Op( Freyd( ... ) ) ) ) )


zickgraf:

mohamed-barakat:
I would also suggest the other order `Wrapper( Freyd( Op( Freyd( Op( ... ) ) ) ) )`.

zickgraf:
My naive idea was: The wrapper category should use this non-trivial equivalence to translate between the data structures of `AdelmanCategory( ... )` and `Op( Freyd( Op( Freyd( ... ) ) ) )`, so that `Wrapper( Op( Freyd( Op( Freyd( ... ) ) ) ) )` actually behaves like and compiles to `AdelmanCategory( ... )`. Do you think this is feasible?

sebastianpos:
I think I'm simply not aware of the capabilities of `Wrapper` categories. My naive thoughts on this project were the following: your compiler gets as an input 
1) the code of `Op( Freyd( Op( Freyd( ... ) )`
2) the code that implements the non-trivial equivalence (in both directions) with the data structures of Adelman objects and Adelman morphisms

and outputs a compiled category whose algorithms somehow look like a compiled version the Adelman category.

But maybe the `Wrapper` categories somehow merge 1) and 2) into a single category, and you are going to simply compile that category?

In any case, I would suspect that during this process, several identites and zero morphisms are going to blow up the objects, and the only way to get rid of them would be applying simplifications that somehow need to be aware of the big picture (e.g., just looking at the underlying matrix code and applying equalities of matrices could never get rid of stuff that blows up the representation of an object).

zickgraf:
> But maybe the `Wrapper` categories somehow merge 1) and 2) into a single category, and you are going to simply compile that category?

Exactly, `WrapperCategory` gets the code from your point 2) and thus CompilerForCAP does not have to do anything special :-)

> In any case, I would suspect that during this process, several identites and zero morphisms are going to blow up the objects, and the only way to get rid of them would be applying simplifications that somehow need to be aware of the big picture (e.g., just looking at the underlying matrix code and applying equalities of matrices could never get rid of stuff that blows up the representation of an object).

I (naively?) hope that this big picture optimizations could be encoded as rewrite rules for the compiler. So this sound like a promising example for testing the limits of the current approach.

[...]

PR #989, opened 2022-08-17
Add more documentation to WrapperCategory

zickgraf:
@sebastianpos This describes how `WrapperCategory` works and how it relates to `CompilerForCAP` as promised in #936.

@ all I'm happy if you have feedback but I will also merge the PR without feedback because having this is better than nothing in any case :D We can of course change things later on.

### DIFF
--- a/CAP/gap/WrapperCategory.gd
+++ b/CAP/gap/WrapperCategory.gd
@@ -5,7 +5,9 @@
 #
 
 #! @Chapter Create wrapper hulls of a category
-
+#!
+#! @Section Introduction
+#!
 #! The support for building towers of category constructors is one
 #! of the main design features of CAP. Many categories that
 #! appear in the various applications can be modeled by towers
@@ -20,18 +22,18 @@
 #! * ObjectDatum
 #! * MorphismDatum
 #! in order to reflect the desired interpretation with
-#! a user-interface that is independent of the modeling tower.
+#! a user-interface that is independent of the modeling tower (see below for details).
 #! Note that the same tower might have multiple interpretations.
 #! 
 #! <Table Align="|c|">
-#! <Caption>A tower of categories modeling the category <C>W</C></Caption>
+#! <Caption>A tower of categories modeling the category `W`</Caption>
 #! <HorLine/>
 #! <Row>
-#!   <Item><C>W := WrapperCategory( cat_n )</C></Item>
+#!   <Item>`W := WrapperCategory( cat_n )`</Item>
 #! </Row>
 #! <HorLine/>
 #! <Row>
-#!   <Item><C>cat_n := CategoryConstructor_n( cat_{n-1} )</C></Item>
+#!   <Item>`cat_n := CategoryConstructor_n( cat_{n-1} )`</Item>
 #! </Row>
 #! <HorLine/>
 #! <Row>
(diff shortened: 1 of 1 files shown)


CAP/gap/WrapperCategory.gd
```diff
@@ -20,33 +22,142 @@
...
+#!     Set the options `category_filter`, `category_object_filter`, and `category_morphism_filter`
+#!     to the filters corresponding to the custom data structure, e.g. `IsCategoryOfColumns`,
+#!     `IsCategoryOfColumsObject`, and `IsCategoryOfColumsMorphism`.
+#! </Item>
+#! <Item>
+#!     Set `object_constructor`, `object_datum`, `morphism_constructor`, and `morphism_datum`
+#!     to the functions one would write for `ObjectConstructor` and so on for a primitive implementation.
+#!     In our example, `object_constructor` takes the wrapper category (which `IsCategoryOfColumns` due
+#!     to the filter set in the first step) and an integer, and returns a CAP object
+#!     in the category with attribute `Dimension` set to the integer, just like a primitive
```
mohamed-barakat, line 72: `Dimension` or `RankOfObject`?

CAP/gap/WrapperCategory.gd
```diff
@@ -20,33 +22,142 @@
...
+#!     implementation of `CategoryOfColumns` would do.
+#! </Item>
+#! <Item>
+#!     Set `modeling_tower_object_constructor`, `modeling_tower_object_datum`,
+#!    `modeling_tower_morphism_constructor`, and `modeling_tower_morphism_datum`:
+#!    `modeling_tower_object_constructor` gets the same input as `object_constructor` but must return
+#!    the corresponding object in the tower. `modeling_tower_object_datum` has the same output as `object_datum`
+#!    but gets the wrapper category and an object in the tower as an input.
+#!    In our example, `modeling_tower_object_constructor` gets the wrapper category and an integer
+#!    as in step 2 and wraps the integer as a `CategoryOfRowsObject` and the result as an opposite
```
mohamed-barakat, line 82: opposite object -> object in the opposite category?

CAP/gap/WrapperCategory.gd
```diff
@@ -20,33 +22,142 @@
...
+#!    object. `modeling_tower_object_datum` gets the wrapper category and an object in
+#!    `Opposite( CategoryOfRows( R ) )` (that is, an integer wrapped in a category of rows object
+#!    wrapped in an opposite object) and returns the underlying integer.
```
mohamed-barakat, line 85: dito

CAP/gap/WrapperCategory.gd
```diff
@@ -20,33 +22,142 @@
...
+#!     object or a morphism in the wrapper category.
+#! </Item>
+#! </Enum>
+#! The first two steps define the functor "Unwrapping" and the last two steps define the functor "Wrapping".
+#! "Wrapping" on objects and morphisms is called `ModeledObject` and `ModeledMorphism` in the code.
+#! "Unwrapping" on objects and morphisms is called `ModelingObject` and `ModelingMorphism` in the code.
+#!
+#! @Section Relation to `CompilerForCAP`
+#!
+#! The operation of the tower (step 3 above) usually unwraps objects and morphisms,
```
mohamed-barakat, line 142: This is confusion. In step 3 you are already working with the unwrapped objects/morphisms in cat_n.

mohamed-barakat (changes requested):
Wonderful!

CAP/gap/WrapperCategory.gd
zickgraf: Ah, correct, what I mean is our usual kind of unwrapping which has nothing to do with WrapperCategory: obtaining the underlying data from a (tower) object or morphism. I have thought about calling this "(un)boxing" for some time because that seems to be the commonly used term for this concept: https://en.wikipedia.org/wiki/Boxing_(computer_science)

And while we're at it: I also don't like the term "WrapperCategory" anymore because "reinterpreting data"  is certainly not the same as "wrapping data". However, I currently have no good suggestion for a better name. Some obvious ideas:
* ReinterpretationCategory (or ReinterpretedCategory)
* TranslationCategory (or TranslatedCategory)
* RewritingCategory

In any case, we could keep the constructor "WrapperCategory" for the kind of wrapper category which really only wraps the data just as the original version of WrapperCategory did. This would maybe also disentangle some of the concepts which currently live in WrapperCategory.gi. For example, WrappingFunctor is only valid for this basic version of WrapperCategory and does not coincide with my definition of the functor "Wrapping" in the documentation.

mohamed-barakat: I agree and I like the name `ReinterpretationCategory` very much, this is the name I keep using when I explain the concept to others. What the documentation describes is indeed not a mere wrapping process, but a reinterpretation process. However, I still see `WrapperCategory` as a particular case of `ReinterpretationCategory` and the constructor's implementation confirms this; I would say the current implementation of the `WrappingFunctor` is broken.

zickgraf: > I agree and I like the name `ReinterpretationCategory` very much

I also think "reinterpretation" is the term we want to use here, I'm just not convinced by `ReinterpretationCategory( C )` because spelling this out as "the reinterpretation category of C" sounds a bit weird. I would like to say `Reinterpretation( C )`, but the term `Reinterpretation` is too generic I think. `ReinterpretedCategory` has the problem that it could both mean the category being reinterpreted and the result of the reinterpretation. What about `ReinterpretationOfCategory` or `ReinterpretCategory`? I admit these sound less like category constructors, but then again, what else could they be? :D

Either way, we have to also fit this into the existing terms `ModelingObject`, `ModeledObject` etc. If we follow this scheme, we could rename `WrapperCategory` to `ModeledCategory`. But this would clash a bit with "model categories". Of course we could also rename `ModeledObject` to `ReinterpretedObject`, and maybe keep `ModelingObject`.

The functor from `cat_n` to `W` would best be called "Reinterpretation", I think. How would we call its inverse: "OriginalInterpretation", "Model" (or "ReReinterpretation" :D )?

Disclaimer: I'm very in favor of renaming WrapperCategory, I just want to make sure we keep things consistent, that's why I'm bringing up so many questions :D

sebastianpos: I'm sorry for the non-constructive nature of the following comment: Please do not opt for the name "translation category". There is a concept of a translation groupoid associated to a group acting on a set, and a natural extension of this idea would be to call translation category a category associated to a monoid acting on a set. I would like to reserve this name for that particular concept.

mohamed-barakat: What about `ReinterpretedCategory`?

mohamed-barakat: > What about `ReinterpretedCategory`?

No, `ReinterpretedCategory` could also be a name for the current `ModelingCategory`. Hmm.

zickgraf: > No, `ReinterpretedCategory` could also be a name for the current `ModelingCategory`. Hmm.

Exactly, that's what I meant above:
> `ReinterpretedCategory` has the problem that it could both mean the category being reinterpreted [i.e. the current `ModelingCategory`] and the result of the reinterpretation.

zickgraf: We just discussed the following naming convention:

* `WrapperCategory`, `ModeledObject` and `ModeledMorphism` are renamed to `ReinterpretationOf{Category,Object,Morphism}`.
* `Modeling{Category,Object,Morphism}` are kept.
* `WrappingFunctor` is renamed to `ReinterpretationFunctor` with component functions `ReinterpretationOf{Object,Morphism}`.

So only one question remains: How to call the inverse of `ReinterpretationFunctor`? Ideas:
* `ModelingFunctor` or `ModelFunctor`
* `OriginalInterpretationFunctor`, `InitialInterpretationFunctor` or `PrimaryInterpretationFunctor`
* `ReverseInterpretationFunctor`
* `InverseInterpretationFunctor` or `InterpretationFunctorInverse`

zickgraf: The problem I see with `ModelingFunctor` is that if one interprets it in the same way as `Modeling{Category,Object,Morphism}`, then one would expect `ModelingFunctor( F )` to return a functor which somehow models `F`. In this sense, it would not be consistent with the other names. `ModelFunctor` what be perfect, I think, but that clashes with "model categories" again :/

kamalsaleh: Oo Sorry, I thought the `ReinterpretationFunctor` and _ModelingFunctor_ are operations that take as input the reinterpretation category and output equivalence functors between the the modeling category and its reinterpretation category. But as you mentioned, the input is a functor which then goes a completely different direction.

zickgraf: No, you are right: `ReinterpretationFunctor` and `ModelingFunctor` get the reinterpretation of a category as input, I confused myself :D Still, when writing `ModelingFunctor( C )` one might expect to get a functor modeling `C` somehow.

But I just noticed: There is no `UnwrappingFunctor` in the code right now, so I only need the term for the documentation. And in the documentation I think it's not confusing to call the functor "Model(ing)" because the context is clear. So everything should be fine :-)

mohamed-barakat: > So only one question remains: How to call the inverse of `ReinterpretationFunctor`? Ideas:
> 
> * `ModelingFunctor` or `ModelFunctor`
> * `OriginalInterpretationFunctor`, `InitialInterpretationFunctor` or `PrimaryInterpretationFunctor`
> * `ReverseInterpretationFunctor`
> * `InverseInterpretationFunctor` or `InterpretationFunctorInverse`

I like `InverseInterpretationFunctor`.

TKuh: Maybe my impressions help (as I'm pretty much an outsider with this).

I dislike these:

- `OriginalInterpretationFunctor`, `InitialInterpretationFunctor` or `PrimaryInterpretationFunctor`

Only judging from their names, I see their connection with the concept, but I would probably have to rethink their meaning every time I would use them; they seem easy to forget and a bit unintuitive.

On the other hand, I like these names:
- `ReverseInterpretationFunctor`
- `InverseInterpretationFunctor` or `InterpretationFunctorInverse`

they have an `Inverse` in their names and that tells me right away I'm dealing with some sort of equivalence -> this is easy to remember and to understand. But of course, it depends on how much you like this emphasis on being an inverse.

But I'm an amateur, so feel free to ignore this. :D

mohamed-barakat: On the contrary, all your comments are to the point.

zickgraf:
Superseded by #1316.

PR #997, opened 2022-08-22
Add better error messages for some unsupported types of statements

zickgraf:
Example outputs:
```
Error, for function at .gap/pkg/CAP_project/LinearAlgebraForCAP//gap/LinearAlgebraForCAP.gi:185:
Found an empty statement, probably there is a superfluous semicolon in the function mentioned above. at .gap/pkg/CAP_project/CompilerForCAP//gap/EnhancedSyntaxTree.gi:46 called from
ErrorWithFuncLocation( "Found an empty statement, probably there is a superfluous semicolon in the function mentioned above." ); at .gap/pkg/CAP_project/CompilerForCAP//gap/EnhancedSyntaxTree.gi:145 called from
```
and
```
Error, for function at .gap/pkg/CAP_project/LinearAlgebraForCAP//gap/LinearAlgebraForCAP.gi:185:
Found an assignment to the global variable `test`. Maybe this should be a local variable of the function mentioned above? at .gap/pkg/CAP_project/CompilerForCAP//gap/EnhancedSyntaxTree.gi:46 called from
ErrorWithFuncLocation( "Found an assignment to the global variable `", tree.gvar, "`. Maybe this should be a local variable of the function mentioned above?" 
 ); at .gap/pkg/CAP_project/CompilerForCAP//gap/EnhancedSyntaxTree.gi:150 called from
```

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.08-06",
+Version := "2022.08-07",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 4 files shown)


mohamed-barakat (approved):
Wonderful

PR #1008, opened 2022-09-01
allow compiling standardly wrapped categories

mohamed-barakat:


zickgraf:
How urgent is this for you? I can probably look into supporting operations instead of functions properly in CompilerForCAP, but maybe not before the weekend.

mohamed-barakat:
Only needed when compiling  CategoryOfQuivers, so not urgent.

zickgraf:
@mohamed-barakat This should be fixed by #1028. Can you check?

mohamed-barakat:
Yes, indeed, fixed by #1028.

mohamed-barakat:
Thank you very much.

PR #1019, opened 2022-09-05
Enhance final derivations

zickgraf:
* Allow to specify other categories in the preconditions
* Always automatically detect preconditions and warn if they do not match
  the given preconditions
* Add lots of checks


zickgraf:
Close and reopen to get correct code coverage.

PR #1022, opened 2022-09-05
Remove saturation of derivations

zickgraf:
Background: After my recent changes, in principle we do not have to saturate/reevaluate derivations (two relatively expensive operations, responsible for >10% of the creation time of a category) in `Finalize` anymore. However, this requires that all categorical properties and settings like `supports_empty_limits` have to be set before calling any add function (or more precise: before calling an add function for a operation which might trigger derivations depending on those categorical properties). This PR by default still reevaluates and displays a warning if the conditions above are not fulfilled, but does not reevaluate anymore if `overhead` is set to false (i.e. `overhead` not only avoids run time overhead but also creation time overhead).

@ all Is this speedup something we would like to have? If yes, probably some category constructors will have to be adapted to fulfill the above conditions (or as a workaround to trigger a manual reevaluation before finalizing).

### DIFF
--- a/ActionsForCAP/PackageInfo.g
+++ b/ActionsForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "ActionsForCAP",
 Subtitle := "Actions and Coactions for CAP",
-Version := "2022.08-02",
+Version := "2022.09-01",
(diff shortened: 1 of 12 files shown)


mohamed-barakat (approved):
As discussed, I find this a very good idea.

zickgraf:
Again, closing and reopening for Codecov....

PR #1047, opened 2022-09-21
Simplify `List( list, func )[index]` to `func( list[index] )` in post-processing

zickgraf:


zickgraf:
This is a bit hacky/dangerous: This simplification is done in the post-processing, which means that no logic is applied to the result. This may be a source of confusion if a logic template does not match unexpectedly.

PR #1054, opened 2022-09-22
InfoStringOfInstalledOperationsOfCategory now shows further mathematical properties

mohamed-barakat:
the motivation behind this is two-fold:
1. To see the categorical properties which are not linked to CAP operations
2. To see which constructive categorical properties are mathematically valid but not yet constructive

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.09-16",
+Version := "2022.09-17",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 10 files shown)


zickgraf (approved):
modulo test failures.

Also I will request a review from @sebastianpos to check whether he is okay with the new format (@sebastianpos: A quick look at the examples in the diff is sufficient.)

sebastianpos:
Could you explain:

> mathematically but not yet algorithmically

vs

> mathematically

?

mohamed-barakat:
For a property `P`

"mathematically" means: 
1. the property is added using AddCategoricalProperty

"mathematically but not yet algorithmically" means
1. P was added using AddCategoricalProperty
2. there exists a corresponding entry in the record `CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD`, so `CheckConstructivenessOfCategory` applied to this property will not issue an error.
3. `not IsEmpty( CheckConstructivenessOfCategory( cat, P ) )` 

I can add such explanations to the commit message or to the code. Whatever you prefer.

sebastianpos:
Thank you.

No, I think having your explanation here in this thread suffices. It appears to be a feature for programmers who are already quite deep into implementing categories.

sebastianpos (approved):

PR #1090, opened 2022-10-10
Only set RangeCategoryOfHomomorphismStructure for wrapper category if a homomorphism structure is installed

zickgraf:
This reverts #1087 and proposes a cleaner solution.

@mohamed-barakat Please check if this solves the problem which made you create #1087.


zickgraf:
According to @mohamed-barakat I can merge if the tests of FunctorCategories pass. This is the case up to a larger number of primitively installed operations due to the second to last commit. I will merge and adapt FunctorCategories.

Issue #1078, opened 2022-09-30
Make use of the opposite category to automatically generate code for opposite derivations


mohamed-barakat:
As discussed with Fabian it would be desirable to write half of the derivations and automatically generate the opposite versions using `CompilerForCAP` and the current infrastructure.

Background: I have written [code](https://github.com/homalg-project/CategoryConstructor/blob/54bacd34de5c66f9af29bf2c37d145780e86321e/gap/Tools.gi#L956-L1071) to derive finite limits from products and binary equalizers and I would like to generate the code for colimits automatically. The current [code](https://github.com/homalg-project/CategoryConstructor/blob/54bacd34de5c66f9af29bf2c37d145780e86321e/gap/Tools.gi#L1089-L1135) for finite colimits calls the category constructor `Opposite` inside the derivations which is a hack.

zickgraf:
This would also fix #917.

mohamed-barakat:
Wonderful Fabian.

zickgraf:
For CAP this is done in https://github.com/homalg-project/CAP_project/pull/1468. The system could easily be extended to more cases (e.g. include derivations including multiple categories), final derivations, and other packages.

mohamed-barakat:
Excellent.

Issue #917, opened 2022-05-23
Slow coclosed monoidal derivations due to `PreCompose`


TKuh:
Some derivations in the coclosed part of `MonoidalCategories` are slower than their closed part, due to an unfavorable order of matrix multiplications caused by `PreCompose`. This can be in particular triggered with the monoidal test functions via `only_primitive_operations` .

See https://github.com/homalg-project/CAP_project/issues/886#issuecomment-1125983378 for further details.

Creating this issue so that we don't forget about it.

PR #1092, opened 2022-10-10
Use CategoryOfRows as the range category of the homomorphism structure of CategoryOfColumns

zickgraf:
@ all: Is anyone relying on the fact that CategoryOfColumns has itself as the range of its homomorphism structure? If not, I suggest we use the CategoryOfRows as the range instead. This is what the implementation of `Opposite` automatically gives us, so we can drop some code, and it allows to revert two commits which only existed to make sure CategoryOfColumns could change its range category artificially.


mohamed-barakat (dismissed):

sebastianpos:
Me neither. Moreover, I like your suggestion since it implements the generic viewpoint (a B-homomorphism structure on A induces a B-homomorphism structure on A^{op}).

zickgraf:
> Me neither. Moreover, I like your suggestion since it implements the generic viewpoint (a B-homomorphism structure on A induces a B-homomorphism structure on A^{op}).

Just to make sure I understand things correctly: any B-hom structure on A also induces a B^{op}-hom structure on A (and then of course also vice versa), correct? So equipping A^{op} with a B-hom structure and not with a B^{op}-hom structure is a choice, right?

sebastianpos:
I don't think so. If anything, a resulting induced structure would be something like an "op-hom-structure", since you would have the equation Hom( a, b ) = Hom( H(a,b), 1) instead of Hom( a, b )  = Hom( 1, H(a,b) ).

zickgraf:
> I don't think so. If anything, a resulting induced structure would be something like an "op-hom-structure", since you would have the equation Hom( a, b ) = Hom( H(a,b), 1) instead of Hom( a, b ) = Hom( 1, H(a,b) ).

Ah, I didn't think far enough and did not notice that the 1 was in the wrong position. Then a new try to describe the situation:

Rows has a hom-structure over itself induced by the internal hom. Analogously, it has a "co-hom-structure" over itself induced by the internal co-hom. By opping the range of the co-hom-structure and swapping the arguments of the co-hom-structure on objects and morphisms, we could obtain a Cols-hom-structure on Rows.

`Opposite` swaps the arguments of the Rows-hom-structure to obtain a Rows-hom-structure on Cols. It could instead op the range category of the "co-hom-structure" to obtain a Cols-hom-structure on Cols. The latter would correspond to the situation before this PR.

If you agree that those ideas make sense, I would open an issue for the support for "co-hom-structures".

sebastianpos:
Yes, if you introduced a "co-hom-structure", then there would indeed be a choice in how A^{op} derives its hom-structure:

- either from the hom-structure of A
- or the "co-hom-structure" of A^{op}

That would also be fine. Although I have no immediate need for a co-hom-structure yet.

zickgraf:
Thanks for your replies, I'm now every happy with this point of view :-)

> That would also be fine. Although I have no immediate need for a co-hom-structure yet.

Me neither, but asymmetries pop up now and then (e.g. in #917), so I think it's a good thing to keep track of them. And once we have #1078, adding opposite concepts is cheap because it comes down to just giving the definitions, everything else will be handled automatically.

PR #1095, opened 2022-10-11
Remove "fail" from the return_type's of random methods

kamalsaleh:
and, among others, adjust accordingly the random methods in the category of rows.
(See commit messages for the details)


zickgraf (changes requested):
Please add the category filter `IsAbCategory` to the new derivations. Everything else is very nice, thanks :-)

zickgraf (approved):

PR #1109, opened 2022-10-24
Require GAP 4.12 for CompilerForCAP

zickgraf:


zickgraf:
I assumed that I could finally get rid of a workaround for an inconsistency between GAP 4.11 and GAP 4.12. However, it turns out that since Friday GAP master is using the behaviour of GAP 4.11 again: https://github.com/gap-system/gap/commit/892d1acc27e3cbe7ff27c8db471e1f2a6ac546cb So now the workaround is required for making GAP 4.12 and GAP 4.13 consistent...... :D

PR #1114, opened 2022-10-27
Make RingAsCategory of exterior algebra linear over the underlying field

kamalsaleh:
so that `BasisOfExternalHom` can be derived in `FreydCategory( CategoryOfRows( exterior algebra ) )`


zickgraf:
The CI is failing due to `MultiplyWithElementOfCommutativeRingForMorphisms` now being compiled for CategoryOfColumns, but not every nicely because the data type inference cannot handle the input filter `IsRingElement` yet. I will create a bigger PR which will contain the changes here.

zickgraf:
Superseded by #1152.

Issue #1076, opened 2022-09-30
Make Abelian categories constructively finite (co)complete


mohamed-barakat:
The currently missing constructive finite (co)completeness of Abelian categories is due to the lack of derivations for `Equalizer*`/`Coequalizer*` in Abelian categories. Fabian suggested the following:

* Use part of the existing code for deriving `FiberProduct*`/`Pushout*` to derive `Equalizer*`/`Coequalizer*` from `DirectSumDiagonalDifference`/`DirectSumCodiagonalDifference`.
* Use the new derivations to simplify the derivations of `FiberProduct*`/`Pushout*`.

zickgraf:
>     * Use the new derivations to simplify the derivations of `FiberProduct*`/`Pushout*`.

ProjectionInFactorOfFiberProduct is already derived from equalizers and direct products: https://github.com/homalg-project/CAP_project/blob/99154c4d02ef76b9667caf6df3d314f0b92b7748/CAP/gap/DerivedMethods.gi#L286-L301

TKuh:
> Use part of the existing code for deriving `FiberProduct*`/`Pushout*` to derive `Equalizer*`/`Coequalizer*` from `DirectSumDiagonalDifference`/`DirectSumCodiagonalDifference`.

I don't think it's that easy. :/ 

Roughly, what `DirectSumDiagonalDifference` does, is to construct a matrix

$\text{DiagMat}(\beta_1 - \beta_2, \beta_2 - \beta_3, ..., \beta_{n-1} - \beta_n)$

which is constructed as a morphism $\delta \colon \oplus_{i=1}^{n} P_i \rightarrow \oplus_{i=1}^{n-1} B$.

Taking the kernel and composing it with the projection $\pi_j \colon \oplus_{i=1}^{n} P_i \rightarrow P_j$ gives the equations

$\mathrm{KernelEmbedding(\delta)} \cdot  \pi_i \cdot \beta_i = \mathrm{KernelEmbedding(\delta)} \cdot  \pi_j \cdot \beta_j$

For a detailed understanding, one really has to look at the code. [Here](https://github.com/TKuh/CAP_project/blob/equalizer_via_diagonal_diff/CAP/examples/equalizer_via_diagonal_diff.g) is an example of what happens, when one implements equalizers via `DirectSumDiagonalDifference`.

[...]

PR #1124, opened 2022-11-01
Derivations for (Co)Equalizers

TKuh:
And additionally make the derivations of FiberProduct / Pushout rely on it. See https://github.com/homalg-project/CAP_project/issues/1076

Remarks:
- Right now Abelian categories are not constructively Abelian, because the current [derivation](https://github.com/homalg-project/CAP_project/blob/ecddb9abe70e0e8c465a1249d723454362222805/CAP/gap/DerivedMethods.gi#L2721) of `(Co)Equalizer` is blocked by the filter for limits and I first want to ask about the proper way to change it.
- I renamed `DirectSumDiagonalDifference` to `Shifted(Co)DiagonalDifferenceOfMorphisms` since the original name wasn't really fitting, and I called the new version for equalizers `(Co)TelescopingDifferenceOfMorphisms` (in analogy to telescoping sums). Proposals for better names? :)
-  The [compilation](https://github.com/TKuh/CAP_project/tree/FiberProduct_new_derivation) of the new derivation path `FiberProduct` -> `ShiftedDiagonalDifferenceOfMorphisms` -> `TelescopingDifferenceOfMorphisms` gives an equivalent result to the former `FiberProduct` -> `DirectSumDiagonalDifference`. However, the compiled old derivation was shorter.


zickgraf:
Very nice, thanks!

Regarding your remarks:

1. Thanks for noticing this, fixed: https://github.com/homalg-project/CAP_project/pull/1126.
2. I have to think about this.
3. We can easily fix this with suitable logic templates, but we should first understand where the fixed list accesses (`list[1]` and `list[2]`) come from and see if those can be avoided.

Regarding your first commit: One may not derive `EmbeddingOfEqualizer` as `KernelEmbedding( TelescopingDifferenceOfMorphisms )` directly because the user might already have installed `Equalizer` in a different, possibly incompatible, way. The solution for this problem is the following: Introduce `IsomorphismFromEqualizerToKernelOfTelescopingDifferenceOfMorphisms` which translates between a user-given implementation of the equalizer and the derivation as the kernel of the telescoping difference, and set this to the identity in a final derivation only in case the user has not installed any of the operations concerning equalizers (cf. `IsomorphismFromFiberProductToKernelOfDiagonalDifference`).

I think we best go through this commit by commit.

TKuh:
>     3. We can easily fix this with suitable logic templates, but we should first understand where the fixed list accesses (`list[1]` and `list[2]`) come from and see if those can be avoided.

This is probably from `Source( diagram[ 1 ] )` and `Source( diagram[ 2 ] )`.
https://github.com/homalg-project/CAP_project/blob/15c31281800b77dcc7bf2d71ba940b03af0ab678/CAP/gap/DerivedMethods.gi#L1929-L1931
We could give `TelescopingDifferenceOfMorphisms` the common source of all the morphisms in `diagram` directly, then the list access might be prevented? `Equalizer`, `EmbeddingOfEqualizer`, etc. do it, too.

> Regarding your first commit: One may not derive `EmbeddingOfEqualizer` as `KernelEmbedding( TelescopingDifferenceOfMorphisms )` directly because the user might already have installed `Equalizer` in a different, possibly incompatible, way. The solution for this problem is the following: Introduce `IsomorphismFromEqualizerToKernelOfTelescopingDifferenceOfMorphisms` which translates between a user-given implementation of the equalizer and the derivation as the kernel of the telescoping difference, and set this to the identity in a final derivation only in case the user has not installed any of the operations concerning equalizers (cf. `IsomorphismFromFiberProductToKernelOfDiagonalDifference`).

Thank you, now I understand these isomorphisms. :) 
I gave it a try.

zickgraf:
> We could give TelescopingDifferenceOfMorphisms the common source of all the morphisms in diagram directly, then the list access might be prevented? Equalizer, EmbeddingOfEqualizer, etc. do it, too.

Yes, that's the way to go to also support the case of an empty equalizer diagram. See here https://en.wikipedia.org/wiki/Equaliser_(mathematics)#In_category_theory for the background why the source is needed.

zickgraf:
To record what we have discussed on Friday:

1. The name `DirectSumDiagonalDifference` should be kept.
2. `TelescopeDifferenceOfMorphisms` should be renamed to `JointPairwiseDifferencesOfMorphisms`.
3. All derivations should work without relying on concrete list access (i.e. `list[1]`).

mohamed-barakat:
Could you please rebase this branch?

TKuh:
I rebased, but I have not yet worked on the todo's above (I plan to do it either later today or tomorrow).

TKuh:
There are compiler warnings with `LinearAlgebraForCAP`, which I don't know how to remove, but otherwise this is ready for review.

As discussed, there are methods which can be switched from `DirectSum` to `DirectProduct/Coproduct`, but I would like to do this in a separate PR (unless you prefer it here).

CAP/examples/TerminalCategoryWithMultipleObjects.g
```diff
@@ -8,8 +8,7 @@ LoadPackage( "MonoidalCategories" );
 T := TerminalCategoryWithMultipleObjects( );
 #! TerminalCategoryWithMultipleObjects( )
 InfoOfInstalledOperationsOfCategory( T );
-#! 63 primitive operations were used to derive 280 operations for this category
-#! which algorithmically
+#! 63 primitive operations were used to derive 304 operations for this category which algorithmically
```
mohamed-barakat, line 11:
```suggestion
#! 63 primitive operations were used to derive 304 operations for this category
#! which algorithmically
```

mohamed-barakat:
I rebased on PR https://github.com/homalg-project/CAP_project/pull/1197 and resolved the conflicts. I can push onto your branch once the mentioned PR is merged.

mohamed-barakat:
> There are compiler warnings with `LinearAlgebraForCAP`, which I don't know how to remove, but otherwise this is ready for review.

I was unable to quickly locate the cause of the warning.

zickgraf:
Thanks for the progress! I think all that remains is to remove `DirectSumDiagonalDifference` and `DirectSumCodiagonalDifference`. Those are still used in the derivations of FreydCategoriesForCAP, but with a list of fixed length 2.

I suggest to simply replace the calls to `DirectSumDiagonalDifference` in FreydCategoriesForCAP by the code you have written here: https://github.com/homalg-project/CAP_project/pull/1124/commits/3ce90e2a9cd1dad23b1b5cc2fdc54c982a975968#diff-ba2f61a1820f9344a717236211fac2426be05ccd6340fc34e6d4dfa1b81fbc8aR2298-R2309

TKuh:
> Thanks for the progress! I think all that remains is to remove `DirectSumDiagonalDifference` and `DirectSumCodiagonalDifference`. 

The operations `IsomorphismFromKernelOfDiagonalDifferenceToFiberProduct` and `IsomorphismFromFiberProductToKernelOfDiagonalDifference` should also be removed, right?

> I suggest to simply replace the calls to `DirectSumDiagonalDifference` in FreydCategoriesForCAP by the code you have written here: [3ce90e2#diff-ba2f61a1820f9344a717236211fac2426be05ccd6340fc34e6d4dfa1b81fbc8aR2298-R2309](https://github.com/homalg-project/CAP_project/commit/3ce90e2a9cd1dad23b1b5cc2fdc54c982a975968#diff-ba2f61a1820f9344a717236211fac2426be05ccd6340fc34e6d4dfa1b81fbc8aR2298-R2309)

Thank you for this suggestion, this prevented a second question I would have had otherwise. :)

zickgraf:
> The operations `IsomorphismFromKernelOfDiagonalDifferenceToFiberProduct` and `IsomorphismFromFiberProductToKernelOfDiagonalDifference` should also be removed, right?

Correct :-)

zickgraf:
(and all the duals too, of course)

zickgraf:
I think with https://github.com/homalg-project/CAP_project/pull/1237, all the parts are merged, thanks for the work!

PR #1142, opened 2022-11-07
d&a PostInverse/PreInverseForMorphisms together with two derivation methods

kamalsaleh:
These are, up to their names, copy & paste of the same methods in `CategoryConstructor` package:

https://github.com/homalg-project/CategoryConstructor/blob/5b33ddfbefc58f4923fe8ff74fd4337e349e5209/gap/ToolsMethodRecord.gi#L47-L61

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.11-08",
+Version := "2022.11-09",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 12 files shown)


zickgraf (approved):
Thanks, only one small typo.

mohamed-barakat (changes requested):
I would suggest outsourcing the code in `CategoryConstructor`, i.e., dropping the suffix `ForMorphisms`.

zickgraf:
> I would suggest outsourcing the code in `CategoryConstructor`, i.e., dropping the suffix `ForMorphisms`.

I think the analogy to `InverseForMorphisms` is a good thing.

mohamed-barakat:
> I think the analogy to `InverseForMorphisms` is a good thing.

Right, I agree.

mohamed-barakat (dismissed):

kamalsaleh:
I agree with Fabian in keeping the analogy. This would also reduce the interference with other packages that might declare these `PreInverse `and `PostInverse`, e.g., `Modules `or `GradedModules`. 

We can keep the installation name as is while introducing new "convenience" operations/attribute? similar to `InverseForMorphisms` & `Inverse`.

zickgraf (approved):

PR #1143, opened 2022-11-08
Add CAP derivations to create random morphisms

kamalsaleh:


CAP/gap/CategoryMorphisms.gd
```diff
@@ -170,6 +170,10 @@ DeclareProperty( "IsIdempotent",
 #! @Description
 #! The arguments are an object $a$ in a category $C$ and an integer $n$.
 #! The output is a random morphism $\alpha: a \rightarrow b$ for some object $b$ in $C$.
+#! If $C$ is equipped with the methods <C>RandomObjectByInteger</C> and <C>RandomMorphismWithFixedSourceAndRangeByInteger</C>,
```
zickgraf, line 173:
```suggestion
#! If $C$ is equipped with the methods <C>RandomObjectByInteger</C> and <C>RandomMorphismWithFixedSourceAndRangeByInteger</C>
#! and $C$ is an Ab-category,
```
Similarly for the other two derivations from RandomMorphismWithFixedSourceAndRangeByInteger.

zickgraf (changes requested):
Nice, only one small change request :-)

CAP/gap/CategoryMorphisms.gd
zickgraf: And of course also for the three derivations from RandomMorphismWithFixedSourceAndRangeByList.

FreydCategoriesForCAP/gap/CategoryOfRows.gd
```diff
@@ -76,7 +76,7 @@ DeclareOperation( RandomObjectByList, [ IsCategoryOfRows, IsList ] );
 DeclareOperation( RandomObjectByInteger, [ IsCategoryOfRows, IsInt ] );
 
 #! @Description
-#! The arguments are two objects <A>S</A>, <A>R</A> and a non-empty list <A>L</A> of non-negative integers.
+#! The arguments are two objects <A>S</A>, <A>R</A> and a list <A>L</A> of integers.
```
zickgraf, line 79: Two things:
1. What about allowing any homalg ring element while your at it?
2. The documentation of the other `RandomMorphism*ByList` should also be adjusted.

zickgraf (changes requested):

FreydCategoriesForCAP/gap/CategoryOfRows.gd
zickgraf: Addendum to 1.: Respectively any element that can be multiplied to homalg matrices of the given ring (which is not a condition we can check).

kamalsaleh: Okay, why not!
The code is now relaxed to allow lists of integers and ring elements. I have also improved the error-messages in the derived-methods.

FreydCategoriesForCAP/gap/CategoryOfRows.gi
```diff
@@ -1251,54 +1254,10 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_CATEGORY_OF_ROWS,
     AddRandomMorphismWithFixedSourceAndRangeByInteger( category,
       function ( category, S, R, n )
         
-        if IsNegInt( n ) then
-          Error( "The input should be a non-negative integer\n" );
-        fi;
-
```
zickgraf, line 1257: Please either keep this check or adjust the documentation.

zickgraf (changes requested):

FreydCategoriesForCAP/gap/CategoryOfRows.gi
kamalsaleh: I have updated the documentation.

zickgraf (approved):

PR #1146, opened 2022-11-09
Install BasisOfExternalHom and CoefficientsForMorphism for CategoryOfRows over commutative ring

zickgraf:


zickgraf:
@mohamed-barakat You are the author of the original code, could you review this?

FreydCategoriesForCAP/gap/CategoryOfRows.gi
```diff
@@ -876,6 +851,31 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_CATEGORY_OF_ROWS,
             
         end );
         
+        ##
+        AddBasisOfExternalHom( category,
```
mohamed-barakat, line 855: This is a correct but very expensive way to construct the basis. I would rather create `s * t` `HomalgInitialMatrix`, then make the matrices immutable. If this is not compilable, then we need `homalg` commands to create these matrices as immutable matrices immediately.

zickgraf: Thanks for the review! Since this PR only moves existing code, I will merge it as is and open an issue for the improvements you suggest.

mohamed-barakat: I agree.

PR #1148, opened 2022-11-09
Compile all primitive operations of CategoryOfRows/Columns

zickgraf:


mohamed-barakat:
Congratulations :)

zickgraf:
> Congratulations :)

This is nothing new: I'm only compiling the existing primitive operations which essentially only unpacks `CategoryOfRowsObject` etc. Obtaining those primitive operations from the tower AdditiveClosure( RingAsCategory ) is a different thing.

PR #1157, opened 2022-11-14
Add random methods to AdditiveClosure & RingAsCategory

kamalsaleh:


FreydCategoriesForCAP/gap/RingsAsAbCats.gi
```diff
@@ -372,6 +372,29 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_RING_AS_CATEGORY,
...
+    AddRandomMorphismWithFixedSourceAndRangeByList( category,
+      function( cat, S, R, L )
+        
+        if Length( L ) <> 1 or not IsInt( L[1] ) then
+            Error( "the input should be a list consisting of one integer!\n" );
+        fi;
+        
+        return RandomMorphismWithFixedSourceAndRangeByInteger( cat, S, R, L[1] );
+        
+    end );
```
zickgraf, line 396: Do we really need this? If yes, maybe this could be a final derivation mapping all the "ByList" variants to the "ByInteger" variants by requiring an integers as the single entry.

FreydCategoriesForCAP/gap/AdditiveClosure.gd
```diff
@@ -249,6 +251,92 @@ DeclareOperation( "\/",
...
+##
+#! @Section Random methods for additive closure category
+##
+####################################
+
+if false then
+
+#! @Description
+#! The arguments are an additive closure category <A>C</A> of a category <A>U</A> and a list <A>L</A> whose first entry is a non-negative integer and second entry is a list.
+#! The output is an object in <A>C</A> which is a formal direct sum of at most <A>L</A>[1] objects in <A>U</A> each is computed via <C>RandomObjectByList</C>(<A>U</A>,<A>L</A>[2]).
```
zickgraf, line 264:
```suggestion
#! The output is an object in <A>C</A> which is a formal direct sum of at most <A>L</A>[1] objects in <A>U</A> each computed via <C>RandomObjectByList</C>(<A>U</A>,<A>L</A>[2]).
```

Similarly in the other parts of the documentation.

FreydCategoriesForCAP/gap/AdditiveClosure.gi
```diff
@@ -1303,6 +1303,79 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_ADDITIVE_CLOSURE,
...
+    fi;
+    
+    ## Random Methods by Lists
+    ##
+    if CanCompute( underlying_category, "RandomObjectByList" ) then
+      
+      AddRandomObjectByList( category,
+        function( cat, L )
+          
+          if Length( L ) <> 2 or (not IsInt( L[1] ) ) or IsNegInt( L[1] ) or (not IsList( L[2] )) then
```
zickgraf, line 1355:
```suggestion
          if Length( L ) <> 2 or (not IsInt( L[1] )) or IsNegInt( L[1] ) or (not IsList( L[2] )) then
```

zickgraf (changes requested):
Please also add documentation for RingAsCategory and tests covering the new code.

FreydCategoriesForCAP/gap/RingsAsAbCats.gi
kamalsaleh: No, not really, I only added it for the sake of completeness and for testing the methods I added in the additive closure category.  Anyway, the bundle idea is a good.  I will implement it.

zickgraf (approved):

zickgraf:
@kamalsaleh Now that I have used the random methods for AdditiveClosure in https://github.com/homalg-project/Algebroids/pull/198, I think it would be nicer if `RandomObjectByList` would as first entry get a list of integers from which the size of the object is chosen. This would match CategoryOfRows and would make it much easier to create large objects for tests explicitly.

kamalsaleh:
I agree, and the current behavior would still be reproducible. If you wish, I can create PR.

zickgraf:
> If you wish, I can create PR.

That would be nice :-)

PR #1158, opened 2022-11-15
Miscellaneous simplifications

zickgraf:


zickgraf:
@ all The commits are touching very central things around which filters inherit from which other filters. I don't expect any problems, but should you encounter weird "No Method Found" errors after this PR is merged, that might be the cause.

mohamed-barakat:
Yes, `Algebroids` is failing:

https://github.com/homalg-project/Algebroids/actions/runs/3473165377

PR #1172, opened 2022-11-25
Remove unused functions propagating IsWellDefined

zickgraf:
Besides being unused, the whole concept seems obsolete:

> For every categorial construction the outcome is well defined if
> and only if every input object or morphism of the construction is well defined.

1. The output can be well-defined even if the input was not, so the "if and only if" is too strong.
2. CAP does not specify (anymore?) what happens if you call an operation with not well-defined input.
3. IsWellDefined is used to check if an operation produces a well-defined output given a well-defined input.


mohamed-barakat (dismissed):

zickgraf:
Codecov has issues again, the coverage is actually not changing significantly.

PR #1177, opened 2022-11-28
Fix in documentation of DirectSum(Co)DiagonalDifference

TKuh:
Also, it should be 

`AddIsomorphismFromCokernelOf`**Co**`DiagonalDifferenceToPushout`

 instead of 

`AddIsomorphismFromCokernelOfDiagonalDifferenceToPushout`.


CAP/gap/UniversalObjects.gd
```diff
@@ -1701,9 +1701,9 @@ DeclareOperation( "IsomorphismFromEqualizerOfDirectProductDiagramToFiberProduct"
 #! @Description
 #! The argument is a list of morphisms $D = ( \beta_i: P_i \rightarrow B )_{i = 1 \dots n}$.
 #! The output is a morphism
-#! $\bigoplus_{i=1}^n P_i \rightarrow B$
+#! $\bigoplus_{i=1}^{n-1} P_i \rightarrow B$
```
zickgraf, line 1704: I think it should be like this:
```suggestion
#! $\bigoplus_{i=1}^n P_i \rightarrow B^{n-1}$
```

zickgraf (changes requested):
Thanks for spotting this, those long names are horrible :D

Also "codiagonal" does not really make sense, as far as I understand things. So I suggest that as part of the other PR (which uses DirectProduct and Coproduct instead of direct sums) we should rename

DirectSumDiagonalDifference -> DirectProductDiagonalDifference
DirectSumCoDiagonalDifference -> CoproductDiagonalDifference

CAP/gap/UniversalObjects.gd
TKuh: > Thanks for spotting this, those long names are horrible :D

Absolutely. :grinning:

> Also "codiagonal" does not really make sense, as far as I understand things. So I suggest that as part of the other PR (which uses DirectProduct and Coproduct instead of direct sums) we should rename
> 
> DirectSumDiagonalDifference -> DirectProductDiagonalDifference DirectSumCoDiagonalDifference -> CoproductDiagonalDifference

Alright, just to be 100% sure: the isomorphisms like `IsomorphismFromFiberProductToKernelOfDiagonalDifference`
should also be renamed?

What about `FiberProductEmbeddingInDirectSum` / `DirectSumProjectionInPushout`? We will derive them via the new `JointPairwiseDifferences*` with products / coproducts, so I think renaming them is also sensible.

zickgraf: > I guess it's actually
> 
> `\bigoplus_{i=1}^n P_i \rightarrow \bigoplus_{i=1}^{n-1} B`

Yes, of course that's the more precise notation.

> Alright, just to be 100% sure: the isomorphisms like `IsomorphismFromFiberProductToKernelOfDiagonalDifference` should also be renamed?

What would you rename it to? No "DirectSum" appears in the string, so I think it could stay the way it is. We could of course be more precise and say `IsomorphismFromFiberProductToKernelOfDirectProductDiagonalDifference`, but I think that's a matter of taste, so you can decide.

> What about `FiberProductEmbeddingInDirectSum` / `DirectSumProjectionInPushout`? We will derive them via the new `JointPairwiseDifferences*` with products / coproducts, so I think renaming them is also sensible.

I don't have an overview over all the new derivations right now, so let's do this step by step and discuss this when we actually change the derivations.

If you like you can also include the original renaming in this PR again (then at least currently things would be consistent), I simply meant we should do another rename in the future. But you can decide how it's easier for you.

TKuh: > What would you rename it to? No "DirectSum" appears in the string, so I think it could stay the way it is. [...]

> I don't have an overview over all the new derivations right now, so let's do this step by step and discuss this when we actually change the derivations.

Alright, thanks.

> If you like you can also include the original renaming in this PR again (then at least currently things would be consistent), I simply meant we should do another rename in the future. But you can decide how it's easier for you.

It's easier to keep it out. If I now add a `Co` into `AddIsomorphismFromCokernelOfDiagonalDifferenceToPushout`, then once we rename:

`DirectSumDiagonalDifference -> DirectProductDiagonalDifference`
`DirectSumCoDiagonalDifference -> CoproductDiagonalDifference`

I will have to remove the `Co` again (since there won't be a `CoDiagonal` anymore). :D

PR #1188, opened 2022-12-06
changed "Opposite of ..." -> "Opposite( ... )"

mohamed-barakat:
reason: converging towards suggestive names of constructors should improve user experience

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2022.12-02",
+Version := "2022.12-03",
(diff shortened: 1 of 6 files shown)


zickgraf:
While we're at it: Should the constructor be named `OppositeCategory`?

mohamed-barakat:
> While we're at it: Should the constructor be named `OppositeCategory`?

Aren't you using `OppositeCategory` in the sense of `WasCreatedAsOppositeCategory`? If so, then we need to rename
* OppositeCategory -> WasCreatedAsOppositeCategory
* Opposite -> OppositeCategory

Correct?

zickgraf:
> Aren't you using `OppositeCategory` in the sense of `WasCreatedAsOppositeCategory`?

I'm currently using `OppositeCategory` as an attribute of a category which `WasCreatedAsOppositeCategory`. But I don't think this is actually needed anymore, and if it is, we can always find another name (worst case: the generic term `UnderlyingCategory`).

mohamed-barakat:
I suggest doing this as a subsequent commit. It doesn't look trivial.

PR #1192, opened 2022-12-06
Add some asserts

zickgraf:


CAP/gap/InstallAdds.gi
```diff
@@ -134,13 +134,19 @@ InstallGlobalFunction( CapInternalInstallAdd,
...
+                    replaced_filter_list,
                     { arg } -> CallFuncList( ValueGlobal( install_name ),
                             List( arg, function( ar ) if ValueGlobal( "IsJuliaObject" )( ar ) then return ValueGlobal( "ConvertJuliaToGAP" )( ar ); fi; return ar; end ) ) );
             
+            Assert( 0, record.install_convenience_without_category );
+            
             InstallOtherMethod( ValueGlobal( install_name ),
-                    CAP_INTERNAL_REPLACE_STRINGS_WITH_FILTERS_FOR_JULIA( filter_list ){[ 2 .. Length( replaced_filter_list ) ]},
+                    replaced_filter_list{[ 2 .. Length( replaced_filter_list ) ]},
                     { arg } -> CallFuncList( ValueGlobal( install_name ),
```
mohamed-barakat, line 150: Maybe one can extract this line as a common function define above the two installations.

zickgraf: Yes, but if I start deduplicating stuff then I would also unify `CAP_INTERNAL_REPLACE_STRING_WITH_FILTER_FOR_JULIA` and `CAP_INTERNAL_REPLACE_STRING_WITH_FILTER`, and then I would basically have rewritten the whole PR :D

PR #1194, opened 2022-12-07
added signature and logic for BasisOfRows/BasisOfColumns

mohamed-barakat:

PR #1196, opened 2022-12-07
installed derivations for Abelian categories

mohamed-barakat:
* IsomorphismFromImageObjectToKernelOfCokernel
* IsomorphismFromKernelOfCokernelToImageObject
* IsomorphismFromCoimageToCokernelOfKernel
* IsomorphismFromCokernelOfKernelToCoimage

PR #1197, opened 2022-12-07
Add(Co)ImageObject, AddImageEmbedding/CoimageProjectionWithGiven*

mohamed-barakat:
to category of rows over a field

addresses https://github.com/homalg-project/CAP_project/pull/1194#discussion_r1042180284

### DIFF
--- a/FreydCategoriesForCAP/PackageInfo.g
+++ b/FreydCategoriesForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "FreydCategoriesForCAP",
 Subtitle := "Freyd categories - Formal (co)kernels for additive categories",
-Version := "2022.12-01",
+Version := "2022.12-02",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
@@ -87,7 +87,7 @@ PackageDoc := rec(
 
 Dependencies := rec(
   GAP := ">= 4.12.1",
-  NeededOtherPackages := [ [ "CAP", ">= 2022.11-06" ],
+  NeededOtherPackages := [ [ "CAP", ">= 2022.12-07" ],
                            [ "MatricesForHomalg", ">= 2021.07-01" ],
                            [ "GradedRingForHomalg", ">=2019.08.07" ],
                            [ "MonoidalCategories", ">= 2022.06-01" ],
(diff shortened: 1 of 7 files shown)


zickgraf:
Comparing the compiled code here to the code in https://github.com/homalg-project/CAP_project/pull/1194 shows many improvements: often objects are computed in a simpler way, and the isomorphisms actually translate between `BasisOfRows( matrix )` and `SyzygiesOfRows( SyzygiesOfColumns( matrix ) )` (i.e. as expected there would have been inconsistencies if viewing them as equal).

`(Inverse)MorphismFromCoimageToImage` uses `IsomorphismFromKernelOfCokernelToImageObject` and thus has some changes, too. It could potentially be simplified by the following logic template:
```
CapJitAddLogicTemplate(
    rec(
        variable_names := [ "A", "B", "C" ],
        src_template := "RightDivide( B, A ) * RightDivide( A, C )",
        dst_template := "RightDivide( B, C )",
    )
);
```
(I think, please double-check :D )

mohamed-barakat:
I rebased this PR onto `master`, i.e., without the derivations from PR https://github.com/homalg-project/CAP_project/pull/1196. Several operations got higher weights and the compiled code for the following operations got more complicated:

* InverseMorphismFromCoimageToImageWithGivenObjects
* IsomorphismFromCoimageToCokernelOfKernel
* IsomorphismFromKernelOfCokernelToImageObject
* MorphismFromCoimageToImageWithGivenObjects

Without PR https://github.com/homalg-project/CAP_project/pull/1196 the suggested logic template had no effect.

zickgraf:
That's because `IsomorphismFromKernelOfCokernelToImageObject` is derived as the inverse of `IsomorphismFromKernelOfCokernelToImageObject` instead of dually to `IsomorphismFromImageObjectToKernelOfCokernel`. With "dually" I mean using the universality of the kernel (`KernelLift`) instead of the universality of the image (`UniversalMorphismFromImage`). You should still do this as part of #1196, sorry if my comment there was misleading.

zickgraf:
Ah, sorry, copy-paste mistake, it should read:

That's because `IsomorphismFromKernelOfCokernelToImageObject` is derived as the inverse of `IsomorphismFromImageObjectToKernelOfCokernel` instead of dually to the derivation of `IsomorphismFromImageObjectToKernelOfCokernel` which I linked in https://github.com/homalg-project/CAP_project/pull/1196.

zickgraf:
For future reference: The CI failures, e.g. in Algebroids, show that `SyzygiesOfRows( SyzygiesOfColumns( m ) ) <> BasisOfRows( m )` in general as expected. So https://github.com/homalg-project/CAP_project/pull/1194 would indeed have lead to inconsistencies.

PR #1210, opened 2022-12-19
Derive EmbeddingOfEqualizer / ProjectionOntoCoequalizer

TKuh:
Third part of https://github.com/homalg-project/CAP_project/pull/1124.


zickgraf (changes requested):
Please also derive the universal morphisms in the same way as `UniversalMorphismIntoFiberProduct` is derived from `DirectSumDiagonalDifference` and `KernelLift`, instead of relying on the general derivation as a (co)lift. This probably will not change anything for MatrixCategory, because there any (co)lift boils down to a `RightDivide` or `LeftDivide`, but in general computing a special kernel lift could be much faster than computing an arbitrary lift.

TKuh:
> Please also derive the universal morphisms in the same way as `UniversalMorphismIntoFiberProduct` is derived from `DirectSumDiagonalDifference` and `KernelLift`, instead of relying on the general derivation as a (co)lift. This probably will not change anything for MatrixCategory, because there any (co)lift boils down to a `RightDivide` or `LeftDivide`, but in general computing a special kernel lift could be much faster than computing an arbitrary lift.

Done. Thank you for the nice explanation!

CAP/gap/DerivedMethods.gi
```diff
@@ -2489,6 +2531,42 @@ AddDerivationToCAP( DirectSumProjectionInPushout,
...
+                      [ PreCompose, 1 ],
+                      [ IsomorphismFromCoequalizerToCokernelOfJointPairwiseDifferencesOfMorphismsFromCoproduct, 1 ] ],
+                    
+  function( cat, A, diagram )
+    local cokernel_proj_of_pairwise_differences;
+    
+    cokernel_proj_of_pairwise_differences := CokernelProjection( cat, JointPairwiseDifferencesOfMorphismsFromCoproduct( cat, A, diagram ) );
+    
+    return PreCompose( cat, cokernel_proj_of_pairwise_differences,
+                       IsomorphismFromCoequalizerToCokernelOfJointPairwiseDifferencesOfMorphismsFromCoproduct( cat, A, diagram ) );
```
zickgraf, line 2565:
```suggestion
                       IsomorphismFromCokernelOfJointPairwiseDifferencesOfMorphismsFromCoproductToCoequalizer( cat, A, diagram ) );
```

CAP/gap/DerivedMethods.gi
```diff
@@ -2489,6 +2531,42 @@ AddDerivationToCAP( DirectSumProjectionInPushout,
...
+                    
+  function( cat, A, diagram )
+    local kernel_of_pairwise_differences;
+    
+    kernel_of_pairwise_differences := KernelEmbedding( cat, JointPairwiseDifferencesOfMorphismsIntoDirectProduct( cat, A, diagram ) );
+    
+    return PreCompose( cat, IsomorphismFromEqualizerToKernelOfJointPairwiseDifferencesOfMorphismsIntoDirectProduct( cat, A, diagram ),
+                       kernel_of_pairwise_differences );
+    
+end : CategoryFilter := IsPreAbelianCategory,
```
zickgraf, line 2549: Why is `IsPreAbelianCategory` required?

CAP/gap/DerivedMethods.gi
```diff
@@ -618,6 +618,48 @@ AddDerivationToCAP( UniversalMorphismFromCoproduct,
...
+                            
+  function( cat, A, diagram, test_object, tau )
+    local joint_pairwise_differences, cokernel_colift;
+    
+    joint_pairwise_differences := JointPairwiseDifferencesOfMorphismsFromCoproduct( cat, A, diagram );
+    
+    cokernel_colift := CokernelColift( cat, joint_pairwise_differences, test_object, tau );
+    
+    return PreCompose( cat,
+             IsomorphismFromCokernelOfJointPairwiseDifferencesOfMorphismsFromCoproductToCoequalizer( cat, A, diagram ),
```
zickgraf, line 657:
```suggestion
             IsomorphismFromCoequalizerToCokernelOfJointPairwiseDifferencesOfMorphismsFromCoproduct( cat, A, diagram ),
```

zickgraf (changes requested):

CAP/gap/DerivedMethods.gi
TKuh: I think it's not. At the time I wrote the code, I thought it would be good style to add `CategoryFilter` and it survived my changes since then.

zickgraf (approved):

PR #1214, opened 2023-01-09
Only derive (co)image via kernel and cokernel if category is abelian

zickgraf:
This condition originally was part of the final derivation for `ImageEmbedding` (before `IsomorphismFromImageObjectToKernelOfCokernel` etc. existed) but was lost in d2734eeb70542ffcd3206107a972c198967a5d11.


sebastianpos (approved):
Yes, since the specifications in CAP say that an image is given by a universal mono-something factorization, it is indeed safe to assume that the category is abelian.

zickgraf:
Thanks for the fast review!

PR #1221, opened 2023-01-16
Handle `IsNTuple` explicitly in CAP_INTERNAL_REPLACE_STRING_WITH_FILTER

zickgraf:
@mohamed-barakat Please test if this fixes the issues you are seeing in https://github.com/homalg-project/CategoricalTowers/pull/12#issuecomment-1382857958.


mohamed-barakat:
There is still something broken. However, I made a PR to `CategoricalTowers` which will currently fail. For it to load CAP please rebase and bump `CAP`'s version to `2023.01-06`. Then you can see where it is broken.

PR #1224, opened 2023-01-17
Allow to resolve calls of operations without category as first argument

zickgraf:
if the data types of the arguments can be determined.

The recommended best practice remains to explicitly give the category as the first argument.


zickgraf:
Note @ all: This PR will expand the type inference, so don't be surprised if new warnings about functions without type signatures pop up. I have tried to keep the additional requirements minimal. In any case, the warnings should not affect the compilation result at all.

PR #1225, opened 2023-01-17
Replace calls to `DirectSum(Co)DiagonalDifference`

TKuh:
with calls to `JointPairwiseDifferencesOfMorphisms*`.

The last remnants of `DirectSumDiagonalDifference` are in (final) derivations of `IsomorphismFromFiberProductToKernelOfDiagonalDifference` and `IsomorphismFromKernelOfDiagonalDifferenceToFiberProduct` (and their duals).

I will remove them in a follow-up PR.


CAP/gap/DerivedMethods.gi
```diff
@@ -663,45 +663,65 @@ AddDerivationToCAP( UniversalMorphismFromCoequalizer,
 ##
 AddDerivationToCAP( UniversalMorphismIntoFiberProduct,
```
zickgraf, line 664: You should be able to simply remove those derivations. They should be covered by the derivation of `FiberProduct` from `DirectProduct` and `Equalizer`.

TKuh: I was so focused on replacing, that I didn't think about it. Thanks!

Is this also true for `FiberProductEmbeddingInDirectSum` and `DirectSumProjectionInPushout`? If yes, then it's not clear to me.

TKuh: > General answer: If `FiberProductEmbeddingInDirectSum` is more or less only used to derive `FiberProduct` (just like `DirectSumDiagonalDifference`) it should also be deleted. If it has actual applications, it should be generalized to `FiberProductEmbeddingInDirectProduct`

I could only find this instance, which looks like an application:

https://github.com/homalg-project/CAP_project/blob/96d770cff4bbd7768c72444d8dc21d81140b0102/GeneralizedMorphismsForCAP/examples/Sweep.g#L15-L16

And this, which I'm not sure about:

https://github.com/homalg-project/CAP_project/blob/96d770cff4bbd7768c72444d8dc21d81140b0102/AttributeCategoryForCAP/gap/AttributeCategory.gi#L490-L502

> Concrete answer: I cannot really tell without getting an overview over how `FiberProductEmbeddingInDirectSum` is used. If you think it's clear, feel free to vote for one solution or the other :-)

I would say, we postpone this until Thursday, where I would show it to @mohamed-barakat.
But either way, if `FiberProductEmbeddingInDirectSum` needs to use `DirectProduct` or can be deleted, I would do the changes in a separate PR, since both solutions are bigger changes. 

And I will remove the derivation of `UniversalMorphismIntoFiberProduct` together with the remnants of `DirectSumDiagonalDifference` in the next PR.

So I would say, this is ready for review.

zickgraf:
To avoid having to review changes which will be removed anyway, I have (force) pushed a commit which removes `DirectSumDiagonalDifference`, `FiberProductEmbeddingInDirectSum` and their duals together with all the isomorphisms involved. See https://github.com/homalg-project/CAP_project/pull/1230 for the changes required to keep the compiled code unchanged. What remains to be done:

1. Add back `FiberProductEmbeddingInDirectSum` in UniversalObjects.gi as a GAP (not CAP) operation for the example in GeneralizedMorphisms and for AttributeCategory.
2. Derive IsomorphismFromFiberProductToEqualizerOfDirectProductDiagram et al. from the universality of the equalizer et al., cf. https://github.com/homalg-project/CAP_project/blob/e083d25f102a7e28107ff599c968b954aadc7a8e/CAP/gap/DerivedMethods.gi#L2010-L2022 This can happen in a follow-up PR.
3. Review the changes in FreydCategory (@zickgraf).

@TKuh Could you handle the first two points? For the first I will first ask Sebastian.

zickgraf:
@sebastianpos You already agreed that we could remove `DirectSumDiagonalDifference`. Are you also okay with removing `FiberProductEmbeddingInDirectSum` and its dual?

Background:

These also seem to mainly serve the purpose of deriving FiberProduct from Kernel and DirectSum, which is superseded by deriving FiberProduct from Equalizer and DirectProduct (which in turn are derived from Kernel and DirectSum).

There are two remaining users of `FiberProductEmbeddingInDirectSum`, see the code in this comment: https://github.com/homalg-project/CAP_project/pull/1225#discussion_r1072899420. Thus, I suggest to re-introduce `FiberProductEmbeddingInDirectSum` as a convenience GAP operation which uses the universal property of the direct sum to compute `FiberProductEmbeddingInDirectSum` just like https://github.com/homalg-project/CAP_project/blob/e083d25f102a7e28107ff599c968b954aadc7a8e/CAP/gap/DerivedMethods.gi#L2596-L2608 does right now.

TKuh:
> @TKuh Could you handle the first two points? For the first I will first ask Sebastian.

Sure!

sebastianpos:
> @sebastianpos You already agreed that we could remove `DirectSumDiagonalDifference`. Are you also okay with removing `FiberProductEmbeddingInDirectSum` and its dual?

Yes, I'm okay with that. And I think it's also good to have them available as convenience methods, since viewing the fiberproduct as a subobject of the product is a nice entity to compute (for the end user).

FreydCategoriesForCAP/gap/FreydCategoriesDerivedMethods.gi
```diff
@@ -368,11 +370,19 @@ AddFinalDerivationBundle( # WeakBiFiberProductMorphismToDirectSum,
...
+    diagram := [ alpha, beta ];
     
-    return WeakKernelEmbedding( cat, diagonal_difference );
+    direct_sum_diagram := List( diagram, Source );
+    
+    direct_sum := DirectSum( cat, direct_sum_diagram );
+    
+    list_of_morphisms := List( [ 1 .. 2 ], i -> PreCompose( cat, ProjectionInFactorOfDirectSumWithGivenDirectSum( cat, direct_sum_diagram, i, direct_sum ), diagram[ i ] ) );
+    
+    joint_pairwise_difference := JointPairwiseDifferencesOfMorphismsIntoDirectProduct( cat, direct_sum, list_of_morphisms );
```
zickgraf, line 383: I suggest spelling out everything explicitly here, i.e. getting rid of lists and `JointPairwiseDifferencesOfMorphismsIntoDirectProduct` and simply subtracting the two projections composed with `alpha` and `beta`.

zickgraf (changes requested):

CAP/gap/UniversalObjects.gi
```diff
@@ -503,6 +503,74 @@ InstallGlobalFunction( FiberProduct,
     
 end );
 
+##
+InstallOtherMethod( FiberProductEmbeddingInDirectProduct,
```
zickgraf, line 507: InstallOtherMethod -> InstallMethod

CAP/gap/UniversalObjects.gi
```diff
@@ -503,6 +503,74 @@ InstallGlobalFunction( FiberProduct,
...
+        [ IsList ],
+        
+  function( diagram )
+    
+    local sources_of_diagram, cat, test_source;
+    
+    sources_of_diagram := List( diagram, Source );
+    
+    cat := CapCategory( diagram[1] );
```
zickgraf, line 516: This should simply extract the category and then call `FiberProductEmbeddingInDirectProduct` with two arguments.

zickgraf (changes requested):

TKuh:
I just noticed [these declarations](https://github.com/homalg-project/CAP_project/blob/fabd4ac4f4263b496525d465d13b7e2924191910/CAP/gap/UniversalObjects.gd#L1742-L1760). They should be removed, too, right? (Same for the duals)

CAP/gap/UniversalObjects.gi
```diff
@@ -782,6 +834,74 @@ InstallMethod( Pushout,
...
+InstallOtherMethod( PushoutProjectionFromCoproduct,
+        [ IsList ],
+                    
+  function( diagram )
+    
+    local ranges_of_diagram, cat, test_sink;
+    
+    ranges_of_diagram := List( diagram, Range );
+    
+    cat := CapCategory( diagram[1] );
```
zickgraf, line 847: Please adjust this in the same way as `FiberProductEmbeddingInDirectProduct`.

Also, please remove the empty line before `local ...;` in all the newly added functions.

zickgraf:
> I just noticed [these declarations](https://github.com/homalg-project/CAP_project/blob/fabd4ac4f4263b496525d465d13b7e2924191910/CAP/gap/UniversalObjects.gd#L1742-L1760). They should be removed, too, right? (Same for the duals)

Ah, of course, thanks for noticing. I though I had grepped for this, maybe I forgot to commit.

CAP/gap/UniversalObjects.gi
TKuh: Sorry, I forgot about the duals...
It's fixed now.

zickgraf (approved):

mohamed-barakat:
Thanks to all of you.

PR #1232, opened 2023-01-26
Display warning if `category_as_first_argument` is not set explicitly

zickgraf:
This is a preparatory step for changing the default to `true` in the future.

@ all: Should this cause too much trouble for now, we can of course disable the warning again. But let's see how many packages are affected by this.

### DIFF
--- a/ActionsForCAP/PackageInfo.g
+++ b/ActionsForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "ActionsForCAP",
 Subtitle := "Actions and Coactions for CAP",
-Version := "2023.01-01",
+Version := "2023.01-02",
 Date := Concatenation( "01/", ~.Version{[ 6, 7 ]}, "/", ~.Version{[ 1 .. 4 ]} ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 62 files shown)


mohamed-barakat:
Would it make sense to let `CreateCapCategory` to set it to `false` instead of doing it manually in all these examples?

zickgraf:
No: I want to change the default value to `true` in the future (and remove this warning again, so "modern" categories do not have to set it manually). To do this without causing errors, all categories which require the value to be `false` have to set it explicitly. This PR achieves this by displaying a warning for such categories.

zickgraf:
As expected, there aren't too many packages affected by this. I have already opened some PRs. The remaining packages are:

* ExteriorPowersCategories
* IntrinsicGradedModules
* IntrinsicModules
* CategoriesWithAmbientObjects
* HigherHomologicalAlgebra (might also be failing for unrelated reasons, I'm not sure)

@mohamed-barakat @kamalsaleh Could you fix those packages as soon as you find the time?

mohamed-barakat:
> As expected, there aren't too many packages affected by this. I have already opened some PRs. The remaining packages are:
> 
> * ExteriorPowersCategories
> * IntrinsicGradedModules
> * IntrinsicModules
> * CategoriesWithAmbientObjects
> * HigherHomologicalAlgebra (might also be failing for unrelated reasons, I'm not sure)

In my case, only `CategoriesWithAmbientObjects` needed an update.

PR #1233, opened 2023-01-27
Simplify handling of category_as_first_argument

zickgraf:


zickgraf:
Note: Technically, this change is not backwards compatible. If `category_as_first_arguments = false`, functions added for CAP operations with a single argument (the category) must now have 0 arguments (or a variable number of arguments). Previously, this was already the case for some commonly used operations like `ZeroObject` but not for rarely used operations like `IsomorphismFromZeroObjectToInitialObject`. But since those operations are rarely used, I expect nothing to break, and the old behavior could even be called a bug.

PR #1247, opened 2023-02-08
fix and complete commit 3956135b

mohamed-barakat:
avoid broken Add-methods in AttributeCategory

add *WithGiven* to no_install_list if object-part-operation is not supported by AttributeCategory


zickgraf (approved):
Thanks!

PR #1248, opened 2023-02-09
Use only the category for method selection of CAP operations

zickgraf:
The main commit is the 4th commit. For example, the method for `IdentityMorphism` is now installed with filters `[ CategoryFilter( cat ), IsObject ]` instead of `[ CategoryFilter( cat ), ObjectFilter( cat ) ]` (convenience methods are unaffected by this). This has several advantages:

1. It corresponds to how CompilerForCAP selects methods for CAP operations.
2. Method selection might be a bit faster (I have not checked).
3. It allows to model bare categories (categories with primitive data structures like integers for objects) without further modifications (see the last commit).
4. It automatically covers some cases of wrong input which previously were caught by special fallbacks, see the removed code in the 4th commit. For example, when calling `IsEqualForObjects` with two objects in different categories one now automatically gets "Error, The CapCategory of the 3-th argument of the function IsEqualForObjects of the category named ... is not identical to the category named ...."
5. It emphasizes that the category is the single source of truth.

It also has some disadvantages:
1. One cannot install convenience methods for CAP operations with the same number of arguments as the actual CAP operation anymore. However, I've never seen such convenience methods, they would be forbidden when using CompilerForCAP anyway, and I have added a warning catching this case.
2. The type checking now only happens in the input sanity checks. So when input sanity checks are disabled not even basic filters are tested. However, I would say that this should actually be expected by the user when disabling input sanity checks :D

@ all: Are you fine with this change?

### DIFF
--- a/CAP/examples/TerminalCategoryWithMultipleObjects.g
+++ b/CAP/examples/TerminalCategoryWithMultipleObjects.g
@@ -10,7 +10,7 @@ T := TerminalCategoryWithMultipleObjects( );
 Display( T );
 #! A CAP category with name TerminalCategoryWithMultipleObjects( ):
 #! 
-#! 63 primitive operations were used to derive 304 operations for this category
+#! 63 primitive operations were used to derive 306 operations for this category
 #! which algorithmically
 #! * IsCategoryWithDecidableColifts
 #! * IsCategoryWithDecidableLifts
--- a/CAP/examples/TerminalCategoryWithSingleObject.g
+++ b/CAP/examples/TerminalCategoryWithSingleObject.g
@@ -10,7 +10,7 @@ T := TerminalCategoryWithSingleObject( );
 Display( T );
 #! A CAP category with name TerminalCategoryWithSingleObject( ):
 #! 
-#! 63 primitive operations were used to derive 304 operations for this category
+#! 63 primitive operations were used to derive 306 operations for this category
 #! which algorithmically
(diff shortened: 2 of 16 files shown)


mohamed-barakat:
Sounds nice. Do we need to touch existing code?

zickgraf:
> Sounds nice. Do we need to touch existing code?

No, except when using convenience operations with the same name and number of arguments as a CAP operation, but I have not seen those in any of our code.

zickgraf:
Thinking more about this, I do not like the idea anymore.

Ad 1. In an ideal world, CompilerForCAP would have all the type information.
Ad 3. I have a new interpretation: "Bare categories" are actually "set theoretic categories", and thus have all the problems that come with a set-theoretic modelling of categories, for example possibly non-disjoint hom sets. I still plan to add "set theoretic categories" to CAP for didactic purposes, but I do not expect widespread support/applications anymore.
Ad 4: Such errors are actually well covered by GAP's "No method found" errors if one knows how to use them.
Ad 5: That the category is the single source of truth is a technical detail of the implementation, and strictly speaking not true: We have special installations for example for PreCompose with identity morphisms in some categories. This is not widely used and partially obsolete by CompilerForCAP, but shows that there might be a need on some cases. Also, in the future we might want to support multiple homomorphism structures over different ranges. Then we also would need to dispatch on two categories.

Issue #1249, opened 2023-02-09
When using only_primitive_operations in OppositeCategory prevents the installation of the hom structure if the hom structure in the underlying category is derived


mohamed-barakat:

mohamed-barakat:
This issue should be addressed before https://github.com/homalg-project/CAP_project/pull/1240 is merged.

zickgraf:
Does this problem get worse with #1240? If yes, why? If no, why do you want to address it together with #1240?

mohamed-barakat:
Pr #1240 had multiple versions and is not yet in final form so it is difficult to answer the question at this point:

But here is the issue with the opposite category constructor further detailed:

In line
https://github.com/homalg-project/CAP_project/blob/5d2f3fa735cb6095a2f9d5e044a1738b31f2ff4a/CAP/gap/OppositeCategory.gi#L72
the range category is manually set.

1. Currently (i.e., without #1240), this prevents the final derivations from triggering since they all require the range category not to be set, e.g. when the opposite category happens to be cartesian or monoidal closed.
2. On the other side, if the operations of the homomorphism structures of the underlying category are not primitively installed (e.g., themselves installed by a final derivation), they are not installed if the option  `only_primitive_operations` is set to true.
3. The wrapper category constructor applied to this opposite category will find the range category set but won't find the operations for the homomorphism structure, so it will not set the range category for the wrapper category and the final derivations might now be triggered for the latter.

This can be fixed for the current master branch (093bbcd1032bc93cc7ed60ee61f8a1cd28e90be1) and prior to PR #1240 by preventing the opposite category constructor from setting the range category if it is not able to install the operations of the homomorphism structure.

Now PR #1240 reverses the situation and expects the range category to be set for the final derivations to be triggered.

zickgraf:
Ah, thanks for the details, I wasn't aware that WrapperCategory also plays a role here.

PR #1251, opened 2023-02-11
fixed SetRangeCategoryOfHomomorphismStructure for the opposite category

mohamed-barakat:
SetRangeCategoryOfHomomorphismStructure only if the operations of the homomorphism structure can be installed

resolves https://github.com/homalg-project/CAP_project/issues/1249

### DIFF
--- a/CAP/examples/OppositeCategory.g
+++ b/CAP/examples/OppositeCategory.g
@@ -12,8 +12,9 @@ ListKnownCategoricalProperties( op );
 #!   "IsAbelianCategoryWithEnoughProjectives", "IsAdditiveCategory",
 #!   "IsBraidedMonoidalCategory", "IsClosedMonoidalCategory",
 #!   "IsCoclosedMonoidalCategory", "IsEnrichedOverCommutativeRegularSemigroup",
-#!   "IsLinearCategoryOverCommutativeRing", "IsMonoidalCategory",
-#!   "IsPreAbelianCategory", "IsRigidSymmetricClosedMonoidalCategory",
+#!   "IsEquippedWithHomomorphismStructure", "IsLinearCategoryOverCommutativeRing",
+#!   "IsMonoidalCategory", "IsPreAbelianCategory",
+#!   "IsRigidSymmetricClosedMonoidalCategory",
 #!   "IsRigidSymmetricCoclosedMonoidalCategory", "IsSkeletalCategory",
 #!   "IsStrictMonoidalCategory", "IsSymmetricClosedMonoidalCategory",
 #!   "IsSymmetricCoclosedMonoidalCategory", "IsSymmetricMonoidalCategory" ]
--- a/CAP/gap/OppositeCategory.gi
+++ b/CAP/gap/OppositeCategory.gi
@@ -59,8 +59,9 @@ end );
 BindGlobal( "CAP_INTERNAL_INSTALL_OPPOSITE_ADDS_FROM_CATEGORY",
   
   function( opposite_category, category )
-    local only_primitive_operations, recnames, current_recname, current_entry, dual_operation_name,
-          filter_list, input_arguments_names, return_type, func_string,
+    local only_primitive_operations, recnames, list_of_installed_operations,
+          operations_of_homomorphism_structure, operations_of_external_hom,
+          current_recname, current_entry, dual_operation_name, filter_list, input_arguments_names, return_type, func_string,
           dual_preprocessor_func_string, preprocessor_string, dual_arguments, tmp,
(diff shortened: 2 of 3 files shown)


CAP/examples/OppositeCategory.g
```diff
@@ -7,16 +7,20 @@ LoadPackage( "LinearAlgebraForCAP" );
 QQ := HomalgFieldOfRationals();;
 vec := MatrixCategory( QQ );;
 op := Opposite( vec );;
-ListKnownCategoricalProperties( op );
```
zickgraf, line 10: Please keep the old version so we don't have to update even more numbers when adding or removing derivations. Also, this is only one special case. If you want to compare the numbers later on (as the commit message suggests) I would suggest adding an example file which covers all cases but is not executed as part of the tests.

CAP/gap/OppositeCategory.gi
```diff
@@ -89,6 +91,49 @@ BindGlobal( "CAP_INTERNAL_INSTALL_OPPOSITE_ADDS_FROM_CATEGORY",
...
+      [ "MorphismsOfExternalHom", ## still in Toposes but should be moved to CAP
+        ];
+    
+    if HasRangeCategoryOfHomomorphismStructure( category ) and
+       ( not IsEmpty( Intersection( list_of_installed_operations, operations_of_homomorphism_structure ) ) or
+         not IsEmpty( Intersection( list_of_installed_operations, operations_of_external_hom_in_matrix_category ) ) or
+         not IsEmpty( Intersection( list_of_installed_operations, operations_of_external_hom_in_fin_sets ) ) ) then
+        
+        SetRangeCategoryOfHomomorphismStructure( opposite_category, RangeCategoryOfHomomorphismStructure( category ) );
+        SetIsEquippedWithHomomorphismStructure( opposite_category, true );
```
zickgraf, line 133: I don't think this fits into the new system saying "the range category is type information which should be explicitly provided because we don't want to guess what the user wants". For example, "BasisOfExternalHom" and "CoefficientsOfMorphism" might be installed in the underlying category but not be used for deriving the homomorphism structure, but this code would still blindly use its range (and trigger or not trigger some derivation based on this guess).

I suggest the following: keep the case of `operations_of_homomorphism_structure` (if the user wants us to dualize the hom structure, that's fine), but drop the other special cases. If one wants to have a hom structure in those special cases, one must set the intended range explicitly. If this happens at many places, we could add a convenience function which accepts the range category as an additional argument.

CAP/gap/OppositeCategory.gi
```diff
@@ -338,6 +374,10 @@ BindGlobal( "CAP_INTERNAL_INSTALL_OPPOSITE_ADDS_FROM_CATEGORY",
         
     od;
     
+    if HasRangeCategoryOfHomomorphismStructure( opposite_category ) then
+        Assert( 0, [ ] = CheckConstructivenessOfCategory( opposite_category, "IsEquippedWithHomomorphismStructure" ) );
```
zickgraf, line 378: If the underlying category has itself a partially constructive hom structure, the opposite category will too, so this assertion won't hold true.

zickgraf (changes requested):

CAP/gap/OppositeCategory.gi
zickgraf: With my suggestion the tests in CategoricalTowers (excluding Algebroids and FunctorCategories which I have not tested) seem to still succeed.

CAP/examples/OppositeCategory.g
mohamed-barakat: Done.

CAP/gap/OppositeCategory.gi
mohamed-barakat: Done.

CAP/gap/OppositeCategory.gi
mohamed-barakat: Done.

CAP/gap/OppositeCategory.gi
mohamed-barakat: FunctorCategories now fails miserably :) I hope I can get it fixed.

mohamed-barakat: False alarm: After setting FinSetsForCAP to the adjusted branch also FunctorCategories passed 👍

mohamed-barakat: Reminder: The above testing refers to PR #1240 (the "next" PR), not to this one.

zickgraf (commented):
I have some suggestions but to not block this PR further I will do those in a follow-up PR.

zickgraf (approved):

PR #1253, opened 2023-02-12
start testing InternalExteriorAlgebraForCAP

mohamed-barakat:


mohamed-barakat:
It should be easy to bring CAP over 80% coverage and nice to reach 90%.

zickgraf:
Nice :-)

> It should be easy to bring CAP over 80% coverage and nice to reach 90%.

Yes, there are many example files which are not actually executed.

PR #1259, opened 2023-02-21
Use modeling_* options in ModulePresentationsForCAP

### LINKED ISSUE #980: Use new options for explicit conversion of data structures in `WrapperCategory` everywhere
#964 introduces new options for explicit conversion of data structures to `WrapperCategory` but does not use the new options at all places where `WrapperCategory` is currently used. These places are marked with TODO in the code. The old code is still valid, but using the new options would be nice for consistency.

TKuh:
Closes #980

I don't know how to fix the compiler warnings/error. :(

### DIFF
--- a/ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
+++ b/ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
@@ -15,60 +15,86 @@ InstallMethod( LeftPresentationsAsFreydCategoryOfCategoryOfRows,
                [ IsHomalgRing ],
                
   function( ring )
-    local category_of_rows, freyd, object_constructor, object_datum, morphism_constructor, morphism_datum, wrapper;
+    local rows, freyd, object_constructor, modeling_tower_object_constructor, object_datum, modeling_tower_object_datum, morphism_constructor, modeling_tower_morphism_datum, morphism_datum, modeling_tower_morphism_constructor, wrapper;
     
-    category_of_rows := CategoryOfRows( ring : FinalizeCategory := true );
+    rows := CategoryOfRows( ring : FinalizeCategory := true );
     
-    freyd := FREYD_CATEGORY( category_of_rows : FinalizeCategory := true );
+    freyd := FREYD_CATEGORY( rows : FinalizeCategory := true );
     
-    object_constructor := function( cat, object_in_freyd )
+    ##
+    object_constructor := function( cat, matrix)
         
-        return ObjectifyObjectForCAPWithAttributes( rec( ), cat,
-                                                    UnderlyingMatrix, UnderlyingMatrix( RelationMorphism( object_in_freyd ) )
-        );
+        return ObjectifyObjectForCAPWithAttributes( rec(), cat,
+                                                    UnderlyingMatrix, matrix );
+    
+    end;
+    
+    ##
+    modeling_tower_object_constructor := function( cat, matrix )
+        local freyd, rows;
+        
+        freyd := ModelingCategory( cat );
+        
+        rows := UnderlyingCategory( freyd );
+        
+        return FreydCategoryObject( freyd,
+                                    AsCategoryOfRowsMorphism( rows, matrix ) );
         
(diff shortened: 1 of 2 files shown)


zickgraf:
> I don't know how to fix the compiler warnings/error. :(

The compiler tries to take `RankOfObject` of a Freyd object, which is not defined. So probably something is wrong in the new constructors or destructors. You can try to run examples without precompiled code and see if this reproduces the problem independent of CompilerForCAP.

ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
```diff
@@ -15,60 +15,86 @@ InstallMethod( LeftPresentationsAsFreydCategoryOfCategoryOfRows,
                [ IsHomalgRing ],
                
   function( ring )
-    local category_of_rows, freyd, object_constructor, object_datum, morphism_constructor, morphism_datum, wrapper;
+    local category_of_rows, freyd, object_constructor, modeling_tower_object_constructor, object_datum, modeling_tower_object_datum, morphism_constructor, modeling_tower_morphism_datum, morphism_datum, modeling_tower_morphism_constructor, wrapper;
     
     category_of_rows := CategoryOfRows( ring : FinalizeCategory := true );
     
-    freyd := FREYD_CATEGORY( category_of_rows : FinalizeCategory := true );
+    freyd := FreydCategory( category_of_rows : FinalizeCategory := true );
```
zickgraf, line 22: Did you need this change in the context of this PR? If yes, why? If no, please revert it. Using attributes when constructing towers for compilation is dangerous in the sense that the compilation result might depend on whether you have created this tower before or not.

ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
```diff
@@ -15,60 +15,86 @@ InstallMethod( LeftPresentationsAsFreydCategoryOfCategoryOfRows,
                [ IsHomalgRing ],
                
   function( ring )
-    local category_of_rows, freyd, object_constructor, object_datum, morphism_constructor, morphism_datum, wrapper;
+    local category_of_rows, freyd, object_constructor, modeling_tower_object_constructor, object_datum, modeling_tower_object_datum, morphism_constructor, modeling_tower_morphism_datum, morphism_datum, modeling_tower_morphism_constructor, wrapper;
     
     category_of_rows := CategoryOfRows( ring : FinalizeCategory := true );
```
zickgraf, line 20: Please keep this in sync with the nomenclature used below, i.e. either don't rename it below or also rename it here. The same name should be used because the variables refer to the identical category.

ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
```diff
@@ -129,60 +154,86 @@ InstallMethod( RightPresentationsAsFreydCategoryOfCategoryOfColumns,
...
-    object_datum := function( cat, obj )
-      local freyd, category_of_columns, relation_morphism;
+    ##
+    modeling_tower_object_constructor := function( cat, matrix )
+        local freyd, columns;
         
         freyd := ModelingCategory( cat );
         
-        category_of_columns := UnderlyingCategory( freyd );
+        columns := UnderlyingCategory( freyd );
```
zickgraf, line 177: We usually call the short form of `CategoryOfColumns` just `cols`.

zickgraf (changes requested):
Thanks, some small comments below.

ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
TKuh: Thank you, I wasn't aware of this. I reverted it.

zickgraf (approved):
Thanks!

PR #1265, opened 2023-02-28
Warn if a derivation depends on other categories but does not test if those are available

zickgraf:
@mohamed-barakat @kamalsaleh  This might display warnings in your packages which can be fixed by adding correct CategoryFilters similar to the CategoryFilters added in this PR.


zickgraf:
Edit: Also pinging @kamalsaleh instead of pinging @mohamed-barakat twice :D

PR #1285, opened 2023-03-24
Miscellaneous changes

zickgraf:

@sebastianpos @mohamed-barakat @kamalsaleh This PR starts using DisplayString instead of Display at some places (see https://github.com/homalg-project/homalg_project/pull/583#issuecomment-1480710290 for the motivation). If `Display` in a tower of categories does not work as before, you probably have to install `DisplayString` instead of `Display` in the lower levels of the tower.


mohamed-barakat:
Finally 😄

mohamed-barakat (approved):

Issue #1287, opened 2023-03-27
Warning in Locales


mohamed-barakat:
PR https://github.com/homalg-project/CAP_project/pull/1286 now causes the following warning in Locales:
```gap
WARNING: the CAP operation UniqueMorphism was declared as an attribute but gets more than two arguments and can thus probably not be used as an attribute.
```
Is it possible to turn `UniqueMorphism` in CAP into an operation or change its name to something else?

zickgraf:
Ah, I feared that something like this might happen :/

Background: `Coproduct` was declared as an attribute of `IsList`, which does not make sense because `IsList` is not an attribute storing rep, and this caused issues in Julia. To avoid this in the future, I have added warnings for detecting if something was declared as an attribute but cannot be used as one. Now the problem is: Something might be declared as an attribute in a sensible way and later be reused as a CAP operation with arguments where being an attribute does not make sense. This happens for `UniqueMorphism`, which is an attribute of `IsCapTerminalCategoryWithSingleObject`, but as a CAP operation is applied to two morphisms, i.e. is not used as an attribute there.

Is this urgent? If yes, I can simply remove the warning right now, else I will think of a better solution in the course of the day.

mohamed-barakat:
> Background: `Coproduct` was declared as an attribute of `IsList`, which does not make sense because `IsList` is not an attribute storing rep, and this caused issues in Julia. To avoid this in the future, I have added warnings for detecting if something was declared as an attribute but cannot be used as one. Now the problem is: Something might be declared as an attribute in a sensible way and later be reused as a CAP operation with arguments where being an attribute does not make sense. This happens for `UniqueMorphism`, which is an attribute of `IsCapTerminalCategoryWithSingleObject`, but as a CAP operation is applied to two morphisms, i.e. is not used as an attribute there.

I see, good point :)

> Is this urgent? If yes, I can simply remove the warning right now, else I will think of a better solution in the course of the day.

Unfortunately yes. I need the CI to work to get the co-Yoneda stuff into the main branch.

PR #1288, opened 2023-03-28
Remove two warnings for now

zickgraf:
This is a hotfix for #1287. A proper fix should follow.


mohamed-barakat:
Thanks, Fabian.

PR #1311, opened 2023-04-28
Refactor warnings in InstallAdds

zickgraf:


zickgraf:
In particular, one can now call primitively installed operations of a category before finalizing the category without triggering a warning. This should always be safe and is useful for demonstration purposes.

mohamed-barakat:
> In particular, one can now call primitively installed operations of a category before finalizing the category without triggering a warning. This should always be safe (except if the primitive operations are overwritten by derivations, see the commit messages) and is useful for demonstration purposes.

Very nice!

PR #1316, opened 2023-05-02
Introduce ReinterpretationOfCategory

zickgraf:


zickgraf:
This includes and supersedes #989. @mohamed-barakat I would be happy for another round of review of the documentation. But since this is now quite big I don't want to block this PR, so I will merge without review and we can improve the documentation later. Of course I have included the suggestions of the review in #989.

PR #1297, opened 2023-04-11
Use `dual_with_given_objects_reversed`

TKuh:
Fixes #1295

Issue #1295, opened 2023-04-11
BraidingInverseWithGivenTensorProducts has wrong dual_preprocessor_func


zickgraf:
Steps to reproduce:
```
LoadPackage( "FreydCategoriesForCAP" );
LoadPackage( "LazyCategories" );

QQ := HomalgFieldOfRationals();

cat := CategoryOfRows( QQ : no_precompiled_code );

lazy := LazyCategory( cat : only_primitive_operations );

op := Opposite( lazy : only_primitive_operations );

obj1 := ObjectConstructor( cat, 1 ) / lazy;
obj2 := ObjectConstructor( cat, 2 ) / lazy;

op_obj1 := ObjectConstructor( op, obj1 );
op_obj2 := ObjectConstructor( op, obj2 );

Assert( 0, IsWellDefinedForObjects( op_obj1 ) );
Assert( 0, IsWellDefinedForObjects( op_obj2 ) );

mor := BraidingInverse( op_obj1, op_obj2 );
```
throws an error.

@TKuh Could you look at the preprocessor functions for MonoidalCategories and CartesianCategories (mostly added in https://github.com/homalg-project/CAP_project/pull/773) again and maybe enhance the test suite to catch errors like the one above? Using `dual_with_given_objects_reversed` (see the method record entry of `TensorProductOnMorphismsWithGivenTensorProducts`) could be used instead of some of the preprocessor funcs.

TKuh:
Thank you for the hint about `dual_with_given_objects_reversed`, that's useful. 

I'm wondering about the best approach to improve the tests. I guess there's no way around `LazyCategories` to be really sure, right?

- If I use `LazyCategories` directly in the test methods (e.g. [here](https://github.com/homalg-project/CAP_project/blob/master/MonoidalCategories/gap/BraidedMonoidalCategoriesTest.gi)), then `MonoidalCategories` depends on `LazyCategories` ...
- Moving the tests to `LazyCategories` and enhance them there would be fine, but I think they should stay closer to `MonoidalCategories`.
- Last option: introduce new tests similar to your code above directly in `LazyCategories`.

Any other ideas? If not, the last point seems the cleanest to me.

zickgraf:
Thanks for the quick response and PR!

Yes, I currently see no way around using `LazyCategories` and I agree that those tests should happen there to avoid additional dependencies. There are already tests which I assume should catch this:
[...]

PR #1318, opened 2023-05-03
Enhance `Opposite`-tests

TKuh:
Note: this should make `LazyCategories` fail due to #1295 and succeed after PR https://github.com/homalg-project/CAP_project/pull/1297.


CAP/gap/OppositeCategory.gi
```diff
@@ -38,6 +38,27 @@ InstallMethod( Opposite,
...
 end );
 
+##
+InstallOtherMethod( Opposite,
+                    [ IsCapCategory, IsCapCategoryMorphism ],
+                    
+  function( opposite_cat, morphism )
+    local source, range;
+    
+    if not HasOpposite( opposite_cat ) then
```
zickgraf, line 48: I think this check is not what we want: If you compute `Opposite( C )`, the also `C` as the attribute `Opposite`. You could check for `WasCreatedAsOppositeCategory`, but I would suggest to simply remove this check and rely on `ObjectConstructor` and `MorphismConstructor` for checks.

zickgraf (changes requested):

zickgraf (approved):
Thanks a lot!

zickgraf:
Test run in CategoricalTowers: https://github.com/homalg-project/CategoricalTowers/actions/runs/4917003133 Let's hope it fails :D

zickgraf:
FinSetsForCAP is already failing: https://github.com/homalg-project/FinSetsForCAP/actions/runs/4917118682/jobs/8781786322 :-)

@TKuh Please rebase #1297.

mohamed-barakat:
Wonderful, thank you very much @TKuh.

zickgraf:
LazyCategories is also failing :-)

PR #1320, opened 2023-05-08
Explicitly pass the opposite category as second argument to *Test

zickgraf:
This allows to execute those tests in Julia. Creating new categories inside functions and calling operations of these categories in the same function is not possible in Julia: When calling such a function, Julia compiles it in a context where the categorical operations are not available yet.

More places have to be adjusted and I will wait for the current PRs in this context to be merged first.


TKuh (approved):
Thank you!

PR #1331, opened 2023-05-19
fix bug in LinearAlgebraForCAP & bump to V2023.05-04

kamalsaleh:


zickgraf (approved):
Thanks for spotting this! I will create a PR which runs the tests without precompiled code to make sure this code is covered by tests.

PR #1332, opened 2023-05-19
Refactor implications of `IsBicartesian(Co)ClosedCategory`

zickgraf:
Make `IsBicartesianClosedCategory` equivalent to
`IsBicartesianCategory and IsCartesianClosedCategory`.
In particular, the implication
`IsBicartesianClosedCategory => IsBicartesianCategory`
is now explicit and does not need the implication
`IsCartesianCategory and IsCocartesianCategory => IsBicartesianCategory`.
This is useful for Julia.

Also, due to a typo, the implication
`IsBicartesianClosedCategory => IsCartesianCategory`
was installed twice. The second installation should have been
`IsBicartesianCoclosedCategory => IsCartesianCategory`.

The first commit is independent and only simplifies an example, see the commit message.

Also note that the link https://ncatlab.org/nlab/show/bicartesian+coclosed+category in the comments does not work. Did it ever work or was the plan to link there if it the page will be created one day?


mohamed-barakat (approved):

zickgraf:
Codecov is not updating correctly, the coverage actually does not change.

PR #1338, opened 2023-06-03
Propagate argument names of functions from src_template to dst_template

zickgraf:


zickgraf:
@mohamed-barakat This PR will require recompilations in CategoricalTowers (not for validity but for making the tests pass due to changed variable names). I have lost track of all the things that are getting precompiled in the tests of CategoricalTowers. Could you recompile those once this PR is merged? (The merge is not urgent, so we can wait until you have time to take care of CategoricalTowers)

PR #1340, opened 2023-06-05
simplified code

mohamed-barakat:


ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
```diff
@@ -29,11 +29,7 @@ InstallMethod( LeftPresentationsAsFreydCategoryOfCategoryOfRows,
     end;
     
     ##
-    object_datum := function( cat, obj )
-        
-        return UnderlyingMatrix( obj );
-        
-    end;
+    object_datum := { cat, obj } ->  UnderlyingMatrix( obj );
```
zickgraf, line 32:
```suggestion
    object_datum := { cat, obj } -> UnderlyingMatrix( obj );
```

ModulePresentationsForCAP/gap/ModulePresentationsAsFreydCategoryOfCategoryOfRowsOrColumns.gi
```diff
@@ -172,11 +164,7 @@ InstallMethod( RightPresentationsAsFreydCategoryOfCategoryOfColumns,
...
```
zickgraf, line 167:
```suggestion
    object_datum := { cat, obj } -> UnderlyingMatrix( obj );
```

zickgraf (commented):
Two things:

1. Debugging is easier if one can quickly add `Error`s to functions. That's why I also like the existing verbose style. But in the end I don't have a strong opinion.
2. Is there a reason why you did not modify `modeling_tower_object_datum` and `modeling_tower_morphism_datum`?

PR #1341, opened 2023-06-05
renamed local variable: wrapper -> left/right_presentations

mohamed-barakat:


mohamed-barakat:
> My question here is: If I recall correctly, you said that due to the unification using `ObjectConstructor` etc., `LeftPresentationsAsFreydCategoryOfCategoryOfRows` and `RightPresentationsAsFreydCategoryOfCategoryOfColumns` can be made very similar. This PR now seems to go into a different direction.

Similar, only apart from automatically renaming variables for better understanding the context.

PR #1357, opened 2023-06-20
Make sure data_type of arguments is known when resolving known non-categorical methods

zickgraf:
Ensuring that the function call has a data type is not enough because the arguments might have changed since the last application of CapJitInferredDataTypes.

@mohamed-barakat This should fix the CI of CategoricalTowers.


mohamed-barakat:
Thank you very much. You triggered the CI and it turned green again 👍🏽

PR #1358, opened 2023-06-26
coding style

mohamed-barakat:


zickgraf:
I prefer the old style, for example because it matches the usual style used in both GAP and Julia.

mohamed-barakat:
> I prefer the old style, for example because it matches the usual style used in both GAP and Julia.

I don't have a strong opinion, but this is the only coding style that works with the emacs support for GAP :)

mohamed-barakat:
I force-pushed an update which at least complies with the existing coding style

zickgraf (approved):

Issue #1269, opened 2023-03-01
Deprecate and remove AddObjectRepresentation and AddMorphismRepresentation


zickgraf:
This would also allow to get rid of some workarounds in `ObjectifyObjectForCAPWithAttributes`, `CreateCapCategoryObjectWithAttributes`, etc.

zickgraf:
Are any known users of `AddObjectRepresentation` or `AddMorphismRepresentation` left?

mohamed-barakat:
> Are any known users of `AddObjectRepresentation` or `AddMorphismRepresentation` left?

Neither in `CategoricalTowers` nor in `HigherHomologicalAlgebra`.

zickgraf:
@TKuh Is there anything left that you are aware of? If not, I will deprecate this :-)

TKuh:
Should be only the manual left, but we are already working on it in #1372. :+1:

PR #1370, opened 2023-07-03
Use `CreateCapCategoryWithDataTypes` at more places

TKuh:
In particular, get rid of `AddObjectRepresentation` and `AddMorphismRepresentation`.

Step towards #1269.

### DIFF
--- a/FreydCategoriesForCAP/gap/CokernelImageClosure.gi
+++ b/FreydCategoriesForCAP/gap/CokernelImageClosure.gi
@@ -48,11 +48,17 @@ InstallMethod( CokernelImageClosure,
         
     fi;
     
-    cokernel_image_closure := CreateCapCategory( Concatenation( "Cokernel image closure( ", Name( underlying_category ), " )" ) );
+    cokernel_image_closure := CreateCapCategoryWithDataTypes(
+                                      Concatenation( "Cokernel image closure( ", Name( underlying_category ), " )" ),
+                                      IsCokernelImageClosure,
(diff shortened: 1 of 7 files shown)


zickgraf:
Thanks, very nice :-) Can you please create a PR which contains only those commits which do not add new filters or rename filters? E.g. https://github.com/homalg-project/CAP_project/pull/1370/commits/5801c34d7b8fb2a9c1ddeee9fe0d338bbb0fc507. This simplifies the review.

TKuh:
> Thanks, very nice :-) Can you please create a PR which contains only those commits which do not add new filters or rename filters?

Sure, done. :)

General question: it seems to me, that the current naming convention for object/morphism filters is `IsObjectIn*` and `IsMorphismIn*`. Is that right? If yes, are you fine with renaming filters in that way (I did not dare to touch `FreydCategoriesForCAP` though)? It's of course not a big deal, but reads nicer.

zickgraf (approved):
Thanks!

zickgraf:
> General question: it seems to me, that the current naming convention for object/morphism filters is `IsObjectIn*` and `IsMorphismIn*`. Is that right? If yes, are you fine with renaming filters in that way (I did not dare to touch `FreydCategoriesForCAP` though)? It's of course not a big deal, but reads nicer.

We do not really have an official convention (yet). So I suggest to not do a general rename as part of this series of PRs, except where there is an explicit reason (e.g. if you introduce a new category filter I can imagine that sometimes adjusting the object and morphism filters would be natural).

PR #1371, opened 2023-07-04
Use `CreateCapCategoryWithDataTypes` in all examples

TKuh:
And also in `GeneralizedMorphismsForCAP`.

Next step towards #1269.

@zickgraf if you would like me to keep the original names in the first two commits, I'll revert the names.


zickgraf:
Thanks, very nice :-)

Regarding the renames: Since this is only in examples, we do not have to think about backwards compatibility, so this is fine :-)

Please put the commit adjusting the example in the manual into a separate PR: The text in the manual also needs adjustments.

TKuh:
> Please put the commit adjusting the example in the manual into a separate PR: The text in the manual also needs adjustments.

Done. :+1:

zickgraf (approved):
Thanks!

PR #1372, opened 2023-07-05
Remove `Add*Representation` in the CAP manual

TKuh:


zickgraf:
Could you change the code to use `category_as_first_argument := true`? Then you could simply say `category_as_first_argument := true only has to be set for legacy reasons and will not be required in the future` without further explanation.

zickgraf:
An additional idea: We should prominently mention the CAP version in the manual. Then people know why things possibly fail when using a newer version (e.g. master) or an older version (e.g. the one distributed with GAP) of CAP. Maybe you can do this as part of this PR?

TKuh:
Nice ideas, I'll do it. Do you mind if I also append the changes `Objectify*` -> `CreateCapCategory*` for the manual in this PR? Then we would have all changes related to the manual bundled here.

zickgraf:
> Nice ideas, I'll do it. Do you mind if I also append the changes `Objectify*` -> `CreateCapCategory*` for the manual in this PR? Then we would have all changes related to the manual bundled here.

Go ahead :-)

Manual/GAP_code/vecspaces_example/complete_example.g
```diff
@@ -20,14 +23,18 @@ DeclareOperation( "SQVecObj",
...
-BindGlobal( "SQVec", CreateCapCategory( "Skeletal category of rational vector spaces" ) );
+BindGlobal( "SQVec", CreateCapCategoryWithDataTypes(
+                             "Skeletal category of rational vector spaces",
+                             IsSQVecCat,
+                             IsSQVecObj,
+                             IsSQVecMor,
+                             IsCapCategoryTwoCell,
+                             fail,
+                             fail,
+                             fail ) );
```
zickgraf, line 34: I suggest to simply use `CreateCapCategory` to avoid having to talk about the `fail`s.

Manual/Tutorial.tex
```diff
@@ -110,10 +110,12 @@ \subsection{Implementation}
 objects of the same category.
 
 After the declarations of the constructors we create a global category \GAP object, which will be the category $\Svec$.
-Since we are going to create an Abelian category, we tell the system that we are doing this, to have all the computational
-properties of an Abelian category. 
-Next, we tell \CapPkg that every \GAP object which lies in the filter \texttt{IsSQVecObj} is an object of $\Svec$,
+The call of \texttt{CreateCapCategoryWithDataTypes} lets \CapPkg create a category with name
+``\texttt{Skeletal category of rational vector spaces}'' such that every \GAP object which lies in the filter \texttt{IsSQVecObj} is an object of $\Svec$,
 and every \GAP object which lies in the filter \texttt{IsSQVecMor} is a morphism of $\Svec$.
```
zickgraf, line 119: Technically this is wrong, the converse is correct: Every object of Svec lies in `IsSQVecObj`, but an object in `IsSQVecObj` does not necessarily belong to Svec. But this was there before, so we can defer this to another PR, your choice.

Manual/GAP_code/vecspaces_example/complete_example.g
```diff
@@ -35,27 +35,39 @@ BindGlobal( "SQVec", CreateCapCategoryWithDataTypes(
...
+
 BindGlobal( "QQ", HomalgFieldOfRationals( ) );
 
 ##
 InstallMethod( SQVecObj,
-               [ IsInt ],
+               [ IsSQVecCat, IsInt ],
                
-  function( dim )
+  function( cat, dim )
```
zickgraf, line 46: Since this is only a demo, I suggest not making `cat` a variable here and just keeping the hard-coded reference to `SQVec`. Of course this is not how we would suggest organizing the code in more complicated examples, but we would also not recommend creating a global category `SQVec`, so.... :D

zickgraf (commented):
Thanks! Some comments and suggestions below. Please also adjust the example in chapter 2.1 by using `category_as_first_argument := true`.

Manual/GAP_code/vecspaces_example/complete_example.g
TKuh: Done.

Manual/Tutorial.tex
TKuh: I corrected it.

Manual/GAP_code/vecspaces_example/complete_example.g
TKuh: Done. I had the same inner conflict. ^^

Manual/Tutorial.tex
```diff
@@ -110,10 +110,11 @@ \subsection{Implementation}
 objects of the same category.
 
 After the declarations of the constructors we create a global category \GAP object, which will be the category $\Svec$.
+The call of \texttt{CreateCapCategory} lets \CapPkg create a category with name
+``\texttt{Skeletal category of rational vector spaces}'' such that every \GAP object which is an object of $\Svec$, lies in the filter \texttt{IsSQVecObj}
+and every \GAP object which is a morphism of $\Svec$, lies in the filter \texttt{IsSQVecMor}.
```
zickgraf, line 120: Please also mention the role of `IsSQVecCat` here and remove the commas before `lies`.

Manual/Tutorial.tex
```diff
@@ -122,7 +122,7 @@ \subsection{Implementation}
 The first constructor creates the objects of $\Svec$.
 Here, the \CapPkg command
 \begin{itemize}
-  \item \texttt{ObjectifyObjectForCAPWithAttributes}
+  \item \texttt{CreateCapCategoryObjectWithAttributes}
 \end{itemize}
 creates a \GAP object that resides as an object in the category\footnote{
   Note that the usual creation of \GAP types is done automatically by \CapPkg.
```
zickgraf, line 133: I don't think this note applies anymore -> just remove it.

zickgraf (changes requested):
This PR is getting a bit large, please address the comments regarding the first two commits and split those of into a separate PR.

Manual/Tutorial.tex
TKuh: You only meant the footnote, right?

Manual/Tutorial.tex
TKuh: Done.

TKuh:
> This PR is getting a bit large, please address the comments regarding the first two commits and split those of into a separate PR.

This PR now only contains the `Add*Representation` changes. I´ll create another one with the category as first argument (and the CAP version) after this one.

Manual/Tutorial.tex
zickgraf: Yes :-)

Manual/Tutorial.tex
zickgraf: Sorry, I missed another thing which I think should be mentioned here: the role of `IsCapCategoryTwoCell`. Or maybe we simply make the two cell filter optional? The only place I know of where this is used is `CapCat`.

TKuh: Everywhere I changed things to `CreateCapCategory` I also only wrote `IsCapCategoryTwoCell`. So maybe yes? :)

zickgraf: I thought about this and I'm a bit reluctant about adding this convenience: If we add such a convenience for `CreateCapCategory`, we might also want to add it for `CreateCapCategoryWithDataTypes`. But `CreateCapCategoryWithDataTypes` is a function, not an operation, so we would have to do method selection via the number of arguments. This all sounds like it might cause problems if we want to add other kinds of convenience later on. Hence I suggest to keep things as they are and simply mention the role of `IsCapCategoryTwoCell`, i.e. something like "We do not want to consider two cells in the category, so we simply pass the generic filter."

zickgraf (approved):
Thanks!

PR #1375, opened 2023-07-06
Warn about categories marked as being computable but not having an implementation of IsCongruentForMorphisms

zickgraf:


zickgraf:
@sebastianpos If a category has neither `IsEqualForMorphisms` nor `IsCongruentForMorphisms` installed, currently those will fall-back to "if IsIdenticalObj then ReturnTrue, else ReturnFail". While this makes sense for `IsEqualForMorphisms` (in conjunction with strict caching), I think it does not make sense mathematically, i.e. for `IsCongruentForMorphisms`. I suggest marking the category as "not computable" instead in this case. Do you agree?

sebastianpos:
Could you elaborate on the following:
* why is it necessary to have strict caching in order to make sense of the current fall-back for `IsEqualForMorphisms`?
* what makes you think the fall-back for `IsCongruentForMorphisms` does not make sense mathematically?

zickgraf:
> Could you elaborate on the following:
> 
>     * why is it necessary to have strict caching in order to make sense of the current fall-back for `IsEqualForMorphisms`?

Because only with strict caching the specification "equal input gives equal output" can be fulfilled if `IsIdenticalObj` is involved.

>     * what makes you think the fall-back for `IsCongruentForMorphisms` does not make sense mathematically?

Because for example the mathematical specification
```
IsCongruentForMorphisms(
    PreCompose( PreCompose( alpha, beta ), gamma ),
    PreCompose( alpha, PreCompose( beta, gamma ) )
)
```
will usually not be `true` if `IsIdenticalObj` is involved.

Of course in both cases one could say that `fail` does not mean that the equality or congruence does *not* hold.

But all of that aside: I thought the definition of `is_computable := true` would exactly be that we have a decidable congruence of morphisms.

sebastianpos:
> Of course in both cases one could say that `fail` does not mean that the equality or congruence does _not_ hold.

Sorry for being slow on this.

My interpretation of `fail` as an answer of an algorithm is: we do not know what the result is.

zickgraf:
> > Of course in both cases one could say that `fail` does not mean that the equality or congruence does _not_ hold.
> 
> Sorry for being slow on this.
> 
> My interpretation of `fail` as an answer of an algorithm is: we do not know what the result is.

I agree with this point. But the question is how to read the specification "equal inputs give equal outputs". Does "equal outputs" mean
1. `IsEqualFor*( output_1, output_2 ) = true` or
2. `IsEqualFor*( output_1, output_2 ) <> false`?

I read it as in 1., and similarly for the congruence, i.e. that we require
```
IsCongruentForMorphisms(
    PreCompose( PreCompose( alpha, beta ), gamma ),
    PreCompose( alpha, PreCompose( beta, gamma ) )
) = true
```

Probably you read it as in 2.? If yes, I can agree, and will give a different, more specific rationale for this PR (which I probably should have done from the beginning, sorry).

sebastianpos:
I haven't thought about the specification in cases where `IsEqualForMorphisms` can yield `fail`. So maybe your rationale would help here. I always thought of `IsEqualForMorphisms` as a kind of syntactic equality that always should be checkable. But maybe that's too strong?

When it comes to `IsCongruentForMorphisms`, I read it as in 2), i.e., admitting that we cannot compute the answer should not break the specification, since it may nevertheless be provable that we have a sound notion of congruence on our category which is simply not computable.

zickgraf:
Thanks for the clarification! I suggest deferring the discussion about `IsEqualForMorphisms` to #1376 because here `IsEqualForMorphisms` is not touched at all. I should not have mixed up the arguments, sorry for that. I think it's best to forget about all of the above and just look at the following code:
```
gap> LoadPackage( "CAP", false );;
gap> CapCat!.is_computable;
true
gap> CanCompute( CapCat, "IsCongruentForMorphisms" );
true
gap> F := IdentityMorphism( AsCatObject( CapCat ) );;
gap> IsCongruentForMorphisms( PreCompose( PreCompose( F, F ), F ), PreCompose( F, PreCompose( F, F ) ) );
fail
```
As you see, `CapCat` claims to be computable both via the component `is_computable` and via `CanCompute`, but fails to decide the congruence we expect. I think this behaviour is not desirable. What this PR does is:
1. It displays a warning during `Finalize` in this case which encourages the user to install `IsCongruentForMorphisms` (or `IsEqualForMorphisms` because then `IsCongruentForMorphisms` is derived). This warning could also mention `is_computable`, but I thought this is not something which we would like to advertise. Consequently, this PR marks many categories as not computable, e.g. `CapCat`.
2. It removes the fallback final derivation of `IsCongruentForMorphisms` because: Either the category is not computable or the category is computable but the user has not provided `IsCongruentForMorphisms`. In both cases I do not see a reason why anyone should call `IsCongruentForMorphisms` at all.

sebastianpos (approved):
Yes, I agree to your reasoning.

zickgraf:
> Yes, I agree to your reasoning.

Perfect, thanks! (And again sorry for the unnecessarily complicated discussion.)

@mohamed-barakat I will wait with the merge until the CI of CategoricalTowers is green again to not mix too many warnings at once.

PR #1376, opened 2023-07-06
Handle fallbacks for IsEqualForObjects and IsEqualForMorphisms in a better way

zickgraf:
Currently, IsEqualForObjects and IsEqualForMorphisms fall back to
ReturnFail via derivations, with the redirect functions checking
IsIdenticalObj.
This has the following advantages:
1. IsEqualForObjects/Morphisms are always available.
2. As long as things are identical, everything "just works".
3. If things are not identical, one gets a more or less clear error message
   in pre-functions.

It has the following disadvantages:
1. Returning `fail` is not covered by the specification.
2. Every pre-function has to explicitly take `fail` into account.
3. Derivations like IsEndomorphism get triggered although they are not
   computable.
4. The fallback is not compatible with the specification "equal input gives
   equal output" if strict caching is not enabled manually.

This commit implements a better solution:
Instead of relying on fallback derivations for IsEqualForObjects and
IsEqualForMorphisms, we rely on the existing global plausibility checks
for IsEqualForObjects and IsEqualForMorphisms. Hence, no other derivations
are triggered and the specification "equal input gives equal output" simply
has no meaning. Instead of returning `fail` an error message is displayed.
Thus, pre-functions do not have to take `fail` into account. Also, the
error messages are more helpful then the ones in the pre-functions were.

### DIFF
--- a/CAP/gap/CategoryMorphisms.gi
+++ b/CAP/gap/CategoryMorphisms.gi
@@ -393,44 +393,88 @@ InstallMethod( CoefficientsOfMorphismWithGivenBasisOfExternalHom,
 ##
 ######################################
 
-# This method should usually not be selected when the two morphisms belong to the same category
+# This method should usually not be selected when the two morphisms belong to the same category and the category can compute IsEqualForMorphisms.
 InstallOtherMethod( IsEqualForMorphisms,
                     [ IsCapCategory, IsCapCategoryMorphism, IsCapCategoryMorphism ],
 
   function( cat, morphism_1, morphism_2 )
     
     if not HasCapCategory( morphism_1 ) then
+        
         Error( Concatenation( "the morphism \"", String( morphism_1 ), "\" has no CAP category" ) );
+        
     fi;
+    
     if not HasCapCategory( morphism_2 ) then
+        
         Error( Concatenation( "the morphism \"", String( morphism_2 ), "\" has no CAP category" ) );
+        
     fi;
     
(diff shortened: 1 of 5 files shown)


zickgraf:
I think we could even combine this with #621, but let us first handle #1375 to get the mathematical equality correct. So just ignore this PR for now @sebastianpos until we have discussed #1375 :-)

zickgraf:
@sebastianpos This is now an implementation of your suggestion in https://github.com/homalg-project/CAP_project/pull/612#issuecomment-722307457 with the modification that instead of returning `fail` an error is signaled. `fail` would have triggered errors in the pre-functions anyway, so from the user side this is not a big difference, but for us it greatly simplifies things because the pre-functions do not have to take `fail` into account. From my side you do not have to look at the code, but I would like to have your okay for the general idea, see the PR description for details. If you still want to look at the code, you can ignore the changes to the method records, they simply remove the `fail` case.

zickgraf:
Regarding the discussion in https://github.com/homalg-project/CAP_project/pull/1375#issuecomment-1625074541 and https://github.com/homalg-project/CAP_project/pull/1375#issuecomment-1625496878, which I suggested to move here: This does not apply anymore because `fail` does not occur anymore :-)

sebastianpos:
I'm okay with the general idea :-)

It's better to have more meaningful error messages and get rid of `fail` as an unspecified output.

zickgraf:
Perfect, thanks! I'm very happy that we can finally remove the weird `ReturnFail` derivations! :-)

PR #1383, opened 2023-07-16
Adjust documentation of `MakeDerivation`

TKuh:
~~And a small simplification of a check in `MakeDerivation`.~~

### DIFF
--- a/CAP/gap/Derivations.gd
+++ b/CAP/gap/Derivations.gd
@@ -41,15 +41,18 @@ DeclareCategory( "IsDerivedMethod", IsAttributeStoringRep );
 #!  purposes.
 #!  The argument <A>target_op</A> is the operation which
 #!  the derived method implements.
-#!  The argument <A>used_ops_with_multiples</A> contains the name of each
-#!  operation used by the derived method, together with a positive
-#!  integer specifying how many times that operation is used and
-#!  optionally a category getter.
+#!  The argument <A>used_ops_with_multiples</A> contains
+#!  
+#!  * the name of each operation used by the derived method,
+#!  * together with a positive integer specifying how many times that operation is used and
+#!  * either a category getter or <C>fail</C>.
+#!  
 #!  This is given as a list of lists, where each sublist has as
 #!  first entry the name of an operation, as second entry an integer and as
-#!  third entry optionally a function. This function should accept the
-#!  category and return a category for which the operation in the first
-#!  entry must be installed for the derivation to be considered valid.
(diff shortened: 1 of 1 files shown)


CAP/gap/Derivations.gi
```diff
@@ -43,7 +43,7 @@ function( name, target_op, used_op_names_with_multiples_and_category_getters, we
     fi;
     # =#
     
-    if NumberArgumentsFunction( category_filter ) = 0 or NumberArgumentsFunction( category_filter ) > 1 then
```
zickgraf, line 46: This is deliberate: Functions with a variable number of arguments have a negative `NumberArgumentsFunction`, e.g.
```
gap> NumberArgumentsFunction( {x...} -> 1 );
-1
```
and those should work here too, although I would not encourage this, which is why it's not mentioned in the error message.

TKuh: Ah, thank you! Sorry for the inconvenience.

CAP/gap/Derivations.gd
```diff
@@ -41,15 +41,18 @@ DeclareCategory( "IsDerivedMethod", IsAttributeStoringRep );
...
+#!  
 #!  This is given as a list of lists, where each sublist has as
 #!  first entry the name of an operation, as second entry an integer and as
-#!  third entry optionally a function. This function should accept the
-#!  category and return a category for which the operation in the first
-#!  entry must be installed for the derivation to be considered valid.
+#!  third entry either a function or <C>fail</C>. This function should accept the
+#!  category for which this derivation will be installed,
+#!  and return a category for which the operation in the first
+#!  entry is installed.
```
zickgraf, line 55:
```suggestion
#!  entry must be installed for the derivation to be considered applicable.
```
It's okay if the operation in the first entry is not installed, then the derivation simply will not be applied.

CAP/gap/Derivations.gd
```diff
@@ -58,7 +61,7 @@ DeclareCategory( "IsDerivedMethod", IsAttributeStoringRep );
 #!  The argument <A>func</A> contains the actual implementation of the
 #!  derived method.
 #!  The argument <A>category_filter</A> is a filter (or function) describing
-#!  which categories the derivation is valid for.  If it is valid
+#!  for which categories the derivation is valid for. If it is valid
```
zickgraf, line 64: This sentence now contains "for" two times.

zickgraf (approved):
Thanks!

PR #1394, opened 2023-07-24
added missing category as first argument in some derivations

mohamed-barakat:


mohamed-barakat:
For the records: I ran into this while trying to compile `LeftCartesianDistributivityExpandingWithGivenObjects`.

Issue #973, opened 2022-08-10
Update all existing categories in CAP_project to use `category_as_first_argument := true`


zickgraf:
Set `category!.category_as_first_argument := true` right after creating the category and make sure all added functions (e.g. via `AddPreCompose`) accept the category as the first argument.

zickgraf:
The categories which should be updated can now easily be found by grepping for `category_as_first_argument := false`.

TKuh:
In `AttributeCategoryForCAP`, generic operations are added depending on the return type of the operation.
See for example this part:

https://github.com/homalg-project/CAP_project/blob/f369742f174d47079589f1b76e45bc222951aace/AttributeCategoryForCAP/gap/AttributeCategory.gi#L447-L454

which eventually adds this generic function as the operation:

https://github.com/homalg-project/CAP_project/blob/f369742f174d47079589f1b76e45bc222951aace/AttributeCategoryForCAP/gap/AttributeCategory.gi#L209-L223

Now, when adding the category as first argument, it is not enough to just change
`function( arg )` into `function( cat, arg )`. If `arg` contains more than one element, then the `Add*` mechanism still thinks, it has too few elements. Like here (with [these](https://github.com/TKuh/CAP_project/commit/a205d7407c0da02e00e51158b5d0416319ce4464) changes):

```gap
gap examples/example.g 

Error, In AddAdditionForMorphisms: given function 1 has 2 arguments but should have 3 at /home/tom/.gap/pkg/CAP_project/CAP/gap/InstallAdds.gi:319 called from
add_function( category, [ [ func, [  ] ] ], weight 
 ); at /home/tom/.gap/pkg/CAP_project/CAP/gap/InstallAdds.gi:175 called from
add_function( category, func, -1 
 ); at /home/tom/.gap/pkg/CAP_project/CAP/gap/InstallAdds.gi:166 called from
add_function( category_with_attributes, function_to_add 
 ); at /home/tom/.gap/pkg/CAP_project/AttributeCategoryForCAP/gap/Attr\
ibuteCategory.gi:454 called from
CAP_INTERNAL_INSTALL_ADDS_FOR_CATEGORY_WITH_ATTRIBUTES( 
 structure_record 
 ); at /home/tom/.gap/pkg/CAP_project/AttributeCategoryForCAP/gap/Attr\
ibuteCategory.gi:759 called from
<function "EnhancementWithAttributes">( <arguments> )
 called from read-eval loop at examples/example.g:132
[...]

PR #1399, opened 2023-07-28
Category as first argument for `FieldAsCategory` and `Schemes`

TKuh:
Part of #973.


zickgraf:
Could you create a (separate) PR which removes all occurrences of `AddIsEqualForCacheFor*` installing `IsIdenticalObj`? Then we don't have to always deal with this in this sequence of PRs.

TKuh:
You were faster commenting than I was with rebasing. :D

PR #1409, opened 2023-08-02
Introduce "Target" as an alias of "Range"

zickgraf:
Do not declare it as a synonym to avoid conflicts with other GAP packages.

"Target" seems to be used more widely and has typographical advantages:
1. "Target" naturally comes after "Source" when sorting alphabetically.
2. "Source" and "Target" have the same number of letters.
3. It better matches the maps "s" and "t" which are sometimes used to define categories.

@mohamed-barakat I have already talked to @sebastianpos and @kamalsaleh about this and both are in favor of this change. Do you also approve of this idea?

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2023.08-03",
+Version := "2023.08-04",
 Date := (function ( ) if IsBound( GAPInfo.SystemEnvironment.GAP_PKG_RELEASE_DATE ) then return GAPInfo.SystemEnvironment.GAP_PKG_RELEASE_DATE; else return Concatenation( ~.Version{[ 1 .. 4 ]}, "-", ~.Version{[ 6, 7 ]}, "-01" ); fi; end)( ),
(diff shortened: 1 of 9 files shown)


mohamed-barakat:
> @mohamed-barakat I have already talked to @sebastianpos and @kamalsaleh about this and both are in favor of this change. Do you also approve of this idea?

I find this idea great and would have suggested it anyway :) Thank you very much.

mohamed-barakat:
I would even go so far and make Range an alias of Target, if this would technically be possible.

zickgraf:
> I find this idea great and would have suggested it anyway :) Thank you very much.

Perfect, then I will discuss this once more verbally with @sebastianpos and then merge this PR.

> I would even go so far and make `Range` an alias of `Target`, if this would technically be possible.

This can certainly be a next step, but for now I want to keep things 100% compatible (and first see if this causes problems with other GAP packages). If we want to make `Target` the new default, we should also discuss and change some other things like `...WithGivenRange` and `RangeCategoryOfHomomorphismStructure`.

mohamed-barakat:
Agree on all points.

zickgraf:
For the record: `Range` was introduced in 13fe1b03ed950a059b1f2a93935bf136f97678a6, the fourth commit of CAP :D So it has really been there since the beginning.

PR #1413, opened 2023-08-09
Introduce IsIsomorphicForObjects and SomeIsomorphismBetweenObjects

zickgraf:
@sebastianpos Is there a reason why `IsIsomorphicForObjects` was not added before? I'm asking because it seems like a very obvious operation :D But maybe it simply was never needed? In fact, I also have no concrete application, I just want to make `IsSkeletalCategory` precise, which can now be defined as `IsEqualForObjects` and `IsIsomorphicForObjects` coinciding.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2023.08-05",
+Version := "2023.08-06",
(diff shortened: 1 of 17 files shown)


zickgraf:
> Together with this operation, it would be beneficial to introduce an operation that enables computing an isomorphism between two isomorphic objects.

Do you have a suggestion for a name? `SomeIsomorphismBetweenObjects`?

kamalsaleh:
Nope, I like the name you suggested.

sebastianpos:
No, there is no particular reason. You can introduce it.

PR #1431, opened 2023-08-17
d&i TransposedListList and use it in MorphismBetweenDirectSums

kamalsaleh:
to ensure that the translated inputs comming from opposite categories still respect the specifications of `MorphismBetweenDirectSums` and bump CAP to V2023.08.10.


zickgraf (approved):
Very nice, thank you!

PR #1433, opened 2023-08-17
added sanity checks to AsMorphismInWrapperCategory

mohamed-barakat:
This code would have helped me discover sources of errors much earlier.

### DIFF
--- a/CAP/gap/WrapperCategory.gi
+++ b/CAP/gap/WrapperCategory.gi
@@ -55,6 +55,14 @@ InstallOtherMethodForCompilerForCAP( AsMorphismInWrapperCategory,
     #% CAP_JIT_DROP_NEXT_STATEMENT
     CAP_INTERNAL_ASSERT_IS_MORPHISM_OF_CATEGORY( morphism, ModelingCategory( D ), [ "the morphism given to AsMorphismInWrapperCategory" ] );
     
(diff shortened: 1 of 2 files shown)


CAP/gap/WrapperCategory.gi
```diff
@@ -55,6 +55,13 @@ InstallOtherMethodForCompilerForCAP( AsMorphismInWrapperCategory,
     #% CAP_JIT_DROP_NEXT_STATEMENT
     CAP_INTERNAL_ASSERT_IS_MORPHISM_OF_CATEGORY( morphism, ModelingCategory( D ), [ "the morphism given to AsMorphismInWrapperCategory" ] );
     
+    #% CAP_JIT_DROP_NEXT_STATEMENT
+    if not IsEqualForObjects( ModelingCategory( D ), ObjectDatum( D, source ), Source( morphism ) ) then
+        Error( "ObjectDatum( source ) and Source( morphism ) do not coincide\n" );
+    elif not IsEqualForObjects( ModelingCategory( D ), ObjectDatum( D, range ), Range( morphism ) ) then
```
zickgraf, line 61: These checks are semantically independent, so I suggest two `if ... fi;` instead of using `elif`.

mohamed-barakat: Done.

PR #1434, opened 2023-08-17
Category as first argument for `AttributeCategoryForCAP` and `ActionsForCAP`

TKuh:
Part of #973.

### DIFF
--- a/AttributeCategoryForCAP/examples/example.g
+++ b/AttributeCategoryForCAP/examples/example.g
@@ -54,7 +54,7 @@ underlying_category := MatrixCategory( Q );;
 ## We create a "void" CapCategory here and set its properties (abelian, monoidal) in the beginning
 category_of_objects_with_endomorphism := CreateCapCategory( "Category of objects with endomorphisms" );;
 
-category_of_objects_with_endomorphism!.category_as_first_argument := false;;
+category_of_objects_with_endomorphism!.category_as_first_argument := true;;
 
 SetIsAbelianCategory( category_of_objects_with_endomorphism, true );;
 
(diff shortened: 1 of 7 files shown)


AttributeCategoryForCAP/gap/AttributeCategory.gi
```diff
@@ -233,9 +233,9 @@ InstallGlobalFunction( CAP_INTERNAL_INSTALL_ADDS_FOR_CATEGORY_WITH_ATTRIBUTES,
         
         output_source_getter := CAP_INTERNAL_METHOD_NAME_RECORD.(operation_name).output_source_getter;
         output_range_getter := CAP_INTERNAL_METHOD_NAME_RECORD.(operation_name).output_range_getter;
-        
+       
         return
-          function( arg )
+        function( cat, arg... )
```
zickgraf, line 238: I think the old whitespace was correct.

AttributeCategoryForCAP/examples/example.g
```diff
@@ -82,44 +82,44 @@ structure_record.morphism_type := TheTypeOfMorphismsOfObjectsWithEndomorphism;;
 
 ## This is the mathematical core: provide the functions that enhance objects with attributes
 structure_record.ZeroObject :=
-  function( underlying_zero_object )
+  function( cat, underlying_zero_object )
```
zickgraf, line 85: This is a backwards incompatible change, right? Does passing the category have a concrete advantage?

I agree that passing the category would be the "modern way" of doing things, but I would suggest not breaking the backwards compatibility without a concrete advantage. (Additionally, it looks like `cat` is the AttributeCategory but maybe the underlying category would make more sense.)

zickgraf (changes requested):

AttributeCategoryForCAP/gap/AttributeCategory.gi
TKuh: Sorry, I missed that.

AttributeCategoryForCAP/examples/example.g
zickgraf: My comment was meant to apply to all entries of `structure_record`. What is special about `ZeroObject` and `DualOnObjects` that you only reverted those two?

AttributeCategoryForCAP/examples/example.g
```diff
@@ -119,7 +119,7 @@ structure_record.DualOnObjects :=
 ## In this particular example, this function would be not necessary since it can be 
 ## derived from the rigidity of the category
 structure_record.InternalHomOnObjects :=
-  function( object1, object2, underlying_internal_hom )
+  function( cat, object1, object2, underlying_internal_hom )
```
zickgraf, line 122: There is still a `cat` here, is this deliberate?

TKuh: Argh... no, it was probably too close to the `AddIsEqual...` below, so I overlooked it.
Thank you for spotting it.

zickgraf (approved):
Thanks!

PR #1441, opened 2023-08-21
Replace the 'vecspaces' categories with the `MatrixCategory`

TKuh:
And move some tests to other packages.

I'll delete the remaining `vecspaces` files in a follow-up. This will in particular remove the last instances of #973.


zickgraf:
Thanks!

Issue #1411, opened 2023-08-06
Coequalizer( common_range, list ) does not work


mohamed-barakat:
```gap
gap> LoadPackage( "CAP" );;
gap> T := TerminalCategoryWithSingleObject();;
gap> t := TerminalObject( T );;
gap> id := IdentityMorphism( t );;
gap> Equalizer( t, [ id, id ] );;
gap> Coequalizer( t, [ id, id ] );;
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `Range' on 1 arguments at /Users/mo/software/gap/build/gap-4.12.1/lib/methsel2.g:249 called from
Range( list_of_morphisms[1] ) at ~/.gap/pkg/CAP_project/CAP//gap/UniversalObjects.gi:634 called from
CoequalizerOp( CapCategory( arg[1] ), arg ) at ~/.gap/pkg/CAP_project/CAP//gap/UniversalObjects.gi:603 called from
<function "Coequalizer">( <arguments> )
 called from read-eval loop at *stdin*:6
type 'quit;' to quit to outer loop
brk> list_of_morphisms;
[ <A zero object in TerminalCategoryWithSingleObject( )>,
  [ <A zero, identity morphism in TerminalCategoryWithSingleObject( )>,
    <A zero, identity morphism in TerminalCategoryWithSingleObject( )> ] ]
```

mohamed-barakat:
As one can see, `list_of_morphisms` is not the expected list of morphisms.

zickgraf:
This probably is just a typo, have you looked at the implementation of "Coequalizer"?

PR #1448, opened 2023-08-25
fixed index typo in Coequalizer

mohamed-barakat:
fixes https://github.com/homalg-project/CAP_project/issues/1411


mohamed-barakat:
Code coverage was uploaded but not reported here. Is it ok to merge?

PR #1457, opened 2023-09-12
do not extend the range of the hom-structure by identity twice

mohamed-barakat:


mohamed-barakat:
I would like to discuss the potential risks of such a change.

zickgraf:
I do not see a risk, but it only prevents your error in a very specific situation, i.e.
```
gap> WrapperCategory( SkeletalFinSets, rec( ) );
gap> WrapperCategory( SkeletalFinSets, rec( ) );
```
should still throw an error with the PR if I am not mistaken.

If you still want to merge this, please rename the component to a more descriptive name.

mohamed-barakat:
> I do not see a risk, but it only prevents your error in a very specific situation, i.e.
> 
> ```
> gap> WrapperCategory( SkeletalFinSets, rec( ) );
> gap> WrapperCategory( SkeletalFinSets, rec( ) );
> ```
> 
> should still throw an error with the PR if I am not mistaken.

Yes, it does. For it to work I would have to take care of the `if`-case in `ReinterpretationOfCategory.gi`, which I currently do not need. I wanted the change to be minimal.

> If you still want to merge this, please rename the component to a more descriptive name.

Done.

PR #1465, opened 2023-09-20
Fix derivation of IsDominating

zickgraf:
The dual version

https://github.com/homalg-project/CAP_project/blob/27ad4ab68123184d4a026803b69dc0fe0024bf1d/CAP/gap/DerivedMethods.gi#L1661-L1675

seems to be correct, that's how I discovered this problem. But I may be wrong. @sebastianpos Could you double-check this? It's very easy to make mistakes with these things :D

Issue #1460, opened 2023-09-18
The role of `CanonicalIdentificationFromCoimageToImageObject`


zickgraf:
`CanonicalIdentificationFromCoimageToImageObject` was added relatively recently in 6c409ef825ab8b9f9a6e72b920f761ed5840174b (2019), but grepping shows no code using it. @sebastianpos Do you recall for which application you added it?

The role of `CanonicalIdentificationFromCoimageToImageObject` and its dual are unclear to me. Is it the same as `MorphismFromCoimageToImage`, just in an Abelian category? Or in any category where images and coimages are isomorphic?

Also, I missed to rename `CanonicalIdentificationFromCoimageToImageObject` to `CanonicalIdentificationFromCoimage*Object*ToImageObject` in 81d776732406aada5f5eef845d659538abe8cd4a. But then again, `MorphismFromCoimageToImage` does not contain `Object` at all. I guess the names should be consistent.

sebastianpos:
I cannot recall. I'm sorry.

It probably was meant for the abelian context. Anyway, you may get rid of it if you want to.

zickgraf:
@sebastianpos Thanks for the fast reply! I'm planning to automatically dualize derivations as an application of CompilerForCAP, and since the derivations around `CanonicalIdentificationFromCoimageToImageObject` are not symmetric with regard to dualization, I indeed would like to get rid of it.

While trying to remove it, I noticed that there actually is a derivation for it, in Toposes: https://github.com/homalg-project/Toposes/commit/3f9d8f0fac6de69ef130ed920d5880551e127498. This derivation was added right after the introduction in CAP, so that seems to be the application. @mohamed-barakat Do you need this derivation? If yes, could you also use `MorphismFromCoimageToImage`?

mohamed-barakat:
@mohamed-barakat Do you need this derivation? If yes, could you also use `MorphismFromCoimageToImage`?

I don't know. You can delete it an we can see :)

zickgraf:
Resolved by https://github.com/homalg-project/CAP_project/pull/1470.

PR #1468, opened 2023-09-25
Generate dual derivations automatically

zickgraf:
We can generate about 1/3 of all derivations in CAP automatically by dualizing existing derivations.

This might contain some regressions due to https://github.com/homalg-project/CAP_project/issues/1460 and https://github.com/homalg-project/CAP_project/pull/1465 which will automatically fixed once the issue/PR are addressed.


zickgraf:
This PR now also speeds up precompilation of categories in tests by using `overhead := false`. See https://github.com/homalg-project/CAP_project/issues/1469 for why this is needed.

mohamed-barakat:
Congratulations 🎉

PR #1478, opened 2023-10-03
Add `PackageOfCAPOperation` to `HelpForCAP`

TKuh:
I think this is helpful for beginners.

### DIFF
--- a/CAP/help_for_CAP.md
+++ b/CAP/help_for_CAP.md
@@ -1,7 +1,7 @@
 ## CAP info functions
 
 * CanCompute( category, string )
-  Returns true if the the function with name <string> is computable in <category>.
+  Returns true if the operation with name <string> is computable in <category>.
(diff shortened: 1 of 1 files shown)


zickgraf:
I'm not opposed to this, but I'm curious: In which context do you use or expect to use this function?

TKuh:
By now, it's not for me. I just remember when I had no overview of all the various operations provided by `CAP`, the packages in `CategoricalTowers`, etc. I often spend quite some time to find operations, especially when I made mistakes I couldn't figure out. Maybe it would have saved me some time and gave me more confidence, because I would have been less "lost". :) 

I just expect this to be a little helpful for future beginners/students who try to learn.

zickgraf:
I see. I think similar information should be available via the help system, but having multiple ways of getting information might make things easier :-)

Issue #1483, opened 2023-10-04
Fix the order of arguments in `InverseMorphismFromCoimageToImageWithGivenObjects`


mohamed-barakat:
I am also unhappy with the order of arguments in `InverseMorphismFromCoimageToImageWithGivenObjects`. It does not align with the ordering convention used by other `*WithGivenObjects` operations.

_Originally posted by @mohamed-barakat in https://github.com/homalg-project/CAP_project/issues/1475#issuecomment-1744853771_

PR #1482, opened 2023-10-04
Fix documentation of MorphismFromCoimageToImageWithGivenObjects

mohamed-barakat:
closes https://github.com/homalg-project/CAP_project/issues/1475

PR #1484, opened 2023-10-04
renamed InverseMorphismFromCoimageToImage* -> InverseOfMorphismFromCoimageToImage*

mohamed-barakat:
and switched the order of arguments

resolves https://github.com/homalg-project/CAP_project/issues/1483

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2023.10-02",
+Version := "2023.10-03",
 Date := (function ( ) if IsBound( GAPInfo.SystemEnvironment.GAP_PKG_RELEASE_DATE ) then return GAPInfo.SystemEnvironment.GAP_PKG_RELEASE_DATE; else return Concatenation( ~.Version{[ 1 .. 4 ]}, "-", ~.Version{[ 6, 7 ]}, "-01" ); fi; end)( ),
 License := "GPL-2.0-or-later",
 
--- a/CAP/gap/AddFunctions.autogen.gd
+++ b/CAP/gap/AddFunctions.autogen.gd
@@ -1627,20 +1627,20 @@ DeclareOperation( "AddInverseForMorphisms",
 #! @Description
 #! The arguments are a category $C$ and a function $F$.
 #! This operation adds the given function $F$
-#! to the category for the basic operation `InverseMorphismFromCoimageToImageWithGivenObjects`.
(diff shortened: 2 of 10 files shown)


mohamed-barakat:
Should be rebased onto https://github.com/homalg-project/CAP_project/pull/1482. I just wanted to free my stack.

mohamed-barakat:
`FreydCategoriesForCAP` is failing since the mechanism that counts the number of CAP operations per derivation does not take synonyms into account. In this case, the derivation is in `Toposes` and uses the synonym (=old name) `InverseMorphismFromCoimageToImageWithGivenObjects`.

zickgraf:
> `FreydCategoriesForCAP` is failing since the mechanism that counts the number of CAP operations per derivation does not take synonyms into account. In this case, the derivation is in `Toposes` and uses the synonym (=old name) `InverseMorphismFromCoimageToImageWithGivenObjects`.

`Toposes` is not loaded in the CAP_project CI, so this cannot be the culprit. Also, the synonym is wrong because the arguments are swapped. In this case, I suggest to not provide backwards compatibility to explicitly draw attention to the fact that the arguments are now swapped.

mohamed-barakat:
> `Toposes` is not loaded in the CAP_project CI, so this cannot be the culprit.

I falsely assumed that the CAP_project tests would work on my laptop as they do in the GitHub CI.

```gap
Executing with non-reversed RecNames
Architecture: aarch64-apple-darwin21-default64-kv8

testing: /Users/mo/.gap/pkg/CAP_project/FreydCategoriesForCAP//tst/050_load_with_suggested_dependencies.tst
########> Diff in /Users/mo/.gap/pkg/CAP_project/FreydCategoriesForCAP//tst/050_load_with_suggested_dependencies.tst:8
# Input is:
LoadPackage( "FreydCategoriesForCAP", false );
# Expected output:
true
# But found:
Error, Variable: 'InverseMorphismFromCoimageToImageWithGivenObjects' must have\
 a value
Syntax warning: Unbound global variable in /Users/mo/.gap/pkg/CategoricalTower\
s/Toposes//gap/ToposDerivedMethods.gi:1048
    iso := InverseMorphismFromCoimageToImageWithGivenObjects( cat, coimage, mo\
r, image );
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error, was not in any namespace
Error, was not in any namespace
########
########> Diff in /Users/mo/.gap/pkg/CAP_project/FreydCategoriesForCAP//tst/050_load_with_suggested_dependencies.tst:11
# Input is:
LoadPackage( "FreydCategoriesForCAP" );
# Expected output:
true
# But found:
#I  Toposes: Do not call `LoadPackage( "freydcategoriesforcap", ... )' in the package file
```

> Also, the synonym is wrong because the arguments are swapped. In this case, I suggest to not provide backwards compatibility to explicitly draw attention to the fact that the arguments are now swapped.

Done.

PR #1492, opened 2023-10-12
Add derivation for `*AdjunctionMapWithGivenTensorProduct`

TKuh:
and regenerate `CartesianCategories`.


MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -91,6 +91,37 @@ AddDerivationToCAP( InternalHomToTensorProductAdjunctionMap,
...
+    #      |
+    #      | g ⊗ id_b
+    #      v
+    # Hom(b,c) ⊗ b
+    #      |
+    #      | ev_bc
+    #      v
+    #      c
+     
+     ev_bc := EvaluationMorphism( cat, b, c );
```
mohamed-barakat, line 117: there is an obsolete space in front of `ev_bc` I think.

Would it make sense to compute `Hom(b,c)` first and pass it both to `EvaluationMorphismWithGivenSource` and to `TensorProductOnMorphismsWithGivenTensorProducts`?

CartesianCategories/gap/SymmetricCartesianClosedCategoriesDerivedMethods.gi
```diff
@@ -94,6 +94,37 @@ AddDerivationToCAP( ExponentialToDirectProductAdjunctionMap,
...
+    #      |
+    #      | g × id_b
+    #      v
+    # Exp(b,c) × b
+    #      |
+    #      | ev_bc
+    #      v
+    #      c
+     
+     ev_bc := CartesianEvaluationMorphism( cat, b, c );
```
mohamed-barakat, line 120: Is there an extra space character here?

TKuh: Fixed.

MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
TKuh: I'm not sure. In case the compiler can optimize `EvaluationMorphism` away, it does not make a difference (see the compilation of MatrixCategory in https://github.com/TKuh/CAP_project/commit/d1daac3068bc94465029a11bef20eb5fbd26e880).  If `EvaluationMorphism` has to be computed, then `Hom(b,c) ⊗ b` will be computed either way as the source and then passed on? But I agree that your suggestion would be a more minimal approach. I tried to stay consistent with the other derivation https://github.com/homalg-project/CAP_project/blob/add7ebdb3f535e7a5628770663bd0d766ab2fc43/MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi#L58-L64

mohamed-barakat: @zickgraf, what do you think?

mohamed-barakat:
Could you also please rebase on the current CAP?

MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
mohamed-barakat: > I'm not sure. In case the compiler can optimize `EvaluationMorphism` away, it does not make a difference

Is it possible to check on the compiled code in the PR if it would make a difference?

TKuh: I'm not sure if I understand you correctly, but you can see the compiled change between your idea and my implementation in this branch: https://github.com/TKuh/CAP_project/tree/adjunction_suggestion. The code in `MatrixCategory` does only change in the weight at the end.

Or do you mean something else by "make a difference"?

mohamed-barakat: Yes, exactly, thanks for the branch. I see, there is only a change of weights. However, I don't understand why, but this is good enough for me.

zickgraf: If a category does not have both the WithGiven and the non-WithGiven variant installed, then all variants are more or less equivalent, especially when using CompilerForCAP which will eliminate redudancies and drop dead code.

In any case, consistency should be the top priority, so either one should change both places or none.

PR #1493, opened 2023-10-16
Fix passing a type signature with a category as first argument to CapJitCompiledFunction

zickgraf:


mohamed-barakat:
I am curious, where did you encounter this bug?

zickgraf:
> I am curious, where did you encounter this bug?

I discovered it when manually compiling a piece of code where I wanted homalg matrices as inputs. There, I could not use the usual strings but had to actually pass a type signature to `CapJitCompiledFunction`. This does not happen inside CompilerForCAP itself, that's why this did not cause problems earlier.

Issue #1239, opened 2023-02-02
Verify that all derivations have correct source and range


zickgraf:
using the code in https://github.com/zickgraf/CAP_project/commit/e0b9edfa7ea48092ab91184158ca98040fd08ecf.

Currently, there are many false positives because of missing `output_source/range_getter_strings` (which means CompilerForCAP cannot determine the source/range of an operation) and because of missing WithGiven versions of operations (which means the given source or range cannot be passed through to the result).

TKuh:
> using the code in [zickgraf@e0b9edf](https://github.com/zickgraf/CAP_project/commit/e0b9edfa7ea48092ab91184158ca98040fd08ecf).

I'm currently working through the list your code generates and several derivations are from `MonoidalCategories`. Typically, both the closed and coclosed cases of a derivation are affected. In theory, I should now be able to use your code from #1468 to generate the coclosed derivations, right? :)
I just want to be sure I can (for now) safely skip the coclosed cases.

zickgraf:
> I'm currently working through the list your code generates and several derivations are from `MonoidalCategories`.

Nice :-)

> Typically, both the closed and coclosed cases of a derivation are affected. In theory, I should now be able to use your code from #1468 to generate the coclosed derivations, right? :)

In theory, yes. In practice I'm not working on this right now and there are still some engineering problems to be solved, so there is no timeline for this from my side.

> I just want to be sure I can (for now) safely skip the coclosed cases.

"Skip" in the sense of "not look at" or in the sense of "not fix"? If you find an error in the closed case, it would be very nice if you could also fix the coclosed case. 

When applying #1468 for the first time, one should compare the autogenerated derivations to the existing ones to catch errors like https://github.com/homalg-project/CAP_project/pull/1465. So any error/inconsistency fixed beforehand makes this comparison easier.

zickgraf:
> So I hoped, if I add these `WithGiven`'s in the closed case, I could for now skip adding them in the coclosed case and later just generate the coclosed case, and get the `WithGiven`'s there too.

Ah, now I understand. I think modifying only the closed case but not the coclosed case is not a good idea: This introduces unexpected asymmetries which in turn can lead to weird issues like https://github.com/homalg-project/CAP_project/issues/917.

> I'm fine with doing both cases by hand if you advise this; it's just quite time consuming and error prone, so I hoped to be more efficient. :)

I see the following options:

1. Ignore these cases manually for now.
[...]

PR #1494, opened 2023-10-17
WithGivens for monoidal and (co)cartesian derivations

TKuh:
Depends on #1492 

Part of #1239


mohamed-barakat (changes requested):
Could you please
1. rebase;
2. add a first commit to bump the version and the dependency of the three packages `CAP`, `FreydCategoriesForCAP` and `LinearAlgebraForCAP`;
3. add a last commit bumping the version of `MonoidalCategories` and `CartesianCategories`.

mohamed-barakat:
> 2. add a first commit to bump the version and the dependency of the three packages `CAP`, `FreydCategoriesForCAP` and `LinearAlgebraForCAP`;

In the first commit please also bump the dependency on the released `MonoidalCategories` and `CartesianCategories` where possible.

MonoidalCategories/gap/BraidedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -6,24 +6,22 @@
...
-                    [ [ InverseForMorphisms, 1 ],
-                      [ Braiding, 1 ] ],
+                    "BraidingInverseWithGivenTensorProducts via BraidingWithGivenTensorProducts with reordered arguments",
+                    [ [ BraidingWithGivenTensorProducts, 1 ] ],
                     
   function( cat, object_2_tensored_object_1, object_1, object_2, object_1_tensored_object_2 )
-    ##TODO: Use BraidingWithGiven
-    return InverseForMorphisms( cat, Braiding( cat, object_1, object_2 ) );
+    
+    return BraidingWithGivenTensorProducts( cat, object_2_tensored_object_1, object_2, object_1, object_1_tensored_object_2 );
```
zickgraf, line 14: This only holds in symmetric categories, and already exists in `SymmetricMonoidalCategoriesDerivedMethods.gi`.

CartesianCategories/gap/SymmetricCartesianClosedCategoriesDerivedMethods.gi
```diff
@@ -879,7 +879,7 @@ AddDerivationToCAP( CartesianPostComposeMorphismWithGivenObjects,
     
     braiding := CartesianBraiding( cat, ExponentialOnObjects( cat, b, c ), ExponentialOnObjects( cat, a, b ) );
     
-    return PreCompose( cat, braiding, CartesianPreComposeMorphism( cat, a, b, c ) );
+    return PreComposeList( cat, source, [ braiding, CartesianPreComposeMorphism( cat, a, b, c ) ], range );
```
zickgraf, line 885: I don't think this is the desired solution. The source of `PreCompose` is the source of the first morphism, i.e. the source of `braiding`. So the desired solution is that `braiding` has the expected source in the first place.

Thinking more about this, I think we should never use `PreComposeList` for literal lists. The information about source and range is already available in this case, so adding source and range explicitly does not add anything if everything else is correct. And if things are wrong, adding source and range only hides actual errors. For example, if `braiding` would have the wrong source here, using PreComposeList would just hide the wrong source from CompilerForCAP.

mohamed-barakat:
@TKuh: Could you please make a separate PR out of the first commit? This should fix https://github.com/gap-system/PackageDistro/pull/846.

PR #1495, opened 2023-10-17
derived ImageEmbedding as the colift along the coastriction to image

mohamed-barakat:


zickgraf:
Please also add a WithGiven version of the derivation to make it consistent with the analogous derivations of `CoastrictionToImage(WithGiven...)`.

zickgraf:
I just noticed that CAP's definition of Image does not require the coastriction to the image to be epi. Hence, in general this derivation is not valid, or am I missing something?

mohamed-barakat:
> I just noticed that CAP's definition of Image does not require the coastriction to the image to be epi. Hence, in general this derivation is not valid, or am I missing something?

You are right, I have to use the general lift an colift.

mohamed-barakat:
> You are right, I have to use the general lift an colift.

Such a factorization is always [unique up to unique isomorphism](https://ncatlab.org/nlab/show/image#definition).

mohamed-barakat:
It is now unclear to me why the colift is necessarily a monomorphism.

mohamed-barakat:
I restricted the derivation to `IsAbelianCategory`. I remember that the natural morphism from the coimage to image is both mono and epi, making these derivations also valid for `IsPreAbelianCategory`. I can relax the conditions in a later PR once I have a proof.

PR #1496, opened 2023-10-19
Bump versions and dependencies

TKuh:
See https://github.com/homalg-project/CAP_project/pull/1494#issuecomment-1771735230


mohamed-barakat (approved):
Thanks

PR #1499, opened 2023-11-05
relocated code for better readability

mohamed-barakat:


CartesianCategories/gap/CartesianClosedCategoriesMethodRecord.gi
```diff
@@ -11,123 +11,123 @@ InstallValue( CARTESIAN_CLOSED_CATEGORIES_METHOD_NAME_RECORD, rec(
...
   return_type := "object",
+  input_arguments_names := [ "cat", "a", "b" ],
   dual_operation := "CoexponentialOnObjects",
   dual_arguments_reversed := true,
   # Test in CartesianClosedCategoriesTest
 ),
 
 ExponentialOnMorphisms := rec(
   filter_list := [ "category", "morphism", "morphism" ],
+  return_type := "morphism",
```
zickgraf, line 23: There is a kind of standard given by
https://github.com/homalg-project/CAP_project/blob/31d981c477085bb5d193eba1521e05054fe23875/CAP/gap/MethodRecord.gi#L32-L52

In particular, I usually first put `filter_list` and `input_arguments_names`, because the belong together and specify the input, before I put `return_type`.

PR #1504, opened 2023-11-23
LinearClosure can now use a custom range_of_HomStructure

mohamed-barakat:


FreydCategoriesForCAP/gap/LinearClosure.gi
```diff
@@ -364,9 +431,13 @@ end );
...
 InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_LINEAR_CLOSURE,
-    function( category )
-        local ring, underlying_category, sorting_function, mul_coeffs, mul_supp, with_nf,
-              equality_func, finsets, rows, t_obj, t_finsets, FunctorMor, FunctorObj, cocycle;
+  function( rows, category )
+    local ring, underlying_category, sorting_function, mul_coeffs, mul_supp, with_nf,
+          equality_func, finsets, t_obj, t_finsets, FunctorMor, FunctorObj, cocycle;
+    
+    if not IsCategoryOfRows( rows ) then
+        Error( "the second argument `rows` must be a category of rows\n" );
```
zickgraf, line 439:
```suggestion
        Error( "the first argument `rows` must be a category of rows\n" );
```
or maybe swap the arguments?

FreydCategoriesForCAP/gap/LinearClosure.gi
```diff
@@ -150,6 +164,59 @@ InstallMethod( TwistedLinearClosure,
     
 end );
 
+##
+InstallMethod( LinearClosure,
```
zickgraf, line 168: I suggest putting the `InstallMethod`s in the same order as the `DeclareOperation`s (or the other way round).

zickgraf (changes requested):

FreydCategoriesForCAP/gap/LinearClosure.gi
zickgraf: The latest force-push seems to move some installations around, but the order of the declarations and the order of the installations still do not match.

mohamed-barakat: I thought you meant I should just put the installations using `IsCategoryOfRows` before the installations of `IsHomalgRing`. Otherwise, I retained the exact order of the existing code.

zickgraf (approved):

PR #1511, opened 2023-12-01
Introduce AsCapCategoryObject and AsCapCategoryMorphism

zickgraf:
which are versions of CreateCapCategory*WithAttributes with fixed attributes, which are determined from the object/morphism datum types.

More explanation follows, for now I only want to trigger the CI.


zickgraf:
Rationale of this PR:
The name "**Create**CapCategory*WithAttributes" emphasizes that a new GAP object is created, which immediately raises the question in which sense the GAP object faithfully represents the mathematical input we started with. In my thesis, I want to adapt a different point of view: For example, in the context of a matrix category, I want to view the integer `1` and the result of `CreateCapCategoryObjectWithAttributes( mat, Dimension, 1 )` as "the same thing" in the following sense: The latter is just the integer `1` with the category attached, so we could in theory also compute with `1` directly as long as we provide the category in the context. However, `CreateCapCategoryObjectWithAttributes( mat, Dimension, 1 )` does not reflect this well: The name emphasizes the creation of a GAP object and the role of the attribute `Dimension` might be unclear to a novice user.

To solve this, this PR introduces "AsCapCategoryObject", so we can simply write `AsCapCategoryObject( mat, 1 )`. This much better fits the point of view that `AsCapCategoryObject( mat, 1 )` is just `1` explicitly interpreted as an object in a CAP category. We can obtain `1` again by calling `AsInteger`.

Of course on a technical level, `1` and `AsCapCategoryObject( mat, 1 )` simply cannot be "the same thing". However, in my thesis, I use the concept of [pure functions](https://en.wikipedia.org/wiki/Pure_function) to formalize that we could in theory replace `AsCapCategoryObject( mat, 1 )` by `1` in the computation without changing the result, as long as the category is clear from the context.

The attribute `AsInteger` is chosen based on the object datum type. Currently, only `AsInteger` and `AsHomalgMatrix` corresponding to `IsInt` and `IsHomalgMatrix` are implemented, and already used for the matrix category. In the future, we could expand this, which would have the additional benefit that one does not have to create new attributes for every category.

@mohamed-barakat @sebastianpos I would like to ask for your general approval of the idea before merging.

zickgraf:
Addendum:
1. If no object datum type is given, the generic attribute `AsPrimitiveValue` is used instead of `AsInteger`, `AsHomalgMatrix` etc. Here, *primitive* is meant with regard to the category. For example, for an opposite category, an object in the underlying category is a primitive value, although it certainly is not a primitive value in the usual sense in the context of programming languages.
2. All of this of course also applies to morphisms analogously.

zickgraf:
I would like to have this for didactical reasons for my thesis and @mohamed-barakat has verbally given approval. Hence, I will merge this PR (once the CI succeeds and I'm happy with the code coverage).

@sebastianpos If you have any doubts about things in this PR, of course we can always discuss and polish things also after the merge.

PR #1532, opened 2023-12-10
Make well-definedness of morphisms in AdditiveClosure more tight

zickgraf:
Currently, we allow to encode $m \times 0$ matrices as empty lists instead of  lists with $m$ empty lists. I suggest to drop this exception: I don't see why cleanly written code would give an empty matrix in the degenerate case. The only example I know of where this ever lead to a problem was in the context of `TransposedMat`, but we have `TransposedMatWithGivenDimensions` for this.

@mohamed-barakat @kamalsaleh Do you have any doubts about this?

### DIFF
--- a/FreydCategoriesForCAP/PackageInfo.g
+++ b/FreydCategoriesForCAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "FreydCategoriesForCAP",
 Subtitle := "Freyd categories - Formal (co)kernels for additive categories",
-Version := "2023.12-09",
+Version := "2023.12-10",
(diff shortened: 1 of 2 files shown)


mohamed-barakat:
I am ok with this change. However, I suggest waiting until `CategoricalTowers` complies with the current CAP.

mohamed-barakat (dismissed):

mohamed-barakat:
> I am ok with this change. However, I suggest waiting until `CategoricalTowers` complies with the current CAP.

I now need this change to avoid warning about untyped empty lists in `precompile_AdditiveClosureOfAlgebroid`. Please rebase, bump the version, and merge.

mohamed-barakat (approved):

zickgraf:
> > I am ok with this change. However, I suggest waiting until `CategoricalTowers` complies with the current CAP.
> 
> I now need this change to avoid warning about untyped empty lists in `precompile_AdditiveClosureOfAlgebroid`. Please rebase, bump the version, and merge.

Sorry, I didn't realize that the empty list might cause problems when compiling other packages.

mohamed-barakat:
> Sorry, I didn't realize that the empty list might cause problems when compiling other packages.

No problem, it was easily located :)

PR #1535, opened 2023-12-11
Introduce IsLinearCategoryOverCommutativeRingWithFinitelyGeneratedFreeExternalHoms

zickgraf:
I would like to introduce a property for categories having `BasisOfExternalHom` and `CoefficientsOfExternalHom`. My first suggestion is
```
IsLinearCategoryOverCommutativeRingWithFiniteFreeHomSpaces
```
There are various possible alternatives:
* `...WithFinitelyGeneratedFree...`: is a bit longer but more precise
* `...HomSets`: to avoid a connection to vector spaces
* `...ExternalHoms`: matches `BasisOfExternalHom`

I don't have a strong preference. @mohamed-barakat @kamalsaleh What would be your preference?

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2023.12-07",
+Version := "2023.12-08",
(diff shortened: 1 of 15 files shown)


kamalsaleh:
I would probably have used the term `IsLinearCategoryOverCommutativeRingWithFinitelyGeneratedFreeExternalHoms`, it is long but still a bit shorter than `InterpretMorphismFromDistinguishedObjectToHomomorphismStructureAsMorphism` :-)

zickgraf:
> I would probably have used the term `IsLinearCategoryOverCommutativeRingWithFinitelyGeneratedFreeExternalHoms`, it is long but still a bit shorter than `InterpretMorphismFromDistinguishedObjectToHomomorphismStructureAsMorphism` :-)

Sounds good! I agree that we should prefer precision over brevity :-)

zickgraf:
> > Sounds good! I agree that we should prefer precision over brevity :-)
> 
> Me too.

Perfect! Thank you both for the fast reply.

PR #1536, opened 2023-12-12
Remove warnings about operations not compatible with the congruence of morphisms in proof assistant mode

zickgraf:


mohamed-barakat:
After merging this I will create a PR for `FreydCategoriesForCAP` :)

PR #1539, opened 2023-12-12
enable FreydCategory to set IsLinearCategoryOverCommutativeRingWithFinitelyGeneratedFreeExternalHoms

mohamed-barakat:


mohamed-barakat:
This PR is needed to fix https://github.com/homalg-project/CategoricalTowers/pull/465. Please feel free to alter and force push.

zickgraf:
I have updated the PR:
1. I have moved the new check inside the old check for `IsLinearCategory...`. This avoids checking `HasCommutativeRingOfLinearCategory` twice and allows to ensure that all the checks refer to the same ring.
2. We should not use `and` for filters if we want future compatibility with Julia.

PR #1546, opened 2023-12-13
Add various logic templates for proofs

zickgraf:


mohamed-barakat:
Wonderful!

CompilerForCAP/gap/ProofAssistant.gi
```diff
@@ -17,3 +17,171 @@ InstallGlobalFunction( CapJitDisableProofAssistantMode, function ( )
...
+    )
+);
+
+# ForAll( list, l -> true ) => true
+CapJitAddLogicTemplate(
+    rec(
+        variable_names := [ "list" ],
+        src_template := "ForAll( list, l -> true )",
+        dst_template := "true",
+        apply_in_proof_assistant_mode := "only",
```
mohamed-barakat, line 136: Why is this logic template, for example, only restricted to the proof assistant?

zickgraf: Because I do not want to slow down the "normal" compilation by logic templates which typically only appear in proofs. If any of these templates also appears more widespread in "normal" compilation, we can of course turn it into a regular logic template.

mohamed-barakat: Thanks :)

PR #1574, opened 2024-01-21
outsourced the adjunction operations to seperate files

mohamed-barakat:
in order to use their left AND right versions for the symmetric monoidal structures


mohamed-barakat:
Do not need this anymore. Will close without merging.

PR #1586, opened 2024-02-06
derived left from right tensor-internal hom-adjunction and vice versa

mohamed-barakat:


MonoidalCategories/gap/SymmetricClosedMonoidalCategoriesDerivedMethods.gi
```diff
@@ -4,6 +4,132 @@
...
+                                   TensorProductOnObjects( cat, b, a ),
+                                   a,
+                                   b,
+                                   Source( f ) ),
+                           ## f: a ⊗ b → c
+                           f ),
+                   ## Hom( a, c )
+                   i );
+    
+end : CategoryFilter := IsSymmetricClosedMonoidalCategory );
```
TKuh, line 34: From today's discussion: would `IsBraidedMonoidalCategory` suffice?

PR #1589, opened 2024-02-15
updated constructive record for IsEquippedWithHomomorphismStructure

mohamed-barakat:


mohamed-barakat:
Closed after discussion with Fabian: For experts, the advantages of WithGiven are known and for newcomers, the WithGiven variants are probably confusing when shown as an output of `MissingOperationsForConstructivenessOfCategory`.

PR #1592, opened 2024-02-19
Introduce FunctionWithNamedArguments

zickgraf:


zickgraf:
This simulates named arguments in GAP in a way which can be efficiently translated to Julia.

PR #1598, opened 2024-02-23
Defer installation of derivations to `Finalize`

zickgraf:
This avoids overwriting GAP methods if cheaper derivations become available while adding primitive installations to a category. The only difference to the previous behavior should be that derivations cannot be called anymore before a category is finalized.

@mohamed-barakat @sebastianpos I think derivations were previously installed immediately for didactical purposes. However, this causes issues in Julia, where methods may not be overwritten during precompilation. I do not think the original purpose is too important anymore, so I will merge if you do not oppose :-)


mohamed-barakat:
Very nice :)

PR #1606, opened 2024-03-19
Make use of features introduced in GAP 4.13

zickgraf:
This is a copy of #1604.


zickgraf:
@mohamed-barakat @kamalsaleh Can we update to GAP 4.13 officially?

PR #1616, opened 2024-06-27
the opposite category constructor now respects the enrichment of the input category

mohamed-barakat:
and SetRangeCategoryOfHomomorphismStructure in ReinterpretationOfCategory if it is set for its input, making ReinterpretationOfCategory behave more like its input.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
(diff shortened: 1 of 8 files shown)


mohamed-barakat:
The corresponding PR in CategoricalTowers eliminates the annoying if-else code and passes my local tests.

mohamed-barakat:
Merged after OK from @zickgraf.

zickgraf:
> Merged after OK from @zickgraf.

As discussed, I expect that this causes maintenance and/or documentation problems, which will have to be fixed sooner or later.

PR #1617, opened 2024-06-28
declared IsCategoryWithInitial/Terminal/ZeroObject

mohamed-barakat:


CAP/gap/ConstructiveCategoriesRecord.gi
```diff
@@ -41,6 +41,32 @@ CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.IsCategoryWithDecidableColifts := Co
...
+    CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.EveryCategory,
+    [
+        "TerminalObject",
+        "UniversalMorphismIntoTerminalObject",
+    ]
+);
+
+CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.IsCategoryWithZeroObject := Concatenation(
+    CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.IsCategoryWithInitialObject,
+    CAP_INTERNAL_CONSTRUCTIVE_CATEGORIES_RECORD.IsCategoryWithTerminalObject,
```
zickgraf, line 62: I would not include `IsCategoryWithInitialObject` and `IsCategoryWithTerminalObject` here for the following reasons:
1. Consistency with `IsAdditiveCategory`, which does not include `DirectProduct` etc.
2. The following situation: A user installs `ZeroObject`, `UniversalMorphismFromZero/IntoObject`, and  `TerminalObject`, but not `UniversalMorphismIntoTerminalObject`. Then the final derivation deriving `TerminalObject` from `ZeroObject` does not trigger, so the category algorithmically has a zero object, but not a terminal object.

mohamed-barakat: Thanks, this is something I was not sure about.

PR #1620, opened 2024-07-01
Use HomalgMatrixListList at more places

zickgraf:

PR #1621, opened 2024-07-02
Use HomalgMatrixListList at more places

zickgraf:
Copy of #1620 from a different branch because we defer the merge of #1620.


mohamed-barakat:
Could you please bump the package versions?

zickgraf:
> Could you please bump the package versions?

Ah, the version bumps were lost during the rebase.

PR #1624, opened 2024-07-03
SetIsLinearCategoryOverCommutativeRingWithFinitelyGeneratedFreeExternalHoms in LinearClosure

mohamed-barakat:


zickgraf:
For the record: As discussed verbally, the behaviour in this PR is not correct, probably due to the interface of `LINEAR_CLOSURE_CONSTRUCTOR_USING_CategoryOfRows` being misleading, i.e. the interface should be refactored.

PR #1633, opened 2024-07-17
Distinguish between small and big integers

zickgraf:
This is a tentative attempt to distinguish between small and big integers in GAP to mirror the situation in Julia. Doing this across the whole codebase seems like an enormous effort. I'm not sure if this effort is worth it.

@mohamed-barakat If you are interested in this, we can look at some examples (e.g. the CategoryOfRows) in more detail and check what would have to be done.

### DIFF
--- a/CAP/gap/InstallAdds.gi
+++ b/CAP/gap/InstallAdds.gi
@@ -127,7 +127,7 @@ InstallGlobalFunction( CapInternalInstallAdd,
             
         fi;
         
-        InstallMethod( CAP_operation, replaced_filter_list{[ 2 .. Length( replaced_filter_list ) ]}, get_convenience_function( CAP_operation ) );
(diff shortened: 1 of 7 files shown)


mohamed-barakat:
Yes, I am interested. I would also like to have a version of SkeletalFinSets which uses the small integers and not the gmp-integers.

zickgraf:
> Yes, I am interested. I would also like to have a version of SkeletalFinSets which uses the small integers and not the gmp-integers.

What application do you have in mind? This category would not have products etc. anymore, correct?

zickgraf:
There are also some subtleties in Julia when working with ranges of big ints:
```
julia> typeof(length(BigInt(1):BigInt(1)))
BigInt

julia> typeof(length(map(x -> x,BigInt(1):BigInt(1))))
Int64
```

mohamed-barakat:
@kamalsaleh: Could this be interesting for what you are currently working on?

PR #1634, opened 2024-07-18
Re-encode data types of empty lists

zickgraf:


zickgraf:
@mohamed-barakat If I compile FinQuivers in FunctorCategories with this PR, I get the following code:
```
    ##
    AddInitialObject( cat,
        
########
function ( cat_1 )
    local deduped_1_1;
    deduped_1_1 := BigInt( 0 );
    return CreateCapCategoryObjectWithAttributes( cat_1, DefiningTripleOfQuiverEnrichedOverSkeletalFinSets, NTuple( 3, deduped_1_1, deduped_1_1, CapJitTypedExpression( [  ], function (  )
                return rec(
                    filter := IsList,
                    element_type := rec(
                        filter := IsInt ) );
            end ) ) );
end
########
        
    , 100 );
```
Notice that the third entry of `DefiningTripleOfQuiverEnrichedOverSkeletalFinSets` is annotated to be a list of integers. However, according to the type declaration, the third entry of `DefiningTripleOfQuiverEnrichedOverSkeletalFinSets` should be a list of pairs of integers. I have checked the data types in FinSetsForCAP and have actually found a problem (https://github.com/homalg-project/FinSetsForCAP/pull/257), but this does not fix the above. Hence, I assume that some type declaration in FunctorCategories is wrong. Would you like to have a look before I merge, or shall I merge and we fix things afterwards?

PR #1639, opened 2024-08-15
Various adjustments for making LinearAlgebraForCAP compatible with Julia

zickgraf:


zickgraf:
The changes in GroupRepresentationsForCAP are best-effort due to parts of the code not being covered by tests.

PR #1641, opened 2024-08-20
Allow to specify weights of operations in CategoryConstructor

zickgraf:
and use this to preserve weights when constructing reinterpretations

@mohamed-barakat With this PR, providing the weights is optional. I suggest to make the weights mandatory to make sure that we always preserve the weights whenever we use CategoryConstructor. Do you agree?

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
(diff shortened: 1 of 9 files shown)


zickgraf:
The [coverage changes](https://app.codecov.io/gh/homalg-project/CAP_project/pull/1641/indirect-changes) show the importance of preserving the weights: The correct weights ensure that redirections to WithGiven operations only happen if we actually expect an advantage (i.e. calling the WithGiven operation is actually cheaper than the non-WithGiven operation).

PR #1678, opened 2024-09-13
Replace operation_weights in CategoryConstructor by a better mechanism

zickgraf:
When using the string "default" for create_func_\*, one must now set the option `underlying_category`, from which the weight is determined. If create_func_\* is a function, one can (optionally) set a weight by returning a pair of a string and an integer (instead of only a string).

@mohamed-barakat This PR partially implements a suggestion you had: You suggested to pass an underlying category to CategoryConstructor from which the weights can be taken. I argued that this does not make sense in general. However, while working with the weights, I noticed that it indeed does make sense in a specific case, namely when using the string "default" for create_func_\*. This PR now requires to set the option `underlying_category` when using the string "default". This sets the correct weights, which probably requires changes like recompiling code. Please tell me when I can merge, i.e. when you have to time to adjust things in CategoricalTowers.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2024.10-08",
+Version := "2024.11-01",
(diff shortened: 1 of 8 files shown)


mohamed-barakat:
Das gefällt mir sehr gut, vielen Dank. Ich bräuchte allerdings etwas Zeit um die notwendigen Anpassungen zu machen.

zickgraf:
@mohamed-barakat You can merge this PR once the changes required in CategoricalTowers are implemented.

PR #1704, opened 2024-10-17
Cache precompiled functions without irreversible optimizations for future use

zickgraf:


zickgraf:
Ignoring the cached functions completely for the code coverage is not as simple as I thought because the `end` of the function is actually marked as covered and ignoring covered lines throws an error to avoid misuse of `COVERAGE_IGNORE_*`. Hence, FinSetsForCAP will have to actually call the cached functions in the tests.

PR #1748, opened 2025-06-21
Clarify category of source/range objects in `ReinterpretationOfMorphism`

TKuh:
I become constantly confused to which category `source` and `range` in a call to `ReinterpretationOfMorphism` belong to and the documentation does not clearly state this.

### DIFF
--- a/CAP/gap/ReinterpretationOfCategory.gd
+++ b/CAP/gap/ReinterpretationOfCategory.gd
@@ -249,8 +249,8 @@ CapJitAddTypeSignature( "ModelingMorphism", [ IsCapCategory, IsCapCategoryMorphi
 end );
 
 #! @Description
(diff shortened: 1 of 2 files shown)


CAP/gap/ReinterpretationOfCategory.gd
```diff
@@ -249,7 +249,7 @@ CapJitAddTypeSignature( "ModelingMorphism", [ IsCapCategory, IsCapCategoryMorphi
 end );
 
 #! @Description
-#!  Returns the reinterpretation in `R` with given source and range of a morphism <A>mor</A> in `ModelingCategory`(<A>R</A>).
+#!  Returns the reinterpretation in `R` with given source and range in `R` of a morphism <A>mor</A> in `ModelingCategory`(<A>R</A>).
 #! @Arguments R, source, obj, range
```
mohamed-barakat, line 253: please also change obj -> mor

PR #1758, opened 2025-06-30
New package `AdditiveClosuresForCAP`

TKuh:
Split from `FreydCatgoriesForCAP`.


mohamed-barakat (changes requested):
Thank you very much. Could you please
1. on this branch pull --rebase CAP_project
2. pull --rebase PackageJanitor
3. apply PackageJanitor (it will delete the subsplit push for GroupsAsCategories)
4. Rebump version of ModulePresentationsForCAP
5. update this PR

TKuh:
> There is an error during simulating the release:
> 
> ```
> ERROR: HTML manual contains absolute paths
> ```

Yes, but I can't figure out where these absolute paths come from. :/ On my machine, I also get absolute paths for `CAP`, and they do not seem to cause trouble?

kamalsaleh:
For some reason, I can see in the action logs the following line:

`makefile:136: target 'ci-test_AdditiveClosuresForCAP' does not exist`

The same for other packages like _AttributeCategoryForCAP_. Maybe investigating that and comparing with old actions might solve the issue.

mohamed-barakat (approved):

PR #1759, opened 2025-07-01
GitHub Actions: build documentation of `AdditiveClosuresForCAP`

TKuh:


mohamed-barakat:
Shouldn't this be part of the original PR?

TKuh:
But does the CI then immediately use the new workflow? I thought it would execute the new PR with its old workflow.

It shouldn't do any harm, since 
```
if [ -d "CAP_project/AdditiveClosuresForCAP" ];
```
executes only if the directory `AdditiveClosuresForCAP` exists anyway.

kamalsaleh:
Uploading to Codecov fails—or takes too long—because the CODECOV_TOKEN secret is, for security reasons, not available in pull requests from external forks. It’s only accessible during push events or pull requests originating from a branch within the same repository. This is why GitHub Actions often run faster once a PR has been merged.

About the workflow file: I believe GitHub-Actions will use the updated version of the workflow file. So what Mohamed suggested about including the line in your original PR is actually valid.

TKuh:
> Uploading to Codecov fails—or takes too long—because the CODECOV_TOKEN secret is, for security reasons, not available in pull requests from external forks. It’s only accessible during push events or pull requests originating from a branch within the same repository. This is why GitHub Actions often run faster once a PR has been merged.
> 

So it will always fail for my PR's? :)

> About the workflow file: I believe GitHub-Actions will use the updated version of the workflow file. So what Mohamed suggested about including the line in your original PR is actually valid.

Ok, thank you. But I leave this PR open for now, just in case.

PR #1762, opened 2025-07-02
Distinguish between small and big integers

mohamed-barakat:
This PR is a rebased resurrection of https://github.com/homalg-project/CAP_project/pull/1633 that I had to close since I cannot push into Fabian's branches:

@zickgraf wrote in https://github.com/homalg-project/CAP_project/pull/1633#issue-2413789218:

> This is a tentative attempt to distinguish between small and big integers in GAP to mirror the situation in Julia. Doing this across the whole codebase seems like an enormous effort. I'm not sure if this effort is worth it.
> 
> @mohamed-barakat If you are interested in this, we can look at some examples (e.g. the CategoryOfRows) in more detail and check what would have to be done.

### DIFF
--- a/AdditiveClosuresForCAP/gap/AdditiveClosure.gi
+++ b/AdditiveClosuresForCAP/gap/AdditiveClosure.gi
@@ -926,7 +926,7 @@ InstallGlobalFunction( INSTALL_FUNCTIONS_FOR_ADDITIVE_CLOSURE,
       function( cat, diagram, test_object, morphisms, direct_sum )
         local listlist;
         
-        listlist := UnionOfColumnsListList( Length( ObjectList( test_object ) ), List( morphisms, tau -> MorphismMatrix( tau ) ) );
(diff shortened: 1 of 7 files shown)


mohamed-barakat:
@mohamed-barakat  wrote in https://github.com/homalg-project/CAP_project/pull/1633#issuecomment-2234045047:

>> This is a tentative attempt to distinguish between small and big integers in GAP to mirror the situation in Julia. Doing this across the whole codebase seems like an enormous effort. I'm not sure if this effort is worth it.
>> 
>> @mohamed-barakat If you are interested in this, we can look at some examples (e.g. the CategoryOfRows) in more detail and check what would have to be done.



> Yes, I am interested. I would also like to have a version of SkeletalFinSets which uses the small integers and not the gmp-integers.

mohamed-barakat:
@zickgraf replied in https://github.com/homalg-project/CAP_project/pull/1633#issuecomment-2235691045:

> > Yes, I am interested. I would also like to have a version of SkeletalFinSets which uses the small integers and not the gmp-integers.
> 
> What application do you have in mind? This category would not have products etc. anymore, correct?

mohamed-barakat:
@kamalsaleh: Could this be interesting for what you are currently working on?

PR #1766, opened 2025-07-06
New package `LinearClosuresForCAP`

TKuh:
Split from `FreydCategoriesForCAP`.

And remove superfluous loading of `AdditiveClosuresForCAP` in `GroupsAsCategoriesForCAP` and `CompilerForCAP`.


mohamed-barakat (changes requested):
1. Why is `AdditiveClosuresForCAP` a hard dependency for `LinearClosuresForCAP`? I would expect it to be the other way.
2. I would then suggest making `AdditiveClosuresForCAP` (and then indirectly `LinearClosuresForCAP`) a hard dependency for `FreydCategoriesForCAP`.

TKuh:
> 1. Why is `AdditiveClosuresForCAP` a hard dependency for `LinearClosuresForCAP`? I would expect it to be the other way.
> 

Because `CategoryOfRows` is now in `AdditiveClosuresForCAP` and `LinearClosure` explicitly uses it:
https://github.com/homalg-project/CAP_project/blob/e387f141b1441147ca566e655d79e26019d4df20/FreydCategoriesForCAP/gap/LinearClosure.gi#L139-L148

So `LinearClosureForCAP` must load `AdditiveClosuresForCAP` and not the other way around.

mohamed-barakat (approved):

PR #1780, opened 2025-07-19
Whitespace: function () -> function()

TKuh:


mohamed-barakat:
`function ( )` is the correct GAP style.

```
gap> Display( x -> x );
function ( x )
    return x;
end
```

PR #1790, opened 2025-08-10
`SumOfMorphisms` for `LinearClosure`

TKuh:
It was previously derived via an iterated application of `AdditionForMorphisms`, but this is more direct.

It is also needed for compiling, otherwise there will be an `Iterated(...)` leading to multiple unresolved `CreateCapCategoryMorphismWithAttributes`

Additionally, some type signatures for compiling `LinearClosure`.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2025.08-02",
+Version := "2025.08-03",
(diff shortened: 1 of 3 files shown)


TKuh:
@mohamed-barakat if you still dislike the implementation of `SumOfMorphisms` I can remove it. Now that I have `LinearClosureOverFiniteSkeletalDiscreteCategory` I don't need it anymore.

mohamed-barakat:
> @mohamed-barakat if you still dislike the implementation of `SumOfMorphisms` I can remove it. Now that I have `LinearClosureOverFiniteSkeletalDiscreteCategory` I don't need it anymore.

You can turn it into a draft or close it, as you like.

mohamed-barakat:
> And what about the type signatures for `MinusOne` and `UnderlyingListOfGapObjects`? Or do you prefer to have them merged later when my compiled code is ready?

Yes, the signatures should only be added when they are used for the first time.

PR #1801, opened 2025-09-22
Declare CallFuncListAtRuntime as a synonym of CallFuncList.

kamalsaleh:
This is used to avoid Julia world-age conflicts by delegating calls to Base.invokelatest(func, arg...).
Use it for installing convenience methods for the CAP operations

bump CAP to v2025.09-04

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2025.09-03",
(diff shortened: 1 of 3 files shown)


kamalsaleh:
This would fix the segmentation-fault in julia-master and hence the failing tests in _CAP_Project.jl_ (https://github.com/homalg-project/CAP_project.jl/actions/runs/17902134350) and _FinSetsForCAP_ (https://github.com/homalg-project/FinSetsForCAP/actions/runs/17904280933). But I still need to investigate what exactly happend on the julia side that now these problems appear.

PR #1819, opened 2026-03-26
Add specialized operations when tensoring with an identity morphism

TKuh:
This PR introduces the operations

- `TensorProductOfMorphismAndIdentity`
- `TensorProductOfMorphismAndIdentityWithGivenTensorProducts`
- `TensorProductOfIdentityAndMorphism`
- `TensorProductOfIdentityAndMorphismWithGivenTensorProducts`

and their counterparts in `CartesianCategoriesForCAP`. This is necessary for optimized operations in the upcoming `PermutationCategory`.


mohamed-barakat:
Please rebase :)

mohamed-barakat (changes requested):
Please bump version and dependencies in `LinearAlgebraForCAP`.

MonoidalCategories/gap/MonoidalCategories.gd
```diff
@@ -34,6 +34,42 @@ DeclareOperation( "TensorProductOnMorphisms",
 DeclareOperation( "TensorProductOnMorphismsWithGivenTensorProducts",
                   [ IsCapCategoryObject, IsCapCategoryMorphism, IsCapCategoryMorphism, IsCapCategoryObject ] );
 
+#! @Description
+#! The arguments are two morphisms $\alpha: a \rightarrow a', \mathrm{id}_b: b \rightarrow b$.
+#! The output is the tensor product $\alpha \otimes \mathrm{id}_b$.
+#! @Returns a morphism in $\mathrm{Hom}(a \otimes b, a' \otimes b)$
+#! @Arguments alpha, id_b
+DeclareOperation( "TensorProductOfMorphismAndIdentity",
+                  [ IsCapCategoryMorphism, IsCapCategoryMorphism ] );
```
mohamed-barakat: Why do we need the identity here, an object would be enough?

MonoidalCategories/gap/MonoidalCategories.gd
```diff
@@ -34,6 +34,42 @@ DeclareOperation( "TensorProductOnMorphisms",
...
+DeclareOperation( "TensorProductOfMorphismAndIdentityWithGivenTensorProducts",
+                  [ IsCapCategoryObject, IsCapCategoryMorphism, IsCapCategoryMorphism, IsCapCategoryObject ] );
+
+#! @Description
+#! The arguments are two morphisms $\mathrm{id}_a: a \rightarrow a, \beta: b \rightarrow b'$.
+#! The output is the tensor product $\mathrm{id}_a \otimes \beta$.
+#! @Returns a morphism in $\mathrm{Hom}(a \otimes b, a \otimes b')$
+#! @Arguments id_a, beta
+DeclareOperation( "TensorProductOfIdentityAndMorphism",
+                  [ IsCapCategoryMorphism, IsCapCategoryMorphism ] );
```
mohamed-barakat: Dito

MonoidalCategories/gap/MonoidalCategories.gd
TKuh: Bifunctors like $\otimes$ are maps of the form

(object, object) $\mapsto$ object
(morphism, morphism) $\mapsto$ morphism

A mixing like (morphism, object) does not type-check. But I am aware that people like to abbreviate identities with their objects, so a compromise would be to introduce convenience functions like

```gap
DeclareOperation( "TensorProductOfMorphismAndIdentity",
                  [ IsCapCategoryMorphism, IsCapCategory**Object** ] );
```

which delegate to the proper CAP operations.

However, I personally don't need this and without a use-case I would prefer to not add these conveniences. I also always thought of these abbreviations as imprecise.

TKuh: We agreed on renaming them to `TensorProductOfMorphismAndObject` and `TensorProductOfObjectAndMorphism`, since in particular the data structure of an identity morphism will likely not be relevant.

mohamed-barakat (approved):

PR #1825, opened 2026-04-16
Apply PackageJanitor

kamalsaleh:


mohamed-barakat (approved):

mohamed-barakat:
Excellent, thanks @kamalsaleh.

mohamed-barakat:
The next step would be to parallelize the tests, right?

PR #1833, opened 2026-05-19
d&i CapJitDataTypeOfSubgroup and added CapJitAddTypeSignature

mohamed-barakat:
for
* ConjugateSubgroup
* IsSubset (for groups)

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2026.05-03",
+Version := "2026.05-04",
 Date := (function ( ) if IsBound( GAPInfo.SystemEnvironment.GAP_PKG_RELEASE_DATE ) then return GAPInfo.SystemEnvironment.GAP_PKG_RELEASE_DATE; else return Concatenation( ~.Version{[ 1 .. 4 ]}, "-", ~.Version{[ 6, 7 ]}, "-01" ); fi; end)( ),
 License := "GPL-2.0-or-later",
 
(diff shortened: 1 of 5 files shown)


kamalsaleh:
Could you retry precompiling **FinGSetsForCAP** using this commit? https://github.com/kamalsaleh/CAP_project/commit/18391c45fc6c68e6ea035348a2e6543c8cd1a6a9

mohamed-barakat:
> Could you retry precompiling **FinGSetsForCAP** using this commit? [kamalsaleh@18391c4](https://github.com/kamalsaleh/CAP_project/commit/18391c45fc6c68e6ea035348a2e6543c8cd1a6a9)

Excellent, this resolves the infinite-loop issue. One million thanks.

kamalsaleh:
The comparison happens at two places via GAP's built-in `<>` on records, which recurses into all fields including `data_type.filter`, where it falls back to identity comparison (`IsIdenticalObj`). There's no dedicated filter-comparison function — the filter equality check is implicit.

The two loop headers in `CompilerForCAP/gap/CompilerForCAP.gi`:

- **Line 304** — resolving phase: `while tree <> orig_tree do`
https://github.com/homalg-project/CAP_project/blob/e13064cf466078004ef7696069b0f165f42297ca/CompilerForCAP/gap/CompilerForCAP.gi#L304
- **Line 386** — rule phase: `while tree <> orig_tree do`
https://github.com/homalg-project/CAP_project/blob/e13064cf466078004ef7696069b0f165f42297ca/CompilerForCAP/gap/CompilerForCAP.gi#L386

To improve the comparison at this level instead of caching, one would need to replace `tree <> orig_tree` with a custom comparison function, e.g. `not CapJitIsEqualTree( tree, orig_tree )`, where `CapJitIsEqualTree` handles `data_type` records by comparing filters with mutual implication:

```gap
# for data_type records
IsSpecializationOfFilter( r1.filter, r2.filter ) and
IsSpecializationOfFilter( r2.filter, r1.filter )
```

That would be more semantically correct than the caching workaround, since it handles any case where different filter objects represent the same filter - not just the specific `IsGroup and HasParent` case. But it also requires writing a full recursive tree equality function, which is more invasive. The caching fix is simpler and already sufficient for the known case.

PR #1834, opened 2026-05-19
Use `WithGiven*` in derivations of distributivity morphisms

TKuh:


mohamed-barakat:
It seems the new code mainly has `CAP_JIT_INCOMPLETE_LOGIC` inserted. Do you understand why?

PR #1857, opened 2026-06-29
Apply PackageJanitor

kamalsaleh:
now makefiles contain targets to match extracted test files to examples.


mohamed-barakat (approved):

mohamed-barakat:
Many thanks.

PR #1871, opened 2026-08-04
Construct `FiniteSkeletalDiscreteCategory` from a single integer argument

TKuh:
And not from a list of GAP objects. This streamlines it with `SkeletaFinSets`.

Additionally, set `overhead := false` in an example, to (hopefully) increase the code coverage.

### DIFF
--- a/CAP/PackageInfo.g
+++ b/CAP/PackageInfo.g
@@ -10,7 +10,7 @@ SetPackageInfo( rec(
 
 PackageName := "CAP",
 Subtitle := "Categories, Algorithms, Programming",
-Version := "2026.07-04",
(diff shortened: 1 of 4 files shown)


mohamed-barakat (changes requested):
I don't like this PR because 
* it is downgrading an existing functionality
* the numbering is not compatible with `SkeletalFinSets` but with `SkeletalFinSets1`.

As you see from the PR, the body of the implementation did not change, so you can use it to create three different constructors, or one constructor with internals depending on the input:
1. the old behavior with an input list
2. `n` means `[ 0 .. n-1 ]`
3. `n` means `[ 1 .. n ]`

